Issue #306, opened 2012-04-24
XStreamConverters on data tables don't work anymore when used on plain step params


tknerr:
Since version 1.0.4 the XStreamConverters work for both data tables and plain step parameters (yay!), but there seems to be an ordering issue:
- when the data table converters get registered first => everything works as expected
- when the plain step param converters get registered first => the converter annotations on the data table are ignored

For more context see the original discussion thread:
http://groups.google.com/group/cukes/browse_thread/thread/ca6a431903c98cdb

Pull request with test cases follows...

mpkorstanje:
Closing this issue. With cucumber-expressions we'll remove xstream.

PR #307, opened 2012-04-24
features for testing XStreamConverters on both regular and data table arguments + test case for issue #306

tknerr:


mpkorstanje:
Closing this issue. With cucumber-expressions we'll remove xstream.

Issue #290, opened 2012-04-04
Ability to configure PicoContainer


rapaul:
This feature request rises from http://groups.google.com/group/cukes/browse_thread/thread/a1e034468df4b5f2

In order to test across a range of environments
I want to configure the instance of a dependency in PicoContainer

Example snipped from the mailing list:

```java
public class MyHooks { 
    private final MutablePicoContainer pico; 
    public MyHooks(MutablePicoContainer pico) { 
        this.pico = pico; 
    } 
    @Before("@foo") 
    public void registerFooThing() { 
        pico.addComponent(new MyThing("FOO")); 
    } 
    @Before("@bar") 
    public void registerBarThing() { 
        pico.addComponent(new MyThing("BAR")); 
    }
    @Before 
    public void registerWebDriverBasedOnSystemProperty() {
        String seleniumGrid = System.getProperty("selenium.grid");
        if (seleniumGrid == null) {
            pico.add(new FirefoxDriver());
        } else {
            pico.addComponent(new RemoteWebDriver(seleniumGrid)); // This isn't exact syntax but you get the idea
        }
    }
} 
```

Currently it isn't possible to have an instance of  `MutablePicoContainer` injected into the `MyHooks`.

PR #450, opened 2013-01-17
The PicoContainer instance can now be configured

rapaul:
I've had a go at issue #290

By implementing your own PicoConfigurer (and pointing to it from
cucumber-picocontainer.properties) you can add arbitrary mappings to the
MutablePicoContainer instance.

Particularly useful if you want to have Step Defs or other classes
depend on an interface. The implementation to use can be mapped in your
custom PicoConfigurer.

This work is based on the custom Module support in cucumber-guice.

### DIFF
--- a/picocontainer/pom.xml
+++ b/picocontainer/pom.xml
@@ -44,6 +44,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-all</artifactId>
+            <scope>test</scope>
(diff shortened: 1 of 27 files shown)


georgecodes:
Is this ever going to be merged?

aslakhellesoy:
I'm afraid not, but I'm planning to provide a different mechanism to configure Pico using the new Plugin API.

georgecodes:
Ah ok, thanks for letting me know. I'll go in another direction then.

PR #505, opened 2013-04-15
Use the bool library to parse tag expressions.

aslakhellesoy:
Scala and Clojure are currently disabled, pending fixes.

UPDATE: Scala and Clojure are uptodate now


mattwynne:
Out of date.

PR #567, opened 2013-08-02
Return exit code 0 when no features are found, or no found features match the tags filter

brasmusson:
This PR is created in response to the following discussion on the mailing list:
[JVM] Exception thrown when no tags are found (https://groups.google.com/forum/#!topic/cukes/wZo0FTQJrVA).

Return exit code 0 when no features are found, or if no found features match the tags filter used. Print the same message, that is used in the currently thrown CucumberException when this occurs, as information before the summary printout.

Jenkins will mark a job as failed even though the exit code was 0, if the JUnit report file contains no testcase elements (http://jenkins.361315.n4.nabble.com/Jenkins-fails-if-there-are-no-testcases-td3899185.html). Therefore add a dummy testcase element to the JUnit report, when no features are found, or no found features match the tags filter used.

### DIFF
--- a/examples/java-no-features/.gitignore
+++ b/examples/java-no-features/.gitignore
@@ -1,0 +1,7 @@
+/.settings
+/.classpath
+/.project
+.idea
+*.iml
+.DS_Store
+
--- a/examples/java-no-features/README.md
+++ b/examples/java-no-features/README.md
@@ -1,0 +1,11 @@
+# java-no-features
+
+If wonder what would Cucumber JVM do when there are no feature files found at all.
+Just run these cukes by
+
+    mvn clean test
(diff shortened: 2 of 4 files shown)


kutzi:
This is already really old, but:
could we at least have the part, where a lot of 'dummy' test results are created, made configurable?
IMO it's not the responsibility of cucumber-jvm to add workarounds, when your Jenkins job configuration is wrong.

kutzi:
Currently, my Jenkins builds are overflowing with this 'dummy' results which is really annoying

kutzi:
See https://github.com/cucumber/cucumber-jvm/issues/1326

PR #588, opened 2013-09-12
Proposal to fix Issue577 -- improved junit runner

BertRobbenSony:
This pull request fixes the junit runner such that:

(1) Each scenario is represented as a test. Each example of a scenario outline is also represented as a test.
[the original junit runner represented each step as a test which is highly confusing].

(2) From the IDE (tested in Eclipse), it is now possible to select a single test (either scenario or one example from a scenario outline) and do right-click run to run only that particular test.


mpkorstanje:
After #1035 each example is its own pickle. Pickles can be presented as a test by passing` --no-step-notifications`  introduced by #1135.

PR #601, opened 2013-10-04
Dependency injection for Cucumber-Scala

VivaceVivo:
This change set changes the way objects are created in the ScalaBackend similar to the JavaBackend to support dependency injection.
The Cucumber-SubCut module supports the light weight Scala dependency Injection mechanism SubCut.
(https://github.com/dickwall/subcut)


dkowis:
The travis build no longer succeeds, mostly because of versioning. I'm not sure how to tell travis to run the test again so you can see the failures...

dkowis:
Also, I'm considering letting this one close, because it's over 2 years old. Does anyone still want this one? (One can just use spring in scala natively for dependency injection if they want...)

aptester:
@dkowis Can you share an example where spring can be used with cucumber-scala to share context between two step classes?

PR #626, opened 2013-11-09
Restructured JUnit class/method name reporting

sirchia:
I propose the following slight tweaks to the way the scenarios/steps are mentioned in the name and classname attributes of the JUnit XML report:
“Scenario( Outline): <scenario name>” map to the ‘package’ of the JUnit reported classname
“|<example row>|” or “Scenario” map to the class name of the JUnit reported classname
“<step keyword> step description” map to the name of the JUnit reported test case

The main reason for this change is to have a more structured view in a JUnit based test overview page like that of Jenkins, where on the 1st page a list of packages (=scenario names) is shown. Clicking on a package shows all test classes (=examples, or just 'Scenario' if it isn't an outline scenario) in that package. Clicking on a class shows all test methods (=steps of that scenario).

Also, failing tests are shown on the overview page with their complete package.classname.testmethod description and with this mapping, it is instantly clear which step of which scenario (example) has failed.

In order to make sure the parsing of packages (substring of 0 - <index of last dot> in classname) and JUnit internal distinction of class and method in its description goes alright, the following character replacements are done in the scenario and descriptions and example rows:
. translates to ,
( translates to {
) translates to }

Finally, the redundant reporting on the scenario itself has been removed.

### DIFF
--- a/core/src/main/java/cucumber/runtime/model/CucumberTagStatement.java
+++ b/core/src/main/java/cucumber/runtime/model/CucumberTagStatement.java
@@ -19,13 +19,13 @@ public abstract class CucumberTagStatement extends StepContainer {
     CucumberTagStatement(CucumberFeature cucumberFeature, TagStatement gherkinModel) {
         super(cucumberFeature, gherkinModel);
         this.gherkinModel = gherkinModel;
-        this.visualName = gherkinModel.getKeyword() + ": " + gherkinModel.getName();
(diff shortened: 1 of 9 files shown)


mpkorstanje:
This PR is based on an implementation that used an older version of Gherkin and can't be merged anymore. However with #1135 we'll more or less have the functionality as desired.

PR #627, opened 2013-11-09
Added support to define skipTags to explicitly skip the execution of scenarios

sirchia:
This marks them and their steps as skipped instead of just not executing them which would be the case if ~@tag was used.

The rationale behind this is to provide a similar functionality as the @Ignore annotation on normal JUnit tests. These tests will not fail the build, but will be clearly visible in the JUnit reports. 

This might be useful for example when a Scenario or even just some example sets of a Scenario Outline are to be temporarily skipped while waiting for a known bug to be solved or when tests for a future sprint are already being defined while engineers are still busy finishing the implementation of the current stories.

Note that this change requires the merge of [this pull request](https://github.com/cucumber/gherkin/pull/286) to address an issue in the Gherkin Java parser where tags on example tables could go lost. Without that merge, tags made on the first n of n+1 example tables in one scenario outline would not be matched as they would be wiped from the internal model before the skip criteria would be checked.

### DIFF
--- a/core/src/main/java/cucumber/api/CucumberOptions.java
+++ b/core/src/main/java/cucumber/api/CucumberOptions.java
@@ -37,6 +37,11 @@ public @interface CucumberOptions {
     String[] tags() default {};
 
     /**
+     * @return what tags in the features should be skipped explicitly
+     */
+    String[] skipTags() default {};
+
+    /**
      * @return what formatter(s) to use
      */
     String[] format() default {};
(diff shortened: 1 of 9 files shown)


mpkorstanje:
I'm sorry nobody has responded to this any earlier. Unfortunately the code base has changed significantly and I'm not able to merge this.

If you still have a use case for this feature it might be a good idea to start a discussion in the google groups first so we can tease out the underlying motivation.

PR #674, opened 2014-02-26
Expose CucumberBackground and CucumberFeature from scenarios

jhorstmann:
For a custom reporting output I need to access the background steps of scenarios. The background was previously available from `CucumberScenario`, but not from `CucumberScenarioOutline`, but it really belongs to a feature. It would also be nice to expose the feature a scenario belongs to, in our report we currently need to either use reflection hacks or pass pairs of feature/scenario objects.
- Add method to retrieve the CucumberFeature from a CucumberScenario
- Move CucumberBackground from CucumberScenario to its parent CucumberTagStatement, so its also available for scenario outlines


intvprep:
Even I am stuck at Background being available in Scenario, but not in ScenarioOutline.  I am using 1.2.5 version.

Issue #745, opened 2014-07-01
Don't integrate cucumber with Spring


DavidGangel:
Hi,

I mean why do we need to have Spring inside cucumber. Spring is a framework that can integrate with any of java class independently from the implementation.
The only thing that is needed the reporting stuff for JUnit but nothing else.
Cucumber makes a headache for me in the last couple of days/weeks when I was trying to create ContextHiererachy based on Parent and Child setup.
It just gives me a black hole how is it creating the contexts without any knowledge how to control it.
And of course there is a big chance that what is documented by spring won't work in the case of cucumber.

lunivore:
Hi @DavidGangel, sorry that the team was so busy nobody looked at your work! Is this still an issue for you? If so, please let us know and consider submitting a working pull request through Github as I notice your current codebase is failing the Travis build.

Otherwise if I don't hear back from you I'll assume the problem is solved, as the codebase has moved on significantly, and I'll close this ticket in a week.

PR #746, opened 2014-07-02
Adding first version of the integration with SpringJUinitrunner.

DavidGangel:
I included the SpringJUnit4ClassRunner to the Cucumber code as addition.
I have one issue in the SpringCucumberFeatureRunner buildFeatureElementRunners() method.
childre.add() does requires SpringJUnit4ClassRunner as type.


mpkorstanje:
This pr is based on an rather old version of cucumber. It would have to be completely redone. 

I appreciate you've made the effort but between this and #745 I can't quite understand what you're trying to do. If this is still an issue for you feel free to start a new discussion.

Issue #741, opened 2014-06-26
@Transpose doesn't work with a single element


rmannibucau:
public void I_should_get(final @Transpose List<Foo> expected) works but public void I_should_get(final @Transpose Foo expected) doesn't. However it makes sense to use it for @Then steps.

Just needs to modify cucumber.runtime.table.TableConverter#convert to not throw throw new CucumberException("Not a Map or List type: " + type); but use type as item type then return a single value.

kalypzo:
Are there plans to get this working with Cucumber?
The documentation seems to show  it supports this, but it doesn't appear to work:
https://cucumber.github.io/api/cucumber/jvm/javadoc/cucumber/api/Transpose.html

Result is:
> cucumber.runtime.CucumberException: Not a Map or List type:

Here's an example of what I tried:

Feature.feature
```
Background:
  Given the following default values:
    | name   | Product A |
    | weight | 16 oz.    |
    | price  | 2.19      |
```

Product.java
```java
public class Product {
  private String name;
  private String weight;
  private Double price;
  
  // Getters and Setters below...
}
```
StepDef.java
```java
  @Given("^the following default values:$")
  public void the_following_default_values(@Transpose Product product) {
    // Code goes here...
  }
```

franklsm1:
This could be fixed changing the line mentioned below to be something like:
```
List<T> list = Collections.unmodifiableList((List<T>) xStream.unmarshal(reader));
```
and add a check that is something like:
```
return (list.size == 1) ? list.get(0) : list;
```

This will allow users to specify a step in my feature file like:
```
    Given a user with the following fields:
      | firstName   | lastName   | birthYear   |
      | Sean           | Franklin      | 1991          |
```
and a step definition like:
```
    @Given("^a user with the following fields:$")
    public void aUserWithTheFollowingFields(User user) {
        this.user = user;
    }
```
instead of what I currently need to do:
```
    @Given("^a user with the following fields:$")
    public void aUserWithTheFollowingFields(List<User> users) {
        this.user = users.get(0);
    }
```
This will not solve the issue with the Transpose doc showing that having the field names in the 1st column with the values in the 2nd since this route assumes the field names are in the first row, but allows a single row table to be mapped to an object.

 https://github.com/cucumber/cucumber-jvm/blob/45ee6743517968ee086f674f941334660bdc4f82/core/src/main/java/cucumber/runtime/table/TableConverter.java#L104

kenneth-campbell-work:
[...]

PR #776, opened 2014-09-14
#741 supporting @Transposes for several examples and item by item eval

rmannibucau:


mpkorstanje:
We're currently working towards using cucumber-expressions. Cucumber expressions has a big impact on how data will be transformed. If this functionality were to be added now it would be very short lived.

serzhshakur:
@mpkorstanje can you please be more specific on this thing, namely when to expect those changes and in which shape?

mpkorstanje:
Check https://docs.cucumber.io/cucumber-expressions/ for an explanation on cucumber expressions. Tables can be expected to work in the same way.

The work on cucumber is mostly done by volunteers. As such there is no schedule or estimate.

PR #799, opened 2014-11-14
Cucumber is ready to run in OSGi containers

HendrikSP:
I have produced 2 commits that allow to run cucumber in OSGi containers.

* The first one adds OSGi headers to core, java and JUnit bundles.
* The second changes the loading of the USAGE.txt in RuntimeOptions, because due to OSGi class loading cause problems. I wrote more details in the commit message.

*NOTE*: Please release cucumber-jvm-deps with my other pull request (see https://github.com/cucumber/cucumber-jvm-deps/pull/3), because this changes make only sense when the jvm-deps bundle has also OSGi headers.

Thank you!

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -86,6 +86,7 @@
                 <artifactId>maven-jar-plugin</artifactId>
                 <configuration>
                     <archive>
+                        <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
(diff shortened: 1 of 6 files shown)


ryayak1460:
Any word on integrating this? I'd like to see the `cucumber-osgi` Maven dependency stood up for use.

HendrikSP:
I have created an updated rull request on this (#873)

PR #805, opened 2014-12-04
Fix to allow lein test to to run successfully

chrishowejones:
Fixed core.clj and associated deps in project to allow (run-cukes) to execute from repl and 'lein test' to succeed.


aslakhellesoy:
LGTM except that the latest version has moved. When this is merged we should update CONTRIBUTING.md to add a reminder to update this when we release.

brasmusson:
The reference to cucumber-core is updated in https://github.com/cucumber/cucumber-jvm/commit/270f88a28b17d1a700563a70d872da00acbe6014. For a reminder to update it in Contributing.md see #879.

PR #809, opened 2014-12-12
added support for execution order for "Before" and "After" hook. closes #807

### LINKED ISSUE #807: "order" for Hooks in Groovy.
Apparently there is no way to set execution order for hooks in groovy. 
In java implementation you can set the execution order in annotation  ```@Before(order = 100)``` . but that's not the case in groovy. 

Looking at "GroovyHookDefinition.groovy" class, one can see following code implemented for "getOrder" which is returning the order based on filename and not be user provided order.

```
    @Override
    public int getOrder() {
        return location.getFileName() == "env.groovy" ? -1 : 0;
    }
```

mhshams:
Since the original method was accepting varargs (```Object... args```) as parameter, there was almost no way to add new method with proper parameters, without breaking backward compatibility.

### DIFF
--- a/groovy/src/main/java/cucumber/api/groovy/Hooks.java
+++ b/groovy/src/main/java/cucumber/api/groovy/Hooks.java
@@ -8,28 +8,56 @@ import java.util.ArrayList;
 import java.util.List;
 
 public class Hooks {
+    private static final int DEFAULT_ORDER = 10000;
+    private static final long DEFAULT_TIMEOUT = 0;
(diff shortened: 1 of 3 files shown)


brasmusson:
I added checks for multiple timeout or order arguments (https://github.com/cucumber/cucumber-jvm/commit/c0a6c0f5f2df381e63c1903b1e8efa07b78d872a).

Thanks for your contribution @mhshams

Issue #343, opened 2012-06-13
feature: add timeout of Step and Scenario


k4124k:
timeout of step or scenario could be used for assert time of duration

PR #811, opened 2014-12-17
Timeout stops threads if they are uninterruptible.

aslakhellesoy:
If a thread does not respond to interrupt() we'll
try to stop() it after twice the specified timeout.

This uses the deprecated Thread.stop() method, but
for a testing tool like Cucumber that should be ok.

Ref #343.


mpkorstanje:
If we want to do things the junit way then it seems to me that the timeout is done at the wrong level. The timeout should be wrapping the calls in `Runner.runPickle`. This would also require that we wrap the `EventBus` to prevent slow PickleSteps from sending events after they've had a timeout and listen to step finished events to reset the timer.

Issue #789, opened 2014-10-24
URLOutputStream Should Check for Directory Existance


joshar1:
The maven surefire plugin offers the ability to [run tests in parallel](http://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html) with the parallel parameter.

While attempting to execute my Cucumber tests in parallel, I noticed that the URLOutputStream class was throwing exceptions because the _ensureParentDirExists_ method was not checking that the directory already existed prior to attempting to create the directory.

I propose the following one line change.

From:

``` java
if (file.getParentFile() != null && !file.getParentFile().isDirectory()) {
```

To:

``` java
if (file.getParentFile() != null && !file.getParentFile().isDirectory() 
     && !file.getParentFile().exists()) {
```

If this change is acceptable to others, I'll fork and issue a pull request.

mpkorstanje:
Hey @joshar1 I'm cleaning up old tickets. Do you still experience this problem?

PR #815, opened 2014-12-28
Additional unit-tests for #789

klausbayrhammer:
* tests whether files are created when the parent directory does not exists
* removes duplicate test case


aslakhellesoy:
LGTM!

PR #817, opened 2015-01-04
TestNG: run each feature as separate TestNG test

brasmusson:
This PR is a copy of #653, rebased on the current tip of master (and with some [comments](https://github.com/cucumber/cucumber-jvm/pull/653#issuecomment-32134204) fixed).

The idea is to let cucumber-testng run each Cucumber Feature as a separate TestNG test case (instead of wrap all the Cucumber Features into one TestNG test case). Then the notifications and/or reports generated natively by the TestNG runner will have results for each Cucumber Feature instead of only one result for the execution as a whole, see #653 for further details.

Closes #653.


aslakhellesoy:
LGTM!

PR #816, opened 2015-01-03
Added a test example for Android Studio.

mdzyuba:
Hello,

I've exported an android example - cukeulator and cukeulator-test into an Android Studio project and modified build.gradle to use cucumber-jvm. 

I've created a README.md and provided information on how to build and run tests.

Hope this will save time and help other cucumber-jvm and Android Studio users.

Thank you,

Mykola

PR #795, opened 2014-11-02
Problem using android-cucumber with newer version of the android-maven-plugin

brasmusson:
Newer versions of the android-maven-plugin does not seem to allow apklib:s to be included when building an instrumentation test debug apk.

With android-maven-plugin v4.0.0-rc.1 (and AFAIK v3.9.0-rc3), the generate source phase for a project that depend on cucumber-android will fails since cucumber-android is not extracted ([see Travis build](https://travis-ci.org/cucumber/cucumber-jvm/jobs/39746939)):
```
[INFO] --- android-maven-plugin:4.0.0-rc.1:generate-sources (default-generate-sources) @ cucumber-android-test ---
[INFO] Extracting libs
[ERROR] Error when generating sources.
org.apache.maven.plugin.MojoExecutionException: AndroidManifest.xml file wasn't found in next place:  /home/travis/build/cucumber/cucumber-jvm/examples/android/android-test/target/unpacked-libs/ic_cucumber-android
```
With android-maven-plugin v3.9.0-rc.1 cucumber-android is extracted, and its classes are included in the built artefact ([see Travis build](https://travis-ci.org/cucumber/cucumber-jvm/jobs/39747984)):
```
[INFO] --- android-maven-plugin:3.9.0-rc.1:generate-sources (default-generate-sources) @ cucumber-android-test ---
[INFO] Manifest copying disabled. Using default manifest only
[INFO] Extracting libs
[INFO] Extracting apklib cucumber-android...
```

PR #819, opened 2015-01-05
clean up of android project and examples

SierraGolf:
the generation of an apklib is not necessary and it leads to dirty workarounds when using gradle, see [PR816](https://github.com/cucumber/cucumber-jvm/pull/816/files#diff-bac00f7b61a0210f30b1d211ddaf2156R12)

with this change maven deployments (install/deploy) will just install a jar-file to the respective repositories.


brasmusson:
If the generation of the apklib is removed, then also the examples using cucumber-android ([android-test](https://github.com/cucumber/cucumber-jvm/blob/6fce323d53185ead992cb8789bf554d8189bb99f/examples/android/android-test/pom.xml#L22), [cukeulator-test](https://github.com/cucumber/cucumber-jvm/blob/6fce323d53185ead992cb8789bf554d8189bb99f/examples/android/cukeulator-test/pom.xml#L22)), need to not declaring that dependency as an apklib.

When not generating an apklib, is there any need for the [AndroidManifest.xml](https://github.com/cucumber/cucumber-jvm/blob/6fce323d53185ead992cb8789bf554d8189bb99f/android/src/main/AndroidManifest.xml) file?

SierraGolf:
It seems that the examples are currently broken, at least for me. I will look into that and clean up the apklib references and the AndroidManifest.xml.

SierraGolf:
I added the following changes:

* removed android maven plugin and related android files from cucumber-android project
* fixed broken example projects
* bumped android-maven-plugin version to 4.0.0

brasmusson:
Have you actually managed to run the android-test and cukeulator-test using android-maven-plugin v4.0.0, I run into [this problem](https://github.com/simpligility/android-maven-plugin/issues/548)? As long as cucumber-android was an apklib project, v4.0.0(-rcN) failed already on not unpacking the apklib. To actually be able or run the examples (on device or emulator), I had to downgrade to v3.9.0-rc1 (in #795, where I took the smaller change of declaring the now non-standard locations on files in [examples/android/pom.xml](https://github.com/brasmusson/cucumber-jvm/blob/6e7b7d19f37f9061b7121748504530787b964b23/examples/android/pom.xml#L54-L57)). #795 also run the android examples on Travis before deploying the cucumber-android snapshot. Just now a merged in the removal of the apklib generation (but not the latest commits on this PR).

SierraGolf:
Yeah, as you can see in the configuration I hacked it. Except for that small hack I was able to run everything now.

pom.xml
```diff
@@ -796,9 +796,9 @@
                 <!-- Non-standard plugins -->
 
                 <plugin>
-                    <groupId>com.jayway.maven.plugins.android.generation2</groupId>
+                    <groupId>com.simpligility.maven.plugins</groupId>
```
brasmusson: Does the android-maven-plugin need to be declared here at all? Previously v3.9.0-rc3 was declared here, but it was actually v4.0.0-rc.1 that was used, and that did not seems to make any difference.

android/wait_for_emulator
```diff
@@ -1,17 +0,0 @@
-#!/bin/bash
```
brasmusson: This file could instead be moved to examples/android. Now that [I now how to run the android examples on Travis](https://github.com/brasmusson/cucumber-jvm/blob/02c45303bb8825da6fb160d3da67ea6653b48c7d/.travis.yml#L10-L33) (with appropriate content in [.travis_android_update](https://github.com/brasmusson/cucumber-jvm/blob/02c45303bb8825da6fb160d3da67ea6653b48c7d/.travis_android_update)), it feels pretty interesting to do that - to avoid that the examples suddenly has stopped working.

pom.xml
SierraGolf: you are right, we don't need it here, since we only have it as a dependency in the android examples now. I will remove it.

android/wait_for_emulator
SierraGolf: ok, I will move it.

brasmusson:
AFAIK, the android module was put in a separate maven profile because to build it, you need to have an android-sdk installed. Now that cucumber-android is built like any other jar, the android module could be moved from the android profile to the common module. The the snapshot release of cucumber-android would be deployed together with the other module by the "non-android" oraclejdk7 job on Travis, and the "android" oraclejdk7 job could be used to build the android examples (similar to the openjdk7 job on Travis, which build the other examples).

SierraGolf:
Sounds good to me, do you want me do amend these changes to this pull request?

brasmusson:
When I experimented with moving the android module from a separate profile, the android job on Travis (now only executing the android examples) [failed](https://travis-ci.org/cucumber/cucumber-jvm/builds/46112221). The log contained the error `Killed emulator -avd test -no-skin -no-audio -no-window`, the Travis docs indicate that the process was killed due to to much resource usage. The handful times that job has run tests on the version of that emulator previously it worked. The fact that it might require some experimentation to find a stable configuration to run the android examples on Travis, could be an argument not to delay this PR until that is done.

SierraGolf:
I think we should do the following:

1. merge this pull request, because it fixes the examples (even though there is currently no ci to proof it)
2. create a new pull request which kills the android profile and puts the android module in the "default" profile, which makes the release easier
3. create a new pull request which adds the examples to ci, which might not be possible in a reliable way since the whole android support on travis is still [beta](http://docs.travis-ci.com/user/languages/android/)

what do you think?

brasmusson:
I think that is a good strategy. It is definitely better to use the build in android support, rather as we do no "manually" install the android sdk on Travis.

aslakhellesoy:
Great work on this @SierraGolf and @brasmusson. Eager to see this improve!

SierraGolf:
So guess I can merge this pull request now?

aslakhellesoy:
Go for it @SierraGolf. Can you add a note in History.md to summarise the changes?

SierraGolf:
Sure, done! Anything against also merging https://github.com/cucumber/cucumber-jvm/pull/820 ?

PR #824, opened 2015-01-13
Run the Android examples on Travis

### LINKED ISSUE #578: Run Android examples on Travis
The test-projects of the Android examples should be run on Travis.

.travis.yml and pom.xml need to be changed accordingly.

**Update:** An Android emulator virtual device probably has the be started via Maven as well.

brasmusson:
Use the new Android [support](http://docs.travis-ci.com/user/languages/android/) (still in beta) to run the Android examples on Travis

[X] Create "walking skeleton" Travis job for the Android examples
[X] Update the job to build the Android examples
[X] Update the job to run the examples on emulator
[ ] Update the job to also run the Android Studio example

Closes #578.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,12 +16,13 @@ matrix:
         - platform-tools
         - extra-android-support
         - android-21
-        - sys-img-armeabi-v7a-android-21
+        - android-15
+        - sys-img-armeabi-v7a-android-15
     script: mvn install -P android-examples -pl examples/android,examples/android/android-test -am -amd -Dno.gem.deploy=true -Dandroid.device=test
     install: mvn install -P android-examples -pl examples/android,examples/android/android-test -am -amd -DskipTests=true -Dmaven.javadoc.skip=true -B -V
     # Emulator Management: Create, Start and Wait
-    before_script:
-      - echo no | android create avd --force -n test -t android-21 --abi armeabi-v7a
+    before_install:
+      - echo no | android create avd --force -n test -t android-15 --abi armeabi-v7a
       - emulator -avd test -no-skin -no-audio -no-window &
       - android-wait-for-emulator
   - jdk: oraclejdk8
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,8 @@ matrix:
     before_install:
       - echo no | android create avd --force -n test -t android-15 --abi armeabi-v7a
       - emulator -avd test -no-skin -no-audio -no-window &
-    # Download the maven plugins and project dependencies before the script runs
-    install: mvn verify -P android-examples -pl examples/android,examples/android/android-test -am -amd dependency:go-offline
+    # Build the Android example dependencies, without test, to install dependencies
+    install: mvn verify -pl android,picocontainer -am -DskipTests=true -Dmaven.javadoc.skip=true -B -V
     # Emulator Management: Wait until the emulator is ready
     before_script:
       - android-wait-for-emulator
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,13 +18,18 @@ matrix:
         - android-21
         - android-15
         - sys-img-armeabi-v7a-android-15
-    script: mvn install -P android-examples -pl examples/android,examples/android/android-test -am -amd -Dno.gem.deploy=true -Dandroid.device=test
-    install: mvn install -P android-examples -pl examples/android,examples/android/android-test -am -amd -DskipTests=true -Dmaven.javadoc.skip=true -B -V
-    # Emulator Management: Create, Start and Wait
+    # Emulator Management: Create and start up in parallel with the install
+    # An old emulator is used to use less resources
     before_install:
       - echo no | android create avd --force -n test -t android-15 --abi armeabi-v7a
       - emulator -avd test -no-skin -no-audio -no-window &
+    # Download the maven plugins and project dependencies before the script runs
+    install: mvn verify -P android-examples -pl examples/android,examples/android/android-test -am -amd dependency:go-offline
+    # Emulator Management: Wait until the emulator is ready
+    before_script:
       - android-wait-for-emulator
+    # Build and test the Android examples
+    script: mvn verify -P android-examples -pl examples/android,examples/android/android-test -am -amd -Dandroid.device=test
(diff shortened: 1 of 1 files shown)


SierraGolf:
Hey @brasmusson I think we can close https://github.com/cucumber/cucumber-jvm/issues/578 in favor for this pull request, WDYT?

brasmusson:
@SierraGolf I think we should rather make this PR close the issue #578 when this PR is merged is merged (I have edited the description to do that).

SierraGolf:
nice!

.travis.yml
```diff
@@ -7,6 +7,23 @@ matrix:
...
+    env: ANDROID_EXAMPLES=true
+    language: android
+    android:
+      components:
+        - build-tools-21.1.1
+        - platform-tools
+        - extra-android-support
+        - android-21
+        - sys-img-armeabi-v7a-android-21
+    script: mvn install -P android-examples -pl examples/android,examples/android/android-test -am -amd -Dno.gem.deploy=true -Dandroid.device=test
```
SierraGolf: do we need the ```-Dno.gem.deploy=true``` I think this is unnecessary.

.travis.yml
```diff
@@ -7,6 +7,23 @@ matrix:
...
+    install: mvn install -P android-examples -pl examples/android,examples/android/android-test -am -amd -DskipTests=true -Dmaven.javadoc.skip=true -B -V
```
SierraGolf: why is this step necessary? the script already installs the artifacts and runs the tests.

.travis.yml
```diff
@@ -7,6 +7,23 @@ matrix:
   include:
   - jdk: oraclejdk7
     script: mvn -q deploy --settings .travis-settings.xml -Dno.gem.deploy=true
+  - jdk: oraclejdk7
+    env: ANDROID_EXAMPLES=true
```
SierraGolf: what is the reason for this environment variable?

.travis.yml
```diff
@@ -7,6 +7,23 @@ matrix:
...
+    language: android
+    android:
```
SierraGolf: any particular reason why the steps are not in chronological order (install, before_script, script)?

.travis.yml
brasmusson: It most likely is, I just did not remove it (yet).

.travis.yml
brasmusson: The intent of the environment variable is to make it obvious on the Travis build page, which of the oraclejdk7 jobs that is running the Android examples (and which is building, testing and deploying the Cucumber-JVM modules). It actually is not used in the execution of the job.

.travis.yml
brasmusson: No, they are ordered as the are for no good reason at all (from copying the definition of the old Android job mostly).

.travis.yml
brasmusson: If there is no explicit `install`step, Travis will execute `mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V` as the `install` step, to make sure that the projects all dependencies have been installed with maven before the `script` command starts (if it detects that is is an maven project). Hm, now the [Travis docs](http://docs.travis-ci.com/user/languages/java/#Dependency-Management) specifies a slightly different command. 

In this case building (but not testing) all the Cucucmber-JVM modules (but not the Android examples), would be a lot of unnecessary work, so here I rather have the same command as the `script` command as `install` command - with the addition of `-DskipTests=true -Dmaven.javadoc.skip=true -B -V`. In this case `integration-test` or `verify` (should there be some extra checks in that step), should work just as well as `install` as the maven target (but why not use `mvn install`, when Travis would use it as default in the `install` step). I don't think `clean` adds anything as the whole job is executed in a fresh sandbox.

.travis.yml
SierraGolf: I see, it would be nice if travis would show the language in the matrix so these hacks wouldn't be necessary.

.travis.yml
SierraGolf: I did not know that travis defaults to that, however I think this is a highly inefficient way of implementing a build pipeline and from the maven point of view actually extremely weird as well.

I wonder why travis would do that, considering they actually pay for more execution time.

Anyway so you are saying that if we would to leave out the install step travis would run it anyway but instead build the whole project since it does not use the correct profiles and locations? If that's the case, then I would recommend skipping the install step explicitly like that ```install: true```. I am so focused on performance, because when we work with emulators we should try to improve every bit of the build pipeline.

Using ```mvn verify``` instead of ```mvn install``` also saves some time by avoiding unnecessary maven steps. I agree that ```clean``` is unnecessary.

brasmusson: Most often maven would detect that the build artefacts is already up to date the second time and not rebuild, so the idea is probably that it should not add so much time to run maven once (without tests) before the actual script phase. For instance [this build](https://travis-ci.org/cucumber/cucumber-jvm/jobs/46895638), where test are not run neither the first nor second time, maven list the build time for cucumber-core as 49s the first time, and 3s the second time. In case of building Android module with the android-maven-plugin, it looks as if everything is rebuild the second time, and then you definitely pay for a lot more execution time.

.travis.yml
```diff
@@ -7,6 +7,29 @@ matrix:
...
+      components:
+        - build-tools-21.1.1
+        - platform-tools
+        - extra-android-support
+        - android-21
+        - android-15
+        - sys-img-armeabi-v7a-android-15
+    # Emulator Management: Create and start up in parallel with the install
+    # An old emulator is used to use less resources
```
SierraGolf: clever!

.travis.yml
```diff
@@ -7,6 +7,29 @@ matrix:
...
+    before_install:
+      - echo no | android create avd --force -n test -t android-15 --abi armeabi-v7a
+      - emulator -avd test -no-skin -no-audio -no-window &
+    # Build the Android example dependencies, without test, to install dependencies
+    install: mvn verify -pl android,picocontainer -am -DskipTests=true -Dmaven.javadoc.skip=true -B -V
```
SierraGolf: yeah, that makes more sense. now we are really installing dependencies before running.

mpkorstanje:
Resolved by #1181.

PR #828, opened 2015-01-18
Make the Summary Printer into a plugin

brasmusson:
This PR turns the summary printing into a plugin, [as discussed on the mailing list](https://groups.google.com/forum/#!searchin/cukes/[JVM]$20Control$20over$20the$20error$20summary$20printing./cukes/BZaKvrmdH9c/CnMPzWHAnKQJ). Two built-in summary printer plugins are created,
 - `default_summary` which basically gives the same summary output as today, and 
 - `null_summary` which gives no output at al

Inspired by Cucumber-Ruby, the `default_summary` is extended (compared to the current behaviour), with the listing of failed scenarios. Or to be precise the listing of the scenarios that makes the exit code non-zero, that is in strict mode also undefined and pending scenarios are listed.

The different supported plugin types are separated into different categories internally, to be able to detect when to add default formatter. The [current code](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/cucumber/runtime/RuntimeOptions.java#L84-L85) is not entirely correct, if a StepDefinitionsReporter is added when the runtime options first is parsed (and not Formatter/Reporter plugin is also added), no Formatter/Reporter will exist during the execution.

### DIFF
--- a/core/src/main/java/cucumber/api/SummaryPrinter.java
+++ b/core/src/main/java/cucumber/api/SummaryPrinter.java
@@ -1,0 +1,5 @@
+package cucumber.api;
+
+public interface SummaryPrinter {
+    public void print(cucumber.runtime.Runtime runtime);
+}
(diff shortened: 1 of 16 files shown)


aslakhellesoy:
LGTM!

PR #829, opened 2015-01-18
Move the deployment to the after_success phase on Travis

brasmusson:
The benefits are that only if the test for all modules passes are the modules deployed, and that the tests for all modules are also executed on Java8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,7 @@ jdk:
 matrix:
   include:
   - jdk: oraclejdk7
-    script: mvn -q install
-    after_success: mvn -q deploy --settings .travis-settings.xml -Dno.gem.deploy=true -DskipTests=true -Dmaven.javadoc.skip=true
+    script: mvn -q deploy --settings .travis-settings.xml -Dno.gem.deploy=true -DdeployAtEnd=true
   - jdk: oraclejdk8
     script: mvn -q install -P java8
     after_success: mvn -q deploy -P java8 -pl java8 --settings .travis-settings.xml -DskipTests=true -Dmaven.javadoc.skip=true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,18 @@
 ---
 language: java
+# skip the default install (mvn install -DskipTests=true) to speed up the build
+install: true
 script: mvn -q install -P examples
 jdk:
 - openjdk7
 matrix:
   include:
   - jdk: oraclejdk7
-    script: mvn verify
-    after_success: mvn deploy --settings .travis-settings.xml -Dno.gem.deploy=true -DskipTests=true -Dmaven.javadoc.skip=true
+    script: mvn -q install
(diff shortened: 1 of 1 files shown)


SierraGolf:
I know it is experimental, but are you aware of the [deployAtEnd]( http://maven.apache.org/plugins/maven-deploy-plugin/deploy-mojo.html#deployAtEnd) flag?

brasmusson:
I did not know about the deplyAtEnd flag. It is an option for the Java7 job, where all modules built are also deployed, but not for the Java8 job, where all modules are built and tested, but only the java8 module is deployed.

.travis.yml
```diff
@@ -1,14 +1,21 @@
...
+# skip the default install (mvn install -DskipTests=true) to speed up the build
+install: true
 script: mvn -q install -P examples
 jdk:
 - openjdk7
 matrix:
   include:
   - jdk: oraclejdk7
-    script: mvn -q deploy --settings .travis-settings.xml -Dno.gem.deploy=true
+    # skip the default install (mvn install -DskipTests=true) to speed up the build
```
SierraGolf: isn't that inherited from above? btw. on java8 you just wrote the comment without the actual skipping

SierraGolf:
Yeah, you are right about the java 8 job. Btw. maybe it is better to use one job (manually calling the ```jdk_switcher```) instead of a matrix in order to make sure that all artifacts are deployed only when their dependent artifacts have been deployed.

as you said the java 8 job only deploys that one artifact, but it can not be sure that its dependent artifacts have been deployed.

.travis.yml
brasmusson: You are right, it is inherited from above, because that oraclejdk8 job did execute `true` for install (even though just the comment was specified on the job itself).

brasmusson:
Actually, the big improvement I saw when starting out on this PR, was to not only build (which the `install` step did), but also run the test for all the modules on Java8. If using the `after_success` there for that, it could also be used in the oraclejdk7 job, and ...

I don't know how important it is, how much effort to spend, to make sure the modules only deployed after tests for all modules have passed, and all dependent artefacts have been deployed. Deploying all modules from the Java8 job, would be one solution to that.

SierraGolf:
I don't see much effort there, you basically just put a few calls in one script, but I would agree that it might be overkill to do these changes for yet theoretical problems. However I think deploying all artifacts from java 8 might cause real problems. I think this would make all modules only work on java 8, or am I wrong?  I think you would get the ```Unsupported major.minor version 52.0``` error.

brasmusson:
[This declaration in the pom](https://github.com/cucumber/cucumber-jvm/blob/master/pom.xml#L680) should make the modules (except the java8 module, of course) to work on Java6 and later, so it should be possible to deploy all modules from the Java8 job, if it is worth while to switch it from the Java7 job.

SierraGolf:
ah, ok.

PR #839, opened 2015-02-23
Fix non running gradle example

judoole:
Would not run out of the box because of:
- missing dependency 1.2.1 in maven-repo, uses 1.2.2 now
- Wrong Cucumber.Options import

Fixed use so --plugin is used instead of --format


aslakhellesoy:
Takk kompis!

judoole:
:+1: :smile:

PR #842, opened 2015-03-15
Call all formatters, also when using the IntelliJ's internal formatter.

### LINKED ISSUE #803: Exception in one formatter prevents other formatters from working
I have encountered this when trying to run RerunFormatter in cucumber-jvm 1.2.0 together with IntelliJ's org.jetbrains.plugins.cucumber.java.run.CucumberJvmSMFormatter. If the IntelliJ formatter is added before the RerunFormatter, the RerunFormatter does not work. The problem is that CucumberJvmSMFormatter does not support new Formatter methods like  startOfScenarioLifeCycle and endOfScenarioLifeCycle. These methods are invoked in a loop for each plugin (formatter) by the pluginProxy method in RuntimeOptions so if one plugin throws an exception - the code exits from the loop and the methods of plugins after the offending one are not called.
Also in the CucumberScenario class's 'run' method the exception from startOfScenarioLifeCycle and endOfScenarioLifeCycle are caught silently which makes it very difficult to understand what is happening and why other plugins are not working:

    cucumber.runtime.model.CucumberSenario

    try {
            formatter.startOfScenarioLifeCycle((Scenario) getGherkinModel());
        } catch (Throwable ignore) {
            // IntelliJ has its own formatter which doesn't yet implement this.
        }

The solution would probably be to catch the exceptions in pluginProxy loop so that all plugins have their methods invoked, or at least provide the output from the exception so that the problem could be detected more easily.

brasmusson:
The IntelliJ's internal formatter does not implement the startOfScenarioLifeCycle and the endOfScenarioLifeCycle methods. The current handling of this has the side effect, that all other formatters (after the IntelliJ one in the plugin list), will not receive calls to these methods, see #803.

This PR changes the handling of exceptions from the the startOfScenarioLifeCycle and the endOfScenarioLifeCycle methods of formatters, so that also in this case all formatters will receive calls to these methods.

Fixes #803.

### DIFF
--- a/core/src/main/java/cucumber/runtime/RuntimeOptions.java
+++ b/core/src/main/java/cucumber/runtime/RuntimeOptions.java
@@ -240,7 +240,14 @@ public class RuntimeOptions {
             public Object invoke(Object target, Method method, Object[] args) throws Throwable {
                 for (Object plugin : getPlugins()) {
                     if (type.isInstance(plugin)) {
-                        Utils.invoke(plugin, method, 0, args);
+                        try {
+                            Utils.invoke(plugin, method, 0, args);
+                        } catch (Throwable t) {
+                            if (!method.getName().equals("startOfScenarioLifeCycle") && !method.getName().equals("endOfScenarioLifeCycle")) {
+                                // IntelliJ has its own formatter which doesn't yet implement these methods.
+                                throw t;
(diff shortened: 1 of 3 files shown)


aslakhellesoy:
LGTM!

PR #845, opened 2015-03-31
extended the scenario interface by exposing the results from all steps.

danielnpu:
This change will make the results of each step in a scenario available for the caller. It is very useful when the scenario hooks needs to access the result of each step.

### DIFF
--- a/core/src/main/java/cucumber/api/Scenario.java
+++ b/core/src/main/java/cucumber/api/Scenario.java
@@ -1,6 +1,9 @@
 package cucumber.api;
 
+import gherkin.formatter.model.Result;
+
 import java.util.Collection;
+import java.util.List;
 
 /**
  * Before or After Hooks that declare a parameter of this type will receive an instance of this class.
@@ -56,4 +59,9 @@ public interface Scenario {
      * @return the id of the Scenario.
      */
     String getId();
+
+    /**
+     * @return a list of results for each step of the Scenario.
+     */
+    List<Result> getStepResults();
 }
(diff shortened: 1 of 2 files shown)


danielnpu:
This change will make the results of each step in a scenario available for the caller. It is very useful when the scenario hooks needs to access the result of each step.

aslakhellesoy:
Can you give an example of why a hook would need to know the status of individual steps?

danielnpu:
```java
    @After
    public void logSummary(Scenario scenario) {
        String name = scenario.getName();
        String id = scenario.getId();

        String logText = name + ": " + scenario.getStatus() + System.lineSeparator();

        List<Result> stepResults = scenario.getStepResults();
        for (Result result: stepResults) {
            if (result.getStatus().equals("pending")) {
                logText += result.getError() + System.lineSeparator();
            } else if (result.getErrorMessage() != null) {
                logText += result.getErrorMessage() + System.lineSeparator();
            }
        }

        String fileName = briefSummaryDirectory + File.separator + id;

        try(PrintWriter out = new PrintWriter(new BufferedWriter(new FileWriter(fileName, false)))) {
            out.println(logText);
        } catch (IOException e) {
            logger.error(e.getMessage(),e);
        }
    }

```

dkowis:
Shouldn't a reporter show you which steps failed, and why they failed?

danielnpu:
Thanks for hint @dkowis ! I managed to use a reporter to achieve the same goal.

Have a happen Easter!

dkowis:
@danielnpu Cool! I'ma close the pull request, re-open it if you still have things you want to do here.

PR #847, opened 2015-04-06
Update Java8StepDefinition.java

davidcoelho:
I have Arity mismatch error when using Java8 step definitions.
Looking at class Java8StepDefinition, I saw that could be solved by returning parameterCounter like JavaStepDefinition actually does. Simple:  return parameterInfos.size();

### DIFF
--- a/java/src/main/java/cucumber/runtime/java/Java8StepDefinition.java
+++ b/java/src/main/java/cucumber/runtime/java/Java8StepDefinition.java
@@ -65,7 +65,7 @@ public class Java8StepDefinition implements StepDefinition {
 
     @Override
     public Integer getParameterCount() {
-        return pattern.matcher("").groupCount();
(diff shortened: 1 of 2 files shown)


davidcoelho:
To get the error, I have a feature and definition step like this:

myfeature.feature
....
    Given I have a list of numbers:
      | 11 |
      | 3  |
      | 7  |

MyStepsDefinition.java
public class MyStepsDefinition implements En {
    public MyStepsDefinition() {
        Given("^I have a list of numbers:$", (DataTable t) -> {
            List<Integer> numbers = t.asList(Integer.class);
            this.calculadoraAvancada = new CalculadoraAvancada(numbers);
        });
        .....

aslakhellesoy:
Can you please add a test?

davidcoelho:
Yes. I added right now.

PR #853, opened 2015-04-13
[Needle] Handle circular dependencies

lbilger:
The problem is best shown by the changes in the test files. Depending on the order in which objects are instantiated by the factory (which seems to be arbitrary), the @NeedleInjectionProvider annotation may work for other classes or not.


mpkorstanje:
Very sensible PR. Looks good. 

A belated thank you for your contribution!

PR #859, opened 2015-04-27
Allow the ability to have failed scenarios retried immediately by passin...

rundmd:
...g in a value for the 'retryScenario' parameter via @CucumberOptions:

@CucumberOptions(
                format={"pretty", "html:target/cucumber"},
                features={"src/test/resources"},
                retryScenario={"2"}
        )

Failures are listened for at the scenario level by passing a FeatureResultListener and retry value to CucumberFeature.run

### DIFF
--- a/core/src/main/java/cucumber/api/CucumberOptions.java
+++ b/core/src/main/java/cucumber/api/CucumberOptions.java
@@ -64,4 +64,9 @@ public @interface CucumberOptions {
      * @return what format should the snippets use. underscore, camelcase
      */
     SnippetType snippets() default SnippetType.UNDERSCORE;
+
+    /**
+     * @return the number of times a scenario should be retried
+     */
+    String[] retryScenario() default {};
 }
--- a/core/src/main/java/cucumber/runtime/RuntimeOptionsFactory.java
+++ b/core/src/main/java/cucumber/runtime/RuntimeOptionsFactory.java
@@ -7,6 +7,7 @@ import cucumber.runtime.io.MultiLoader;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.lang.Integer;
 
 import static java.util.Arrays.asList;
 
@@ -15,6 +16,8 @@ public class RuntimeOptionsFactory {
     private boolean featuresSpecified = false;
     private boolean glueSpecified = false;
     private boolean pluginSpecified = false;
+    private int retry;
+
 
     public RuntimeOptionsFactory(Class clazz) {
         this.clazz = clazz;
@@ -40,6 +43,7 @@ public class RuntimeOptionsFactory {
                 addSnippets(options, args);
                 addGlue(options, args);
                 addFeatures(options, args);
+                addRetryScenario(options);
             }
         }
         addDefaultFeaturePathIfNoFeaturePathIsSpecified(args, clazz);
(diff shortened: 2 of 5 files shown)


aslakhellesoy:
I presume you want to add this feature to make flickering scenarios pass more often.

I think this feature would be harmful. We shouldn't give users a way to sweep their problems under the rug.

Am I missing something here?

rundmd:
Yes, this would allow flickering scenarios to pass more often.  Retrying failed tests without carefully investigating the reason for the failure could lead to a bug that's intermittent in nature being introduced to production.  If there's a consistent and reproducible bug, it should still fail after the retry, whether or not the retry happens immediately after the first failure, or when the rerun file is executed after the test suite completes.  Users can still rerun the failed scenarios without investigating the root cause(s).  Users in my situation could benefit from this feature.   My team is working with a UI in an environment that is prone to latency.  We spend more time than we should investigating failures for missing links, buttons, and data that almost always appears when we check the failure and manually run the scenario.  We've forced our tests to sleep but we can't guarantee the delay is long enough without using a large value, and we would rather not have our tests wait for an unnecessarily long time.

Sorry for the long response.

aslakhellesoy:
There are better ways to deal with asynchronicity than sleeping. Selenium has `waitFor`. We've outlined some recipes in [our book](https://pragprog.com/book/srjcuc/the-cucumber-for-java-book), and I think the GOOS book has more.

Retrying a scenario because it is flickering does not address the underlying problem, and I'm not at all keen on giving people features that undermine what I consider an essential practice of BDD: Consistent behaviour.

WDYT @mattwynne?

rundmd:
Thanks for your insight.  We've tried using the following but still hit false positives:

WebElement someElement = (new WebDriverWait(driver, 30))   .until(ExpectedConditions.presenceOfElementLocated(By.id("someElement")));

We aren't in control of development and our problematic features "work" so time most likely won't be spent trying to improve performance.

mattwynne:
I can see how this is a useful feature, especially in an investigative phase where you want to try experiments to fix a flickering test without having it keep breaking your build at random.

I'd like to give this careful thought though. I fully agree about not encouraging (even implicitly) people to just sweep flickering tests under the rug. I also wonder how you could useful report results here, because a retried scenario that failed, say 99 out of 100 attempts, is very different to one that failed 1 out of 2 attempts. I'd want to see that kind of information collected and reported on, if this were going to be implemented.

brasmusson:
One approach would be to handle this problem by "programming above" Cucumber(-JVM), instead of adding a new feature to Cucumber(-JVM) itself. After running using:
```
@CucumberOptions(plugin={"json:report.json", "rerun:rerun.txt"})
```
the failed scenarios can be rerun adding this to the command line:
`-Dcucumber.options="--plugin json:retry.json @rerun.txt"`
or something like:
```
for "10 times" do
   "run with" -Dcucumber.options="--plugin json_${time}:retry.json @rerun.txt"
end
```
With all the json reports produced, the issue of making useful summary reporting remains, but it is handled outside of Cucumber(-JVM) and not inside it. 

This would be supported better if the "--plugin" option from the command line clobbered the plugins of the same category defined by `@CucumberOptions`, and the "--add-plugin" options was added to be able to add new plugins without clobbering as today. Otherwise the report.json from the original run (and rerun.txt) will be overwritten when rerunning the failed scenarios.

Still now that there are three plugin categories, formatter/reporter, summary and step definition reporter, using "--plugin <formatter>" would only clobber the formatter/reporter plugins from `@CucumberOptions`, and not the summary plugins or step definition reporter plugins (so "--add-plugin" would in reality not be necessary to be able to add a step definition reporter plugin without clobbering the defined formatter plugins).

rundmd:
I like the idea of reporting on the "repeat offenders" and can look into
adding some reporting.  Even if this doesn't become a feature, the
reporting would be something my team could leverage to improve testing.

Thanks

dkowis:
Since --plugin clobbering has been merged, I'm going to close this guy. I think that the rerunning of failed scenarios would better be handled outside of cucumber proper.

PR #860, opened 2015-04-28
Add --add-plugin option and clobber when --plugin is used

brasmusson:
When using --plugin the plugins of that type defined from a different source is clobbered. Using -Dcucumber.options="--plugin pretty" would clobber the formatter/reporter plugins defined in `@CucumberOptions`. --add-plugin does not clobber any plugins defined from a different
source.

The rational for this PR is both consistency, other list type options do clobber, and it enables some use cases not possible today:
* forcing one specific formatter to write to stdout on the CI system even when the users have specified another formatter to write to stdout in the `@CucumberOptions` ([mailing list post](https://groups.google.com/d/msg/cukes/8bZd6dWnSwc/9nF0st9bkmYJ))

### DIFF
--- a/core/src/main/java/cucumber/runtime/RuntimeOptions.java
+++ b/core/src/main/java/cucumber/runtime/RuntimeOptions.java
@@ -96,6 +96,7 @@ public class RuntimeOptions {
         List<Object> parsedFilters = new ArrayList<Object>();
         List<String> parsedFeaturePaths = new ArrayList<String>();
         List<String> parsedGlue = new ArrayList<String>();
+        ParsedPluginData parsedPluginData = new ParsedPluginData();
 
         while (!args.isEmpty()) {
             String arg = args.remove(0).trim();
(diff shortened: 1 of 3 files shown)


core/src/main/java/cucumber/runtime/RuntimeOptions.java
```diff
@@ -114,11 +115,11 @@ private void parse(List<String> args) {
                 parsedGlue.add(gluePath);
             } else if (arg.equals("--tags") || arg.equals("-t")) {
                 parsedFilters.add(args.remove(0));
-            } else if (arg.equals("--plugin") || arg.equals("-p")) {
-                addPluginName(args.remove(0));
+            } else if (arg.equals("--plugin") || arg.equals("--add-plugin") || arg.equals("-p")) {
```
dkowis: I'm not sure I like renaming the semantics of `-p` to be `--add-plugin` now. This could be a suprise break for people who are currently using `-p` and expect that behavior to be the same.

dkowis: Oh wait, you're changing the semantics of `--plugin` entirely! Hm. Same thought process applies. Since it's already released, should we use `--add-plugin` and have those semantics be different than the existing `--plugin`?

brasmusson: Between v1.1.4 and v1.1.5 the `--glue` and `feature` options was changed from "add" to clobber (https://github.com/cucumber/cucumber-jvm/commit/5e3b0ea7e359dff9dbda4a51296c1f00d0cd109e), so this kind of semantic change has been done on already released functionality before. And isn't it more surprising for people the `--plugin` behaves differently from all other list type options?

I know the rationale for the current behaviour of `--plugin` (https://github.com/cucumber/cucumber-jvm/pull/643#issuecomment-47227892), but do think it makes thinks unnecessary hard on users that it is impossible to remove plugins defined in the `@CucumberOptions` annotation, without edit it in the source code. With this PR CucumberPro and others can use `--add-plugin` if they need the current behaviour, and the people the need the clobbering of plugins from the `@CucumberOptions` can do that too (and all list type options behave consistently).

dkowis:
That makes sense to me +1 for merging

PR #867, opened 2015-05-18
Handle doc string and data tables in a single step

davidegrohmann:


aslakhellesoy:
I don't think we want this. Please explain with a real example how this is an improvement.

davidegrohmann:
Hi @aslakhellesoy 

the use case is the following, consider a step like:

```
Given a doc string and a data table
     """
     some text here
     """
     | data  |
     | row1 |
```

I would expect it to match the following step implementation:
```
Given("""^a doc string and a data table$""") { (docString: String, data: DataTable) =>
     //    do something here
  }
```

at the moment the above step parses fine but the scenario fails at runtime with the following obscure error message:
```
cucumber.runtime.CucumberException: Arity mismatch: Step Definition 'GlueSteps.scala:79' with pattern [^a doc string and a data table$] is declared with 2 parameters. However, the gherkin step has 2 arguments [DocString:some text here, Table:[[data], [row1]]].
Step: Given a doc string and a data table
  at cucumber.runtime.StepDefinitionMatch.arityMismatch(StepDefinitionMatch.java:100)
  at cucumber.runtime.StepDefinitionMatch.transformedArgs(StepDefinitionMatch.java:60)
  at cucumber.runtime.StepDefinitionMatch.runStep(StepDefinitionMatch.java:37)
  at cucumber.runtime.Runtime.runStep(Runtime.java:298)
  at cucumber.runtime.model.StepContainer.runStep(StepContainer.java:44)
  at cucumber.runtime.model.StepContainer.runSteps(StepContainer.java:39)
  at cucumber.runtime.model.CucumberScenario.run(CucumberScenario.java:48)
  at cucumber.runtime.model.CucumberFeature.run(CucumberFeature.java:163)
  at cucumber.runtime.Runtime.run(Runtime.java:120)
  at cucumber.api.cli.Main.run(Main.java:36)
  at cucumber.api.cli.Main.main(Main.java:18)
```

This PR is trying to fix this problem by making sure that the above step definition matches the step implementation.  I am not aware of any documentation that forbids the usage of a doc string together with a data table, but I might be mistaken here.

Thanks in advance

aslakhellesoy:
What I'm after is a *real* example, not a made-up one. I understand what you're trying to achieve, I just don't understand why.

You have to convince us (with a real case, from the real world) that this would improve communication between non-technical stakeholders and developers/testers. My gut reaction tells me it would just be confusing and lead to poorly readable scenarios.

davidegrohmann:
My use case is to use features/scenarios to describe query behaviors in order to create a compatibility suite for the Cypher query language (http://neo4j.com/docs/stable/cypher-query-lang.html). The current implementation of this suite is on top of Neo4j (http://neo4j.com/).

I would like to write a scenario in the following way:

```
  Scenario predicate filtering
    Given init db:
        """
          CREATE
            (advertiser {name:"advertiser1"}),
            (thing      {name:"Color"}),
            (red        {name:"red"}),
            (p1         {name:"product1"}),
            (p2         {name:"product4"}),
            (advertiser)-[:adv_has_product]->(p1),
            (advertiser)-[:adv_has_product]->(p2),
            (thing)-[:aa_has_value]->(red),
            (p1)   -[:ap_has_value]->(red),
            (p2)   -[:ap_has_value]->(red)
        """
    When running parametrized query:
        """
          MATCH (advertiser)-[:adv_has_product]->(out)-[:ap_has_value]->(red)<-[:aa_has_value]-(a)
          WHERE id(advertiser) = {advertiserId} AND id(a) = {thingId}
          AND red.name = 'red' and out.name = 'product1'
          RETURN out.name AS name
        """
        | adveriserId | thingId |
        |           0 |       2 |
    Then result:
         | name     |
         | product1 |
```

which will read better than the following (due to the lack of support for new lines in steps):

```
  Scenario predicate filtering
    Given init db: CREATE (advertiser {name:"advertiser1"}), (thing {name:"Color"}), (red {name:"red"}),(p1 {name:"product1"}), (p2 {name:"product4"}), (advertiser)-[:adv_has_product]->(p1), (advertiser)-[:adv_has_product]->(p2), (thing)-[:aa_has_value]->(red), (p1)-[:ap_has_value]->(red), (p2)-[:ap_has_value]->(red)
    When running query: MATCH (advertiser)-[:adv_has_product]->(out)-[:ap_has_value]->(red)<-[:aa_has_value]-(a) WHERE id(advertiser) = {advertiserId} AND id(a) = {thingId} AND red.name = 'red' and out.name = 'product1' RETURN out.name AS name
        | adveriserId | thingId |
        |           0 |       2 |
    Then result:
         | name     |
         | product1 |
```

dkowis:
Is there really value in using cucumber for this? This is a very low level thing, you could accomplish easier with rspec I think. Cucumber is for helping you and the business communicate better. I feel like cucumber is just getting in the way with this much technical detail...

aslakhellesoy:
Every time someone puts SQL (or XML, or JSON, or CSS selectors, or CODE) in Cucumber Scenarios, God kills a kitten.

I like kittens, so I'm going to reject this. Also: https://cucumber.io/blog/2014/03/03/the-worlds-most-misunderstood-collaboration-tool

srbaker:
@aslakhellesoy you are correct in most cases, but not this one. We are building a database, and the query language is the way that our *users* interact with the product. This small patch will dramatically improve feature descriptions in precisely the ways our users interact with the product.

Reconsidering this will really help the way we describe certain bits of our application.

aslakhellesoy:
@srbaker fair enough. While the use cases for this are rare, I can't think of any serious downsides.

I can't reopen the PR though - @davidegrohmann deleted his fork!

davidegrohmann:
@aslakhellesoy I have created a new PR with this fix: https://github.com/cucumber/cucumber-jvm/pull/911

aslakhellesoy:
You *could* rewrite it like this:

```gherkin
Scenario predicate filtering
    Given init db:
        """
          CREATE
            (advertiser {name:"advertiser1"}),
            (thing      {name:"Color"}),
            (red        {name:"red"}),
            (p1         {name:"product1"}),
            (p2         {name:"product4"}),
            (advertiser)-[:adv_has_product]->(p1),
            (advertiser)-[:adv_has_product]->(p2),
            (thing)-[:aa_has_value]->(red),
            (p1)   -[:ap_has_value]->(red),
            (p2)   -[:ap_has_value]->(red)
        """
    And a parametrized query:
        """
          MATCH (advertiser)-[:adv_has_product]->(out)-[:ap_has_value]->(red)<-[:aa_has_value]-(a)
          WHERE id(advertiser) = {advertiserId} AND id(a) = {thingId}
          AND red.name = 'red' and out.name = 'product1'
          RETURN out.name AS name
        """
    When the query is run with:
        | adveriserId | thingId |
        |           0 |       2 |
    Then result:
         | name     |
         | product1 |
```

Changing Gherkin to support more than one [step argument](https://github.com/cucumber/gherkin/blob/master/gherkin.berp#L29) is a substantial amount of work - we have 8 or so implementations to maintain, not to mention the cucumber implementations depending on gherkin.

I'd be open for adding this if someone sends a PR to the gherkin project to support this though.

PR #868, opened 2015-05-25
Cucumber scala tweaks

dkowis:
Made some changes as to how the cucumber-scala artifacts are created. 

Specifically, now it's a hierarchy of maven poms, with a second aggregator used to build the individual projects, and the shared sources moved into a sources directory.

Additionally I'm using `pluginManagement` to avoid copy/pasta for the plugins that generate the I18n sources and the actual building of scala itself. The only dependencies listed in the scala_<version> projects are the scala versions themselves. This is far more scalable than "current" and "previous" we had before, and it makes it easier to know obviously which version of scala we're building against.

I also fixed one compile time flag, since it was talking about using scala 2.9 and we're only building for 2.10 and above. 

Finally, I added the milestone of scala 2.12 in there, because I wanted to see it build us an artifact for it :D


dkowis:
Hm, I just realized that the example project is using the parameter for the "scala current version"

Opinions on just using a 2.11.6 fixed value for it instead? Otherwise, I should have three example projects one for each scala version and I think that's silly.

paoloambrosio:
+1 Very neat!

Regarding the example, I had the same thought. Either the scala binary version (2.11) should be computed from scala.latest.version (and in Maven it's just overkill) or it should be independent. **I would hardcode 2.11.6 as you suggested**. At least if we forget to update the example pom it will be correct and in sync.

PR #889, opened 2015-07-24
add a exception in to pending exception as the org.junit.internal. is…

nablik:
as org.junit.internal.AssumptionViolatedException is depercated, I added the new line in the PENDING_EXCEPTION array [org.junit.AssumptionViolatedException]

### DIFF
--- a/core/src/main/java/cucumber/runtime/Runtime.java
+++ b/core/src/main/java/cucumber/runtime/Runtime.java
@@ -34,7 +34,8 @@ import java.util.Set;
 public class Runtime implements UnreportedStepExecutor {
 
     private static final String[] PENDING_EXCEPTIONS = new String[]{
-            "org.junit.internal.AssumptionViolatedException"
+            "org.junit.internal.AssumptionViolatedException",
+            "org.junit.AssumptionViolatedException"
     };
 
     static {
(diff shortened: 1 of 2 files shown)


aslakhellesoy:
Could you add a test to `RuntimeTest` please?

nablik:
Hi Aslak,
No itis impossible to write unit test now.
The reason is the CucumberCore use the Junit version 4.11
and the PR  is to cover the Junit 4.12.
And upgrading to 4.12 will breaks Android.
So I am bit in tricky situation.

Is there any other way to import 4.12?
in that case I can push the test.

regards,
Karim,

aslakhellesoy:
@nablik I just upgraded to JUnit 4.12. Could you add it now please?

nablik:
Done Sir :)

dkowis:
It's got some conflicts now. @nablik could you check to see if this is still legit?

nablik:
Sorry guys, I lost momentum, @David, which kind of conflict it has ?
in case of upgrade to JUnit 12 I think we can get rid of
org.junit.internal.AssumptionViolatedException
that was historically causing the problem.
Regards,
Karim.

brasmusson:
Duplicate of #961 (which is already merged).

PR #896, opened 2015-08-12
Tear down test in case TestNG groups a set

sventorben:
AbstractTestNGCucumberTests did not call tearDownClass in case TestNG groups are present.
Fixed by setting alwaysRun property on AfterClass annotation to true. This conforms to BeforeClass behaviour.


sventorben:
see cucumber/cucumber-jvm#956

PR #899, opened 2015-08-23
Made @Transform works with DataTable. Ref #898

### LINKED ISSUE #898: Be able to pre transform or overlay the String value of DataTable before it is converted
Originally, I'm looking to support the needs of being able to resolve a set of values like null. For example,

``` gherkin
Given I have a list of Person, configured as follows:
  | firstName | lastName |
  | Sid       | L        |
  | Kate      | [null]   |
```

This can be translated to `List<Person>` nicely, if Person class has firstName and lastName field. However, preferred result of Kate's last name is actually expected to be `null`, not literally [null].

For a single value, this could be done with `@Transform` to convert null string ([null], null, Null, etc.) into `null`. But up to this point, `@Transform` doesn't work with `DataTable`. I'm guessing the reason `@Transform` is not applicable to DataTable because `asList`, and `asMap` already does the job through `TableConverter`. 

However, I can't seems to find a nice solution to actually solve the above scenario. So I'm thinking that it might be good to make `@Transform` applicable to DataTable, but only to do some pre-conversion, more like `List<String>` --> `List<Person>`, so that users can still benefit from the TableConverter without doing extra works, but still be able to redefine some meaning in the gherkin table. 
##### Proposed Solution

According to the current codebase, I understand that there's a code in DataTable that converts `List<DataTableRow>` into `List<List<String>>`, upon its construction. 

Would it be a good idea to apply `Transformer` class upon this process to resolve the mentioned scenario? I can submit a PR that could help illustrate this.
##### Alternative Ideas
1. Using `@Transform` may be confusing in what it actually does. Maybe something like `@Transform(pre = Transformer.class)`
2. Or may be come up with `@PreTransform`.

manat:
This is to add a capability to pre-convert String value into desired object, before applying the usual table converter. Should this PR gets accepted, it would closes #898

### DIFF
--- a/core/src/main/java/cucumber/api/DataTable.java
+++ b/core/src/main/java/cucumber/api/DataTable.java
@@ -58,7 +58,7 @@ public class DataTable {
         List<List<String>> raw = new ArrayList<List<String>>();
         for (Row row : gherkinRows) {
             List<String> list = new ArrayList<String>();
-            list.addAll(row.getCells());
+            list.addAll(tableConverter.preConvert(row.getCells()));
             if (columns != row.getCells().size()) {
                 throw new CucumberException(String.format("Table is unbalanced: expected %s column(s) but found %s.", columns, row.getCells().size()));
             }
--- a/core/src/main/java/cucumber/runtime/ParameterInfo.java
+++ b/core/src/main/java/cucumber/runtime/ParameterInfo.java
@@ -129,6 +129,13 @@ public class ParameterInfo {
     public String toString() {
         return type.toString();
     }
+    
+    public boolean canConvert(Class<?> type) {
+        if (transformer == null) {
+            return false;
+        }
+        return transformer.canConvert(type);
+    }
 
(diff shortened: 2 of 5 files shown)


mpkorstanje:
We're currently working towards cucumber-expressions. Cucumber expressions will contain a different mechanism for transformations. As such I am closing this pull request. If the functionality were added now it would exist only for a very short time period.

s2oBCN:
I'm sorry, I can not see the point of how this issue can be solved with " cucumber-expressions". We have the same issue, that is to transform a Table directly. 

For example, in our case, it is a list of strings:

>  | Things|
  | Sid       |
  | true     |
  | red      |
  | blue    |
  | false    |

And we want to transform the true/false in 1/0.

It is true that we can use an Object and the use the XStreamConverter, but in this case, I think it's an overhead of unnecessary complexity.

Is there any other better way to solve this problem?

Thank you.

PR #909, opened 2015-09-05
Update to the newer, faster infrastructure

dkowis:
I'm pretty sure this will work just fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,9 @@ matrix:
   - jdk: oraclejdk8
     script: mvn -q deploy -P java8 -pl java8 --settings .travis-settings.xml -Dno.gem.deploy=true
 
+# get us on the new container based builds, we're not using any fancyness
+sudo: false
+
 branches:
   only:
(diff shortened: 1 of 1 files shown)


dkowis:
Wow! This is way faster. 8 minutes instead of 30 minutes. MERGINATING.

brasmusson:
Comparing the build time [before](https://travis-ci.org/cucumber/cucumber-jvm/builds/78228274) and [after](https://travis-ci.org/cucumber/cucumber-jvm/builds/78921368) shows that on jdk7 the build is indeed faster (~14 minutes versus ~8 minutes) with this change, but interestingly enough on jdk8 the build is slower (~4½ minutes versus 6½ minutes).

dkowis:
Huh, I saw times that were like 30 minutes, not 14 minutes.... Probably lots of variance in there :(

dkowis:
Eh, nope, I was looking at total time for all three runs, not the individual runs. Interesting.

PR #911, opened 2015-09-14
Handle doc string and data tables in a single step

davidegrohmann:


srbaker:
@aslakhellesoy can we get a merge on this? I want to make use of it. Thx.

haroon-sheikh:
@aslakhellesoy It would be really useful to get this merged in.

brasmusson:
As [the new Gherkin parser (currently at version 4)](https://github.com/cucumber/gherkin) does not support both a doc string and a data table in a single step, I truly believe that this should not be merged until it does. Otherwise it would block the upgrade from the outdated Gherkin v2.12.2 (#1035). As comparison Cucumber-Ruby upgraded from Gherkin v2.12.2 over a year ago.

aslakhellesoy:
I'm skeptical about allowing both doc strings and data tables in the same step. I don't think it encourages good documentation.

I might reconsider if someone can convince me with a real example.

srbaker:
@aslakhellesoy *Our* use case is that the product *is* a database. It's admittedly a narrow group, not everyone is building a database, but we are and it makes sense for us. And the alternative is confusing and shitty. It'll really help us to get this in; please feel free to add "unless you're building a database, you probably don't want this." to documentation.

Thanks!

aslakhellesoy:
I added a comment over in #867. We'd need a Gherkin PR, make a Gherkin 5 release and update all the Cucumbers. I'm open for that if you submit the Gherkin PR.

mpkorstanje:
I'm closing this issue due to inactivity.

Issue #912, opened 2015-09-23
cucumber-java8 fails on OpenJDK 1.8.0


mawis:
Cucumber-Java8 depends on internals of the implementation of Oracle's JDK8. When I started experimenting with it I just always got the following exception:

    Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.056 sec <<< FAILURE! - in soy.wimmer.CucumberIT
    Feature: Cucumber with Java8  Time elapsed: 0.047 sec  <<< ERROR!            
    cucumber.runtime.CucumberException: Failed to instantiate class soy.wimmer.CucumberStepdefs
    […]                                                                          
    Caused by: java.lang.reflect.InvocationTargetException: null                 
    […]                                                                          
    Caused by: cucumber.runtime.CucumberException: java.lang.IllegalArgumentException: Wrong type at constant pool index
    […]                                                                          
    Caused by: java.lang.IllegalArgumentException: Wrong type at constant pool index
        at sun.reflect.ConstantPool.getMemberRefInfoAt0(Native Method)           
        at sun.reflect.ConstantPool.getMemberRefInfoAt(ConstantPool.java:47)     
        at cucumber.runtime.java8.ConstantPoolTypeIntrospector.getTypeString(ConstantPoolTypeIntrospector.java:37)
        at cucumber.runtime.java8.ConstantPoolTypeIntrospector.getGenericTypes(ConstantPoolTypeIntrospector.java:27)
        at cucumber.runtime.java.Java8StepDefinition.<init>(Java8StepDefinition.java:45)
        at cucumber.runtime.java.JavaBackend.addStepDefinition(JavaBackend.java:162)
        at cucumber.api.java8.En.Given(En.java:190)                              
        at soy.wimmer.CucumberStepdefs.<init>(CucumberStepdefs.java:8)           
    […]

The test project causing this output is available at https://github.com/mawis/cucumber-java8-test

While it fails on OpenJDK8 it works without problems on Oracle JDK8. I consider this to be a bug, as I don't think that cucumber should depend on a specific implementation of the compiler.

aslakhellesoy:
Thanks for reporting this.

Do you have a suggestion for a portable implementation?

mawis:
Sorry, no. That's to much in the internals of Java for me.

bofkentucky:
[...]

PR #914, opened 2015-09-30
Fix IllegalArgumentException on JDK 1.8.0_60

dbcowboy:
To determine actual types of parameters for class passed in as lambda to
step definition, code uses old Sun ConstantPool to look for dynamic
linking data.  Previously these values were found at index size()-2;
however as of 1.8.0_60 the location appears to have changed to size()-3.
Given that the following indexes appear to have empty/invalid values,
instead iterate downward from end to find first valid value and use for
parameter type determination.


aslakhellesoy:
Thanks! Can you follow the [code standard](https://github.com/cucumber/cucumber-jvm/blob/master/CONTRIBUTING.md#contributinghacking) please?

dbcowboy:
Thanks for pointing me to the code standards; sorry I didn't have that in place before I checked in.

Let me know if I should re-submit this change with correct formatting.

rnaufal:
I've just downloaded the latest SNAPSHOT build (1.2.5-SNAPSHOT) and the this change isn't included on it yet. When are you planning to release a new snapshot build?

aslakhellesoy:
@rnaufal it will be in a snaphot jar shortly after this PR is merged.

rnaufal:
Thanks @aslakhellesoy. Does it will get merged soon?

aslakhellesoy:
@rnaufal yeah [RSN](https://en.wiktionary.org/wiki/RSN)

rnaufal:
Hi @aslakhellesoy, did you merge this PR? Thanks in advance!

aslakhellesoy:
@rnaufal the PR is open, so no

java8/src/main/java/cucumber/runtime/java8/ConstantPoolTypeIntrospector.java
```diff
@@ -34,7 +34,20 @@
...
-        String[] memberRef = constantPool.getMemberRefInfoAt(constantPool.getSize() - 2);
+        int size = constantPool.getSize();
+        String[] memberRef = null;
+
+        // find last element in constantPool with valid memberRef
+        // - previously always at size-2 index but changed with JDK 1.8.0_60
+        for (int i = size - 1; i > -1; i--) {
+            try {
+                memberRef = constantPool.getMemberRefInfoAt(i);
+                break;
```
clicman: Return should be here.

java8/src/main/java/cucumber/runtime/java8/ConstantPoolTypeIntrospector.java
```diff
@@ -34,7 +34,20 @@
...
+            } catch (IllegalArgumentException e) {
+                // eat error; null entry at ConstantPool index?
+            }
+        }
+
         return memberRef[2];
```
clicman: Here must be exception in case something goes wrong with for loop.

erran-r7:
A colleague @kprzerwa-r7 and I encountered this without being sure whether it was ever released. For the next person who stumbles across this issue it was fixed in 96d1e85df4289e290bb4c61893be5516d4dcc9ee (which was released in `1.2.5`).

> [Java8] Fix IllegalArgumentException on JDK 1.8.0_60 (#912, #914 Michael Wilkerson)

Issue #918, opened 2015-10-13
java8: Use Consumer instead of HookBody and StepdefBody


mhfrantz:
In order to compose with API's such as Durian's `Errors`, the proprietary `HookBody` and `StepdefBody` should be replaced with `java.util.function.Consumer`.  This would allow use of Durian to work around step bodies that contain checked exceptions.
```
Given("^a step with a string argument \"(.*)\"$",
        Errors.rethrow().wrap(
                (final String arg) -> { ... throw new Exception("this might happen"); ... }));
```

aslakhellesoy:
Sounds like a good idea. Send a PR?

mhfrantz:
Attempt to fix failed.  Reopen if there are better ideas.

PR #919, opened 2015-10-14
Java8 standard function interfaces

mhfrantz:
This PR falls short of the goal of #918 because of the difficulty of preserving the introspection of lambda step definitions.  I believe the goal of supporting functional programming API's like Durian Errors is probably jeopardized by this introspection technique, so the changes would be moot.

This PR does improve Before/After by allowing standard Java8 function types instead of proprietary interfaces.


mhfrantz:
Withdrawing this PR, as it appears that it requires Java8.

Issue #916, opened 2015-10-13
Cannot close over local variables


atorstling:
`Java8StepDefinition` and more specifically `ConstantPoolTypeIntrospector` confuses closed variables with arguments to the accept function. This has the consequence that you are unable to close over local variables.

For example, if you  define something like

    class Steps implements En {
      Steps() {
          String prefix = "hello ";
          When("^blabla$", () -> prefix + "world" ); 
      }
    }

, the lambda closes over the `prefix` variable. Subsequently, the `ConstantPoolTypeIntrospector` will read out a `typeString` not of the expected `()V`, but of `(Ljava/lang/String;)V`.  `Java8StepDefinition` will therefore get `typeArguments` of `Type[]{ String.class }`. It will then proceed looking for a method called accept with a single `Object` argument, which will fail (it treats all arguments as Object arguments, disregarding the specific type). In other words, it "thinks" that the lamda looks as follows:

    class Steps implements En {
      Steps() {
          String prefix = "hello ";
          When("^blabla$", (prefix) -> prefix + "world" ); 
      }
    }

, even though `prefix` is implicitly passed. I believe this could be solved by simply calling `clazz.getDeclaredMethod("accept")`, instead of looking in the constant pool, which evidently exposes implicit dependencies as well as explicit ones.

This problem is not present if you close over fields in the `Steps` class.

The exception you get when this happens is:
```
cucumber.runtime.CucumberException: java.lang.NoSuchMethodException:  com.example.Steps$$Lambda$3/358849801.accept(java.lang.Object)
    at cucumber.runtime.java.JavaBackend.addStepDefinition(JavaBackend.java:164)
    at cucumber.api.java8.En.When(En.java:370)
    at com.example.Steps.<init>(Steps.java:54)
```

I'm using cucumber 1.2.4

aslakhellesoy:
Interesting. Would you be able to send a pull request with your proposed fix?

atorstling:
Sure thing, please see above.

PR #924, opened 2015-10-19
Cannot close over local variable

atorstling:
This is a reproducing test and suggested fix for issue #916. The problem turned out to be that the information read out of the ConstantPool didn't actually represent the type signature of the lambda class when we had closed over local variables.

I also saw issue #912 which is related to that the constant pool interface isn't uniform or stable, so I decided to try to avoid it altogether.

So instead of trying to deduce the type signature of the body class, I look at the concrete types of the "accept" method parameters. I'm not sure if this is equivalent to the old method and if there are some cases which it doesn't cover, but at least all tests are passing in the project.

This fix also leads to the TypeIntrospector being unused, but I didn't remove it in the pull request to avoid cluttering.

### DIFF
--- a/java8/src/test/java/cucumber/runtime/java8/test/LambdaStepdefs.java
+++ b/java8/src/test/java/cucumber/runtime/java8/test/LambdaStepdefs.java
@@ -1,8 +1,10 @@
 package cucumber.runtime.java8.test;
 
 import cucumber.api.DataTable;
+import cucumber.api.PendingException;
 import cucumber.api.Scenario;
 import cucumber.api.java8.En;
+import cucumber.api.java8.StepdefBody;
 
 import java.util.List;
 
@@ -17,10 +19,13 @@ public class LambdaStepdefs implements En {
             assertNotSame(this, lastInstance);
             lastInstance = this;
         });
-
         Given("^this data table:$", (DataTable peopleTable) -> {
             List<Person> people = peopleTable.asList(Person.class);
             assertEquals("Hellesøy", people.get(0).last);
+        });
+        Integer alreadyHadThisManyCukes = 1;
+        Given("^I have 42 cukes in my belly$", () -> {
+            assertEquals((Integer) 1, alreadyHadThisManyCukes);
         });
     }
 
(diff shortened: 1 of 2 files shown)


atorstling:
This is also related to #914 - that fix tries to solve a related problem, if not the same.

java8/src/test/java/cucumber/runtime/java8/test/LambdaStepdefs.java
```diff
@@ -17,11 +19,14 @@ public LambdaStepdefs() {
             assertNotSame(this, lastInstance);
             lastInstance = this;
         });
-
         Given("^this data table:$", (DataTable peopleTable) -> {
             List<Person> people = peopleTable.asList(Person.class);
             assertEquals("Hellesøy", people.get(0).last);
         });
+        Integer alreadyHadThisManyCukes = 1;
+        Given("^I have 42 cukes in my belly$", () -> {
```
aslakhellesoy: Looks like your change breaks arguments. Try changing this line to the following to reproduce:

```java
Given("^I have (\\d+) cukes in my belly$", (Integer n) -> {
```

Any thoughts about how to deal with that?

atorstling: Hmm, I didn't expect it to break that easily. What happens? I can have a look in a couple of days, but not right now unfortunately.

aslakhellesoy: The lambda parameters are reported as `java.lang.Object`, so the type conversion breaks.

atorstling: Oh well, so much for that technique then ;) If its not caused by the interface methods being declared accepting Object, I have no immediate ideas for workarounds. Maybe merge the other fix and I'll see if there is a way around it at nearest occation. Should have tested it better :)

atorstling: It's funny though - the documentation for getGenericParameterTypes says "If a formal parameter type is a parameterized type, the Type object returned for it must accurately reflect the actual type parameters used in the source code.".

atorstling:
I found that https://github.com/jhalterman/typetools/blob/master/src/main/java/net/jodah/typetools/TypeResolver.java also uses the constant pool to resolve lambda arguments, so maybe your old method is the only viable. http://stackoverflow.com/questions/21887358/reflection-type-inference-on-java-8-lambdas certainly seems to implicate that. Perhaps using TypeResolver is a good idea, so that support for new JDKs can be added there?

atorstling:
See #929 for a new attempt

PR #940, opened 2015-12-08
Prevent MethodScanner from checking Object.class methods

lsuski:


mpkorstanje:
A belated thank you for your contribution!

PR #944, opened 2015-12-28
Two branches in the same conditional structure should not have exactly the same implementation

kirill-vlasov:
This pull request is focused on resolving occurrences of Sonar rule squid:S1871 - Two branches in the same conditional structure should not have exactly the same implementation
You can find more information about the issue here:
https://dev.eclipse.org/sonar/coding_rules#q=squid:S1871
Please let me know if you have any questions.
Kirill Vlasov


mpkorstanje:
Given that both the TestNgReporter and the Android runners Arguments are mappings I don't see a major problem with the duplication there. I think it is more important to keep the over all structure consistent. The Step Definitions Match has since been re-factored.

Issue #953, opened 2016-01-22
TestNG does not fail if feature files are not parseable


sventorben:
If a feature file has an invalid syntax and is not parseable, the dataprovider in AbstractTestNGCucumberTests does not fail. The dataprovider does not provide any feature to the test method. Accordingly, no tests are executed and the no tests fails.

I would expect the test to fail!

aslakhellesoy:
Would you be interested in sending a pull request with a fix and a test?

PR #954, opened 2016-01-22
Allows to configure TestNG to fail on unparseable feature files

sventorben:
TestNG DataProvider did not fail if a feature file was not parseable.
This commit allows to configure whether a test fails if at least one
feature file is not parseable.
In addition, test always fails if at least one feature file is not
parseable and strict mode is enabled.

Fixes cucumber/cucumber-jvm#953

### DIFF
--- a/testng/src/test/java/cucumber/api/testng/AbstractTestNGCucumberTestsTest.java
+++ b/testng/src/test/java/cucumber/api/testng/AbstractTestNGCucumberTestsTest.java
@@ -1,0 +1,60 @@
+package cucumber.api.testng;
+
+import org.testng.Assert;
+import org.testng.TestListenerAdapter;
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+
+import cucumber.api.CucumberOptions;
+
+public final class AbstractTestNGCucumberTestsTest {
+
+    @Test
+    public void failOnParsingError() throws Exception {
+        TestListenerAdapter tla = runTestNG(FailOnUnparseableFeaturesTest.class);
+        Assert.assertEquals(tla.getFailedTests().size(), 1, "Dataprovider must fail.");
+        Assert.assertEquals(tla.getPassedTests().size(), 0, "No test has passed.");
+        Assert.assertEquals(tla.getSkippedTests().size(), 1, "Actual test has been skipped.");
+    }
+
+    @Test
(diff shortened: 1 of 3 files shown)


testng/src/test/java/cucumber/api/testng/AbstractTestNGCucumberTestsTest.java
```diff
@@ -0,0 +1,60 @@
...
+        TestListenerAdapter tla = runTestNG(FailOnUnparseableFeaturesIfStrictModeTest.class);
+        Assert.assertEquals(tla.getFailedTests().size(), 1, "Dataprovider must fail.");
+        Assert.assertEquals(tla.getPassedTests().size(), 0, "No test has passed.");
+        Assert.assertEquals(tla.getSkippedTests().size(), 1, "Actual test has been skipped.");
+    }
+
+    @Test
+    public void succeedThoughParsingError() throws Exception {
+        TestListenerAdapter tla = runTestNG(SucceedOnUnparseableFeaturesTest.class);
+        Assert.assertEquals(tla.getFailedTests().size(), 0, "Dataprovider must not fail.");
```
aslakhellesoy: Expected result should be the 1st argument - actual the 2nd.

sventorben: True for JUnit, but this uses TestNG where parameters are the other way round.

brasmusson:
The root cause here is that the `CucumberException` caused by the parse error in the `@DataProvider`method seems to be silently swallowed by TestNG. The normal behaviour (Command-Line-Runner, JUnit-Runner, Cucumber-Ruby) is that the `CucumberException` is propagated to the user (and no other feature is executed). In case of the JUnit-Runner the parse error seems to be reported as a test case error.

I'm not sure that a separate TestNG test case checking that the features are parse-able is the best solution, a solution more consistent with the JUnit-Runner would be to catch the `CucumberException` from the parse error and propagate it so it can be re-thrown inside a TestNG test case (as I spiked here): https://github.com/cucumber/cucumber-jvm/compare/testng-parse-error-handling

sventorben:
@brasmusson: I didn't take a look at the other runners. However, your approach looks like a much cleaner design.

brasmusson:
Closed in favour of 27b02a0

Issue #955, opened 2016-01-22
AbstractTestNGCucumberTests does not tear down test in case TestNG groups a set


sventorben:
To execute cucumber tests with AbstractTestNGCucumberTests a group named "cucumber" must be set.
When setting a group, TestNG only executes methods annotated with that group or having thhe alwaysRun-attribute set to true. 

Within AbstractTestNGCucumberTests there are two configuration methods, namely setUpClass and tearDownClass. Only setUpClass is annotated with alwaysRun=true, but tearDownClass is not.
This means that tearDownClass will never be executed when cucumber tests are executed, because the group "cucumber" is set. 

The behaviour of the two configuration methods is inconsistent and should be unified.

PR #956, opened 2016-01-22
Always tear down TestNG cucumber tests.

sventorben:
This ensures that TestNGCucumberRunner.finish() will always be invoked.

Fixes cucumber/cucumber-jvm#955


brasmusson:
Thanks!

PR #961, opened 2016-02-14
Support assume feature with JUnit 4.12.

stefanbirkner:
JUnit 4.12 uses a different class for the `AssumptionViolatedException` than older versions.


brasmusson:
Thanks!

Issue #962, opened 2016-02-14
JunitFormatter fails to report an scenario as skipped if a PendingException is thrown in a hook.


fmarmar:
In cucumber-java 1.2.4 if you throw a PendingException in a scenario hook, all scenario steps are skipped as expected, but JunitFormatter doesn't detect it and it shows the scenario ok.

PR #964, opened 2016-02-14
Mark scenario as skipped in junit reporter if PendingException is thrown in a hook

fmarmar:
Fixes [cucumber/cucumber-jvm#962](https://github.com/cucumber/cucumber-jvm/issues/962)


brasmusson:
To be consistent, with respect to JUnit listeners (for instance Maven Surefire when using Maven to run tests), PendingExceptions should trigger the "Test Ignored" notification in order to be reported as skipped by the listener, see https://github.com/cucumber/cucumber-jvm/commit/8e701df963e547b50c111a5f5c85aa18dd16bccd.

PR #969, opened 2016-03-01
remove reference to decommissioned repository.codehaus.org repository

cheister:
"All Codehaus services will be terminated progressively until May 31st 2015" and repository.codehaus.org is no longer a usable maven repository.


aslakhellesoy:
Ah Codehaus - it used to be a great place!

PR #970, opened 2016-03-08
adding CucumberRule to JUnit module

rmannibucau:
It allows to use RuleChain to build chain of rules and do things before/after cucumber itself reusing existing JUnit integration.

### DIFF
--- a/junit/src/main/java/cucumber/api/junit/CucumberRule.java
+++ b/junit/src/main/java/cucumber/api/junit/CucumberRule.java
@@ -1,13 +1,14 @@
 package cucumber.api.junit;
 
+import cucumber.runner.EventBus;
 import cucumber.runtime.ClassFinder;
 import cucumber.runtime.Runtime;
 import cucumber.runtime.RuntimeOptions;
-import cucumber.runtime.RuntimeOptionsFactory;
 import cucumber.runtime.io.MultiLoader;
 import cucumber.runtime.io.ResourceLoader;
 import cucumber.runtime.io.ResourceLoaderClassFinder;
 import cucumber.runtime.junit.FeatureRunner;
+import cucumber.runtime.junit.JUnitOptions;
 import cucumber.runtime.junit.JUnitReporter;
 import cucumber.runtime.model.CucumberFeature;
 import org.junit.rules.TestRule;
@@ -20,39 +21,16 @@ import org.junit.runners.model.Statement;
 
 import java.lang.reflect.Method;
 import java.util.ArrayList;
-import java.util.Collections;
+import java.util.Collection;
 import java.util.List;
 
 import static java.util.Arrays.asList;
 
 public class CucumberRule implements TestRule {
-    private final RuntimeOptions runtimeOptions;
-
-    public CucumberRule() {
-        this(new RuntimeOptionsFactory(CucumberRule.class /* no conf */).create());
-    }
-
-    public CucumberRule(final RuntimeOptions runtimeOptions) {
-        this.runtimeOptions = runtimeOptions;
-    }
-
-    public CucumberRule(final String defaultPackage, final String... args) {
-        final List<String> options = new ArrayList<String>(args == null ? Collections.<String>emptyList() : asList(args));
-        if (defaultPackage != null) {
-            if (options.isEmpty()) {
-                options.add(MultiLoader.CLASSPATH_SCHEME + defaultPackage.replace('.', '/'));
-            }
-            if (!hasOption(options, "--glue")) {
-                options.add("--glue");
-                options.add(defaultPackage);
-            }
-            if (!hasOption(options, "--plugin")) {
-                options.add("--plugin");
-                options.add("pretty");
-            }
-        }
-        this.runtimeOptions = new RuntimeOptions(options);
-    }
+    private final List<String> runtimeOptions = new ArrayList<String>();
+    private final List<String> junitOptions = new ArrayList<String>();
+    private boolean featureAdded = false;
+    private final Collection<String> potentialFeatures = new ArrayList<String>();
 
     private boolean hasOption(List<String> options, String optName) {
         boolean hasPlugin = false;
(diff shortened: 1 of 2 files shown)


haroon-sheikh:
This looks good to me.

mariaklimenko:
Sounds useful

haroon-sheikh:
@aslakhellesoy can this be merged in please?

junit/src/main/java/cucumber/api/junit/CucumberRule.java
```diff
@@ -0,0 +1,139 @@
...
+                } finally {
+                    jUnitReporter.done();
+                    jUnitReporter.close();
+                    runtime.printSummary();
+                }
+            }
+        };
+    }
+
+    public static class ForceRunner extends BlockJUnit4ClassRunner {
```
mpkorstanje: What is the purpose of this?

junit/src/main/java/cucumber/api/junit/CucumberRule.java
```diff
@@ -0,0 +1,139 @@
...
+                try {
+                    for (final CucumberFeature feature : cucumberFeatures) {
+                        new FeatureRunner(feature, runtime, jUnitReporter).run(new RunNotifier());
+                    }
+                    base.evaluate();
+                } finally {
+                    jUnitReporter.done();
+                    jUnitReporter.close();
+                    runtime.printSummary();
+                }
```
mpkorstanje: This won't work any more. We're using an event based protocol now.

junit/src/main/java/cucumber/api/junit/CucumberRule.java
```diff
@@ -0,0 +1,139 @@
...
+                options.add(MultiLoader.CLASSPATH_SCHEME + defaultPackage.replace('.', '/'));
+            }
+            if (!hasOption(options, "--glue")) {
+                options.add("--glue");
+                options.add(defaultPackage);
+            }
+            if (!hasOption(options, "--plugin")) {
+                options.add("--plugin");
+                options.add("pretty");
+            }
```
mpkorstanje: I reckon a builder pattern would be more suited to pass options.

junit/src/main/java/cucumber/api/junit/CucumberRule.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.Collections;
+import java.util.List;
+
+import static java.util.Arrays.asList;
+
+public class CucumberRule implements TestRule {
+    private final RuntimeOptions runtimeOptions;
+
+    public CucumberRule() {
+        this(new RuntimeOptionsFactory(CucumberRule.class /* no conf */).create());
```
mpkorstanje: This only makes sense for classes that are annotated with @CucumberOptions

junit/src/main/java/cucumber/api/junit/CucumberRule.java
rmannibucau: Agree but less cucumber like. Maybe a core refactoring to build cli on top of it?

junit/src/main/java/cucumber/api/junit/CucumberRule.java
rmannibucau: Make junit happy without loosing stand runner features. Can be replaced with a more advanced/complete custom runner but harder to maintain probably.

junit/src/main/java/cucumber/api/junit/CucumberRule.java
rmannibucau: Not sure what you mean. Goal of the rule was to comibe cucumber with other rules (container, ftp/sft/... server, mysql, nosql databaseq etc...)

mpkorstanje:
>Make junit happy without loosing stand runner features. Can be replaced with a more advanced/complete custom runner but harder to maintain probably.

I'm not happy with the ForceRunner or NoTestRunner. It is essentially a new runner. And having looked at JUnits rules more closely I now understand that this is an alteration _in how a test method, or set of test methods, is run and reported_. So a Cucumber Rule that executes the test goes against the design framework. This makes both Cucumber and Junit harder to use.

So I believe the only correct solution is to implement rules an annotation in the Cucumber Runner. However the current implementation already extends the `ParentRunner` which should be able handle `@ClassRules`. So it is already possible to use rules to do things around the execution of Cucumber.

So I'm not sure what use case this solution actually has.

rmannibucau:
This reasonning is partially valid - it ignores rules ordering which is important - but add to it you can use another runner (arquillian, applicationcomposer, spring, apache directory, ...) and rules still make sense.

Also note this small workaround avoids a new api which is not bad.

We can probably mix cucumber rule ans runner to reuse existing one if needed bit making explicit it is a junit passthrough impl is not that blocking imo.

mpkorstanje:
Unless I'm mistaken the `RuleChain` can be used in `@ClassRule` so there should be sufficient ability to order the test rules. However they have to be ordered **around** `Cucumber`. 

It would be nice if multiple runners could contribute but this is not something that JUnit4 supports. JUnit5 Jupiter does support these things - though it was still in its conceptional phase when this PR was created this is no longer the case.

As such I see no good reason to work around JUnit4s limitations or support such a workaround in the long term as part of cucumber-jvm. Finishing Junit5 support and upgrading to JUnit5 would be the most sensible and sustainable longterm approach.

rmannibucau:
@mpkorstanje multiple runners = rule chain actually. JUnit 5 supports it but it is also not the next version of JUnit 4 since it breaks all the API and support of JUnit 4 (it is a bit like saying "TestNG supports it"). So still think it makes sense - mainly cause there is a huge ecosystem here already - to support rules.

mpkorstanje:
> So still think it makes sense - mainly cause there is a huge ecosystem here already - to support rules.

Cucumber currently supports rules. And this is not what I understand your use case to be. I understand your use case to be integration with multiple runners. This is not something supported by Junit4. So I do not wish to support a workaround in the long term as part of cucumber-jvm.

You are of course free to create your own module that wraps the `Cucumber` runner in rules or any other sort of workaround.

rmannibucau:
@mpkorstanje hmm, JUnit 4 supports chain of runners...this is why rules were created actually. Concretely it is not "runners" but "rules" but it leads to the same outcome, just more powerful. I pushed an example at https://github.com/rmannibucau/rule-them-all#user-content-combine-them. Cucumber would be welcomed being the last rule in the sample and we could append a cukedoctor + asciidoctor rendering rule after cucumber itself so I really think it makes a lot of sense to have cucumber as a rule. What am I missing?

PR #971, opened 2016-03-08
[OpenWebBeans] Add ObjectFactory for Apache OpenWebBeans 

rmannibucau:
An alternative to Weld supporting already running container to smoothly integrate with OpenEJB if needed - and with the rule PR I did earlier.

Side note: the PR uses cucumber.api.openwebbeans.OpenWebBeansConfig as a SPI cause cucumber doesn't provide a "config scanning" out of the box but it would have been nice to have it as an annotation.

### DIFF
--- a/openwebbeanssrc/test/java/cucumber/runtime/java/openwebbeans/OpenWebBeansObjectFactoryTest.java
+++ b/openwebbeanssrc/test/java/cucumber/runtime/java/openwebbeans/OpenWebBeansObjectFactoryTest.java
@@ -1,0 +1,29 @@
+package cucumber.runtime.java.openwebbeans;
+
+import cucumber.api.java.ObjectFactory;
+import org.junit.Test;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNotSame;
+
+public class OpenWebBeansObjectFactoryTest {
+    @Test
+    public void shouldGiveUsNewInstancesForEachScenario() {
+        ObjectFactory factory = new OpenWebBeansObjectFactory();
+        factory.addClass(BellyStepdefs.class);
+
+        // Scenario 1
+        factory.start();
+        BellyStepdefs o1 = factory.getInstance(BellyStepdefs.class);
+        factory.stop();
+
+        // Scenario 2
+        factory.start();
+        BellyStepdefs o2 = factory.getInstance(BellyStepdefs.class);
+        factory.stop();
+
+        assertNotNull(o1);
+        assertNotSame(o1, o2);
+    }
+
+}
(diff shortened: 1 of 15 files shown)


rmannibucau:
Still relevant

mpkorstanje:
@rmannibucau do we still need this with #1616?

rmannibucau:
@mpkorstanje yes and no. Since this PR CDI (v2) has a standard API for standalone apps (SeContainer) so this can be a portable replacement for openwebbeans and weld.

CdiCtrl of deltaspike was for before this time - which means that using deltaspike and CDI class explicitly is likely unintended.

Im fine closing the or if CDI Se API replaces it but i cant always use deltaspike in the testing stack - i want to control the boot api for cucumber.

mpkorstanje:
> CdiCtrl of deltaspike was for before this time - which means that using deltaspike and CDI class explicitly is likely unintended.

So you are saying that the implementation in #1616 is using CDI incorrectly? Could you put that comment in there. I don't know enough about the CDI ecosystem to make definitive statements.

> Im fine closing the or if CDI Se API replaces it but i cant always use deltaspike in the testing stack - i want to control the boot api for cucumber.

Figured.

rmannibucau:
Commented.

toepi:
Sry to ask it her. But what do you want control until boot? Is CDIExtensions (https://docs.oracle.com/javaee/6/api/javax/enterprise/inject/spi/Extension.html) mechanism not enough?

rmannibucau:
No for these cases:

1. Add steps as cdi beans without making them explicitly beans (guess an extension could almost do it but not 100% guaranteed),
2. Reuse started container for all tests (cucumber or not),
3. Test some toggling logic in libs between cdi se api, deltaspike and other modes (requires classpath to not contain org.apache.deltaspike or other packages depending executions)

So nothing strictly wrong with DS but some advanced usages requiring it to not be there.

toepi:
for 1: Producer?

for 2: oh i thought i was the only one who need it. How do you solve it? I use scopes for it and think about it for the other pr - but unsure ....

mpkorstanje:
I'm happy to take both PR's. While CDI containers are used only by a fraction of the users they're relatively low maintenance and have a relatively high number of contributors.

I do however have a rather limited knowledge of either so I'd like both implementations to be correct, in line with the frameworks expected use (so as to limit future questions/issues) and focused on the basic use case. For advanced use cases I expect that people re-implement the factory. At least until we come up with something better.

rmannibucau:
@toepi

1. Extension or just not making them cdi beans but lazily injection target (i use last option generally since it covers more cases but hapoy with first one too)
2. Scope doesnt work since each runner will start/stop the container. In openwebbeans we use a static global instance and if set we dont re-set it. A shutdown hook closes it.

rmannibucau:
closing as being superseeded by https://github.com/cucumber/cucumber-jvm/pull/1626

PR #980, opened 2016-03-31
Groovy: Fixed NPE when using List as parameter type in step definitions

sjacobs:
To reproduce:

The following code caused a NullPointerException because ParameterInfo was initialized without default delimiter:
```
Given(~'^this should be converted to a list:(.+)$') { List list ->
    assertEquals(3, list.size())
    assertEquals("Cucumber-JVM", list.get(0))
    assertEquals("Cucumber", list.get(1))
}
```

The pull request contains the fix and the corresponding test.


haroon-sheikh:
Looks good to me.

Issue #906, opened 2015-09-02
The test SubstitutionStepdefs#I_should_receive_an_email_with_the_body fails on Windows platform


lehmann:
The test is failing when runs on Windows. Apparently is a problem with line termination character.

```
org.junit.ComparisonFailure: expected:<Dear Jojo Binks,[
You have been granted Admin rights.  You are able to manage any user account on the system. Please be responsible.]
-The Admins> but was:<Dear Jojo Binks,[
You have been granted Admin rights.  You are able to manage any user account on the system. Please be responsible.
]
-The Admins>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at cucumber.runtime.java.test.SubstitutionStepdefs.I_should_receive_an_email_with_the_body(SubstitutionStepdefs.java:38)
	at ✽.Then I should receive an email with the body:(cucumber/runtime/java/test/scenario_outline_substitution.feature:5)
```

sebrose:
Fixed

PR #981, opened 2016-04-01
Replace \n with %n so that tests work on Windows.

corinned:
#906


aslakhellesoy:
That's clever - and I learned something new (didn't know about `%n`)

PR #983, opened 2016-04-04
Allow construction of GuiceFactory from Injector

reftel:
While convenient, `InjectorSourceFactory` requires the `InjectorSource` to be default constructible. In my case, it can't really, since I'd like to bind an instance as an eager singleton, and the `InjectorSource` cannot know beforehand which instance it will be. Opening up the `GuiceFactory`constructor that takes an Injector instance should solve this.


mpkorstanje:
While we provide Guice injection as a module, the class it self is not part of the public api and extending it is discouraged. I would recommend that you copy the module and make your own modifications.

Issue #902, opened 2015-08-28
Get Feature file and Feature name in Before hook


dkirrane:
I'd like to be able the obtain the feature file name and Feature name in the `@Before` hook for logging purposes.

Something like the existing `Scenario` object:

```
    @Before("@Before")
    public void beforeScenario(Feature feature, Scenario scenario) {  
    }
```

mikegolod:
For me it would be useful to load resources from a feature file location

hankscrpio:
@sebrose 
I second the usefulness of adding the ability to retrieve the current Feature being executed, as this would largely help for logging purposes. For example, I screen record all tests and capture failures in a separate directory. Ideally, this directory would be named after the current Feature utilizing the current Scenario's name. If you have any questions about my setup, please let me know. Thanks!

zhtaoang:
There is a "scenarioId" in the scenario object, which contains both feature name and scenario name. Try using "scenario.getId()" to get them.

tompahoward:
Sorry to resurrect this, but `scenario.getId()` does provides a sanitised version of the feature name, *not* the actual feature name. For instance, For a feature called "TestCase0.0 Basic" and a scenario "Basic", `scenario.getId()` provides "testcase0.0-basic;basic" *not* "TestCase0.0 Basic;Basic"

However `scenario.getName()` *does* correctly return "Basic"

This issue has been tagged "Won't Fix". If possible, it would be great to get a explanation so we can understand why getting the Feature name would be a bad thing to do.

brasmusson:
With the newer version of the Gherkin library (v3.x.x and v4.x.x) the basis of executing is [Pickles](https://github.com/cucumber/cucumber/tree/master/gherkin#pickles) which the feature files are compiled into. The Pickle is a data structure designed for execution - avoiding feature file complexities like Background and Scenario Outline when executing, and does not assume that is was created from a feature file in the first place - in the future a compiler from markdown to Pickles may be developed. 

Since a Pickle does not contain any information about feature name (by design), it is impossible to provide feature name to Before or After Hooks in the coming Cucumber-JVM 2.0.0 that uses v4 of the Gherkin library (available as a [snapshot release](https://cucumber.io/docs/reference/jvm) - note that the groupId has changed to "io.cucumber").

[...]

PR #984, opened 2016-04-05
Get Feature Information from Scenario

corinned:
Use @Before with a call to getFeature() on the Scenario object to get the instance of the Feature.  Use the Feature instance to get information about the Feature for logging.

public void get_feature_name(Scenario scenario) {
     Feature feature = scenario.getFeature();
     featureName = feature.getName();
     tags = feature.getSourceTagNames();
}
#902

### DIFF
--- a/core/src/main/java/cucumber/api/Feature.java
+++ b/core/src/main/java/cucumber/api/Feature.java
@@ -6,7 +6,7 @@ import java.util.Collection;
  * Before or After Hooks that declare a parameter of this type will receive an instance of this class.
  * It allows writing text and embedding media into reports, as well as inspecting results (in an After block).
  */
-public interface Scenario {
+public interface Feature {
     /**
      * @return source_tag_names. Needed for compatibility with Capybara.
      */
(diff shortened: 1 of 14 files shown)


Purus:
+1

kirillzh:
Would be a great feature to have!

b-lal:
Hi,
I was actually looking for this functionality, great idea Corinne.
When this is due to be release?

brasmusson:
When upgrading Cucumber-JVM to use Gherkin v4.0.0 with the new [Gherkin compiler](https://github.com/cucumber/gherkin#compiler) (which is where I focus my effort currently), the Feature information will not be available when running compiled scenarios. So @b-lal, my answer would be that this PR will not be merged (I'm sorry @corinned).

aldosuwandi:
+1

brasmusson:
Now that the master branch as been updated to use Gherkin v4, this pull-request has been invalidated - and it is not possible to provide the feature name to Before or After Hooks (see [this comment](https://github.com/cucumber/cucumber-jvm/issues/902#issuecomment-309293480)).

nsidhaye:
Hmm. Want to run cleanups after feature file. :(

PR #986, opened 2016-04-06
Use Integer.compare() in HookComparator in order to guard against possible underflow

### LINKED ISSUE #985: @Before and @After hook order comparison susceptible to underflow
If two @Before hooks are created with a great enough difference (i.e. greater than 2^31), they may be executed in the incorrect order due to an arithmetic underflow in HookComparator.

auno:
Fixes #985.


aslakhellesoy:
LGTM!

auno:
@aslakhellesoy, just a friendly reminder to merge this in case it accidentally dropped off your radar. Thanks.

aslakhellesoy:
Not forgotten, but we have a big backlog :-)

Issue #989, opened 2016-04-14
Lazy loading feature file and running (JUnit or TestNG Runner)


stanislav-antufeev:
I am facing the problem with running huge number of feature files, around 10,000. Some of them contains multiple scenarios (more than 5)and i am getting OutOfMemory error on ~1200 executed feature files. As I understand how JUnit and TestNG runner classes have been implemented, they are both loading all given feature file into memory and parsing using Gerkhin library. Current implementation of JUnit and TestNG feature runners has such limitation.
I found that TestNG framework has a special type of DataProvider which returns Iterator<Resource>.
I prepared some implementation of additional TestNG Runner class with lazy loading and running feature one by one.

stanislav-antufeev:
Could you please verify my commit. I can understand that it is not common case but an issue is exist and should be fixed or as I created a separate solution to run huge amount of feature files.

aslakhellesoy:
Please create a pull request. Our CI will then verify it automatically.

stanislav-antufeev:
I am already created and closed. I was confused by GIT Hub process of pull request a bit. @aslakhellesoy Are you seeing my pull request ?

aslakhellesoy:
Yes, I can see #991 thanks

stanislav-antufeev:
Closing issue because feature is implemented not completely.

PR #991, opened 2016-04-17
#989 Lazy loading feature file and running (JUnit or TestNG Runner)

stanislav-antufeev:
Please take a look at my source code changes.

### DIFF
--- a/core/src/main/java/cucumber/runtime/FeatureBuilder.java
+++ b/core/src/main/java/cucumber/runtime/FeatureBuilder.java
@@ -1,163 +1,26 @@
 package cucumber.runtime;
 
-import cucumber.runtime.io.Resource;
 import cucumber.runtime.model.CucumberFeature;
-import gherkin.I18n;
-import gherkin.formatter.FilterFormatter;
-import gherkin.formatter.Formatter;
-import gherkin.formatter.model.Background;
-import gherkin.formatter.model.Examples;
 import gherkin.formatter.model.Feature;
-import gherkin.formatter.model.Scenario;
-import gherkin.formatter.model.ScenarioOutline;
-import gherkin.formatter.model.Step;
-import gherkin.lexer.Encoding;
-import gherkin.parser.Parser;
-import gherkin.util.FixJava;
 
 import java.io.File;
-import java.io.IOException;
-import java.io.InputStreamReader;
-import java.math.BigInteger;
-import java.nio.charset.Charset;
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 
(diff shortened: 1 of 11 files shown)


aslakhellesoy:
Why did you close the PR?

stanislav-antufeev:
Sorry, I will reopen. I am newbie on GitHub.

stanislav-antufeev:
@aslakhellesoy  Are you seeing my pull request ?

stanislav-antufeev:
@aslakhellesoy 
Is it any estimate to review and merge if there is no any issues from you point of view ?

dkowis:
I'm not familiar with TestNG, but all the tests pass. Could I get one more sign off from someone else maybe with experience with TestNG?

stanislav-antufeev:
Guys, please verify my request. Generate JUnit runner classes for about 10000 feature files is not cool solution.

brasmusson:
As far as I can see this PR only implements lazy loading of feature files when using the TestNG module of Cucumber-JVM (that is extending `cucumber.api.testng.AbstractTestNGLazyLoadCucumberTests`), and lazy loading is not possible when using the JUnit module (that is `cucumber.api.junit.Cucumber`) or when using the CLI (that is `cucumber.api.cli.Main`). So I find the title of the PR misleading.

I would be happier if it was possible to lazy load regardless of the which of the above mentioned run interfaces were used. Lazy loading should not be the default behaviour, but rather enabled by an option, I think.

This PR introduces new interface functions on Core module for the TestNG module to use (`RuntimeOptions.getgetCucumberFeaturesAsResources`), I get the feeling that they could make it more difficult to implement a more general lazy loading solution later. All in all, I am a little reluctant to accept the RP as is.

> Generate JUnit runner classes for about 10000 feature files is not cool solution.

Since it is possible to override the setting of the JUnit runner class from the command line (`-Dcucumber.options="<settings>"`), there would not be any need for more than JUnit runner class, it could be called several times to run different feature files.

stanislav-antufeev:
@brasmusson I agreed with you it will be best way to implement through an option lazy loading of feature file regardless of using interface (CLI, JUnit, TestNG). I am extensively using JVM implementation of Cucumber, CLI implementation  cucumber.api.cli.Main does not provide the running feature file one by one (in term of lazy loading) because in cucumber.runtime.model.CucumberFeature:111 Runtime.java is loaded in one shot all given feature files(through tags, paths) and parsing them, populating in memory Cucumber feature instances.
Junit testing library, at lest till version 4 does not provide an interface for lazy loading of resources for test method. Only what I found is TestNG Framework has such ability to lazily loading resources for test method.
I can implement lazy loading for CLI implementation, but I am sure for JUnit it won't work.

stanislav-antufeev:
I will close this pull request because the feature has been implemented not completely.

PR #992, opened 2016-04-21
Add notes about Step class lifecycles

RichardBradley:
See http://stackoverflow.com/a/33063300/8261 -- it is not at all obvious that Step classes (or their dependencies) are destroyed and re-created for each Scenario. This doc should hopefully make that clearer.

Since the Steps are destroyed & recreated for each Scenario, it is very important to clean up any resources allocated by them (e.g. database connections) otherwise projects with lots of Scenarios will grind to a halt with resource leaks. However, the Picocontainer lifecycle annotations for this are ignored by `cucumber-jvm` since it sets `NullLifecycleStrategy`. I have added documentation here about the simplest workaround (@After annotations, which do work).

HTH

### DIFF
--- a/picocontainer/README.md
+++ b/picocontainer/README.md
@@ -1,0 +1,11 @@
+# Step dependencies
+
+The picocontainer will create singleton instances of any Step class dependencies which are constructor parameters and inject them into the Step class instances when constructing them.
+
+# Step scope and lifecycle
+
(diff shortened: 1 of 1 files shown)


dkowis:
Merged, as of c0ffacd7bc1a6fa6e91fab38d67ee483c43ea0da

RichardBradley:
@dkowis, as per my comments on #994, the commit c0ffacd is not contained in any public branch of `cucumber-jvm`, so this hasn't actually been merged yet.

PR #1002, opened 2016-05-05
Add options to the JUnit module

### LINKED ISSUE #972: Could not generate test report because of outline examples
I have following scenario outline with examples:

```
Scenario Outline: Example sale with adjustment
      Given fiscal device in normal state
      When open fiscal receipt with operator 1 and password "0000"
      And user try to sell <text> for $0.99 single price in Tax group "A" with quantity of 2 with adjustment <type> of <value>
      Examples:
        | text                 | type                      | value |
        | "discount by perc"   | "DISCOUNT_BY_PERCENTAGE"  | 10.0  |
        | "discount by value"  | "DISCOUNT_BY_SUM"         | 0.10  |
        | "surcharge by perc"  | "SURCHARGE_BY_PERCENTAGE" | 10.0  |
        | "surcharge by value" | "SURCHARGE_BY_SUM"        | 0.10  |
```

and defined test class as follow:

```
@RunWith(Cucumber.class)
@CucumberOptions(
        plugin = {"pretty", "html:target/cucumber-html-report"},
        features = "src/test/resources/"
)
public class SalesTest {

}
```

Then my build failed because of:
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':devices:test'.
> Could not generate test report to 'D:\Intellij\DatecsFiscalFramework\devices\build\reports\tests'.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 17.053 secs
D:\Intellij\DatecsFiscalFramework\devices\build\reports\tests\packages\| "discount by perc" | "DISCOUNT_BY_PERCENTAGE" | 10.html (The filename, directory name, or volume label syntax is incorrect)
16:08:25: External task execution finished 'build'.

```

I saw there was and --expand option but my project is configured with 1.2.4. How can make my build not failing and optionally to have successfully generated report.

```
dependencies {
    ...
    testCompile 'info.cukes:cucumber-java:1.2.4'
    testCompile 'info.cukes:cucumber-junit:1.2.4'
    testCompile 'junit:junit:4.12'
}
```

I'm on windows.
Thanks.

brasmusson:
The PR adds the mechanisms to have options for the JUnit module. To avoid that the Core module need to know about every option of the JUnit module, the JUnit options is passed through the Core module classes the same way that GCC passes option through to the pre-processor, linker etc. A JUnit option on the command line:
```
    --junit,<option>
```
and in `@CucumberOptions`:
```
@CucumberOptions(junit = { <option> })
``` 

Two JUnit options are added:
- `--allow-started-ignored`
- `--filename-compatible-names`

The `--allow-started-ignored` makes the JUnit module fire the test started notification before each test step is executed. The current behaviour is to wait to fire the test started notification until after the test step has executed. This is done to be able to not fire the test started notifications for pending steps, for which only the test ignored notification is fired. The rational for the current behaviour is that there are listeners that do not expect a test started to be followed by a test ignored. For instance Eclipse (v4.5 Mars) will report the wrong number of tests executed in that case - for instance "37 out of 36 tests executed". A consequence of the current behaviour is that a listener cannot calculate the execution time of a step from the elapsed time between the test started notification and the test finished notification, see #825.

The current behaviour will still be the default behaviour with the PR, but the `--allow-started-ignored` can be used when it is important that the elapsed time between the test started notification and the test finished notification reflects the execution time of the steps. Closes #825.

The `--filename-compatible-names` makes the JUnit module change the names used for "Execution units"/test cases and test steps in the notifications, so that they on all platforms can be used in file names (the names are actually constrained to [A-Za-z0-9_]). For instance Gradle will use the names of the test cases in report file names, and especially Windows cannot handle all characters in filenames, see #972. In some cases also names of the test steps are used for file names, see http://stackoverflow.com/questions/25174141/could-not-generate-test-report-in-gradle-due-to-cucumber-step-syntax. Fixes #972.

### DIFF
--- a/core/src/main/java/cucumber/api/CucumberOptions.java
+++ b/core/src/main/java/cucumber/api/CucumberOptions.java
@@ -64,4 +64,10 @@ public @interface CucumberOptions {
      * @return what format should the snippets use. underscore, camelcase
      */
     SnippetType snippets() default SnippetType.UNDERSCORE;
+
+    /**
+     * @return the options for the JUnit runner
+     */
+    String[] junit() default {};
+
 }
--- a/core/src/main/java/cucumber/runtime/RuntimeOptions.java
+++ b/core/src/main/java/cucumber/runtime/RuntimeOptions.java
@@ -39,6 +39,7 @@ public class RuntimeOptions {
     private final List<String> pluginFormatterNames = new ArrayList<String>();
     private final List<String> pluginStepDefinitionReporterNames = new ArrayList<String>();
     private final List<String> pluginSummaryPrinterNames = new ArrayList<String>();
+    private final List<String> junitOptions = new ArrayList<String>();
     private final PluginFactory pluginFactory;
     private final List<Object> plugins = new ArrayList<Object>();
     private boolean dryRun;
@@ -100,6 +101,7 @@ public class RuntimeOptions {
         List<Object> parsedFilters = new ArrayList<Object>();
         List<String> parsedFeaturePaths = new ArrayList<String>();
         List<String> parsedGlue = new ArrayList<String>();
+        List<String> parsedJunitOptions = new ArrayList<String>();
 
         while (!args.isEmpty()) {
             String arg = args.remove(0).trim();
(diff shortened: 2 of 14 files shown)


dkowis:
This is an excellent solution to those two pull requests! I'm gonna merge it!

clicman:
Wow!!! That`s great! Thank you!

PR #1004, opened 2016-05-14
Add support for spring boot's @SpringApplicationConfiguration

KamikaZeeFu:
Adds support for spring boot's `@SpringApplicationConfiguration` which saves users from having to specify the context loader via `@ContextConfiguration(loader = SpringApplicationContextLoader.class)` when using spring boot.

This PR also updates the used spring dependencies to 4.2.6 (up from 4.1.6) since that was required to be able to test the spring boot support. I manually verified that cucumber-spring continues to work with projects using just spring 4.1.6 (and no spring boot), but the only way to add an automated test would be to add a new submodule, which in my opinion is not worth it.


KamikaZeeFu:
I just saw that #791 basically added soft support for the basic use case (since `@SpringApplicationConfiguration` is a meta annotation).
However, the logic in #791 doesn't work if you use more than the single meta annotation (e.g. `@SpringApplicationConfiguration` together with `@WebIntegrationTest`).

I could adapt the logic used in #791 to work with these cases, or this PR could be kept since they have different goals.

aantono:
This is unlikely to be needed anymore after Spring Boot 1.4 has been released and has completely overhauled the Spring Boot Test foundation with the introduction of `@SpringBootTest` annotation and support for custom `@BootstrapWith` with `SpringBootTestContextBootstrapper`, etc.  See #1061

mpkorstanje:
As `@SpringApplicationConfiguration` and `@WebIntegration` it doesn't seem wise to add support for this. With #1245 we'll support the proper way.

PR #1010, opened 2016-05-29
Allow global registration of custom XStream converters.

rankinc:
This lets us provide converters for any classes that we cannot annotate directly with `@XStreamConverter`. For example,

```
@XStreamConverters(
    @XStreamConverter(MyConverter.class),
    @XStreamConverter(MyOtherConverter.class)
)
@CucumberOptions(...)
@RunWith(Cucumber.class)
public class RunCukesTest {
}
```

The existing `@XStreamConverter` annotations are being reused for consistency. The `RuntimeOptionsFactory` will check for these annotations on the same class that has the `@CucumberOptions` annotation, which keeps this change nicely internal to `cucumber-core`.

### DIFF
--- a/core/src/main/java/cucumber/runtime/xstream/LocalizedXStreams.java
+++ b/core/src/main/java/cucumber/runtime/xstream/LocalizedXStreams.java
@@ -1,15 +1,19 @@
 package cucumber.runtime.xstream;
 
 import cucumber.deps.com.thoughtworks.xstream.XStream;
+import cucumber.deps.com.thoughtworks.xstream.annotations.XStreamConverter;
 import cucumber.deps.com.thoughtworks.xstream.converters.Converter;
 import cucumber.deps.com.thoughtworks.xstream.converters.ConverterLookup;
+import cucumber.deps.com.thoughtworks.xstream.converters.ConverterMatcher;
 import cucumber.deps.com.thoughtworks.xstream.converters.ConverterRegistry;
 import cucumber.deps.com.thoughtworks.xstream.converters.SingleValueConverter;
 import cucumber.deps.com.thoughtworks.xstream.core.DefaultConverterLookup;
+import cucumber.runtime.CucumberException;
 import cucumber.runtime.ParameterInfo;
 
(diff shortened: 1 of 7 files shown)


eduardmanas:
Nice one @rankinc! 

This will enable using classes such as `org.joda.time.DateTime` without the need to annotate the method parameters with `@Tranform` all the time (which is a pain when the class is very common in your domain).

As an alternative, did you consider using `@Transformer` instead of `@XStreamConverters`?  For example, we could have re-used the `@Transformer` annotation to annotate all `Transformer` implementations that should be automatically registered at start up.

```
@Transformer
public class DateTimeTransformer extends Transformer<DateTime> {
    DateTimeFormatter dateTimeFormatter = DateTimeFormat.forPattern("yyyy-MM-dd HH:mm");
    @Override
    public DateTime transform(String value) {
        return dateTimeFormatter.parseDateTime(value);
    }
}
```

rankinc:
I did consider using `@Transform` instead of `@XStreamConverter`, yes. The problem is that `Transformer` extends `SingleValueConverter`, but my `DataTable` has more than one column.

core/src/main/java/cucumber/runtime/xstream/LocalizedXStreams.java
```diff
@@ -37,6 +48,21 @@ private LocalizedXStream newXStream(Locale locale) {
         return new LocalizedXStream(classLoader, lookup, lookup, locale);
     }
 
+    private void registerExtraConverters(LocalizedXStream xStream) {
+        for (XStreamConverter converter : extraConverters) {
+            try {
+                ConverterMatcher matcher = converter.value().newInstance();
+                if (matcher instanceof Converter) {
+                    xStream.registerConverter((Converter) matcher, converter.priority());
+                }
```
davidyscott: I just came across this in attempting to use this feature (which is excellent!) It looks like `SingleValueConverter`s are treated separately from other `Converters` and are not type-compatible, so in order to properly use `SingleValueConverter`s with the global registration, there would need to be a second if statement here to handle them.

PR #1013, opened 2016-06-06
include tags as subelements

### LINKED ISSUE #1012: Tags missing from JUnit reporter and formatter
I can't seem to find the tags anywhere in the JUnit output -- regardless weather I run it with JUnit runner or use the plugin `junit:target/cucumber-junit.xml`

jakub-bochenski:
fixes #1012

### DIFF
--- a/core/src/main/java/cucumber/runtime/formatter/JUnitFormatter.java
+++ b/core/src/main/java/cucumber/runtime/formatter/JUnitFormatter.java
@@ -13,6 +13,7 @@ import gherkin.formatter.model.Result;
 import gherkin.formatter.model.Scenario;
 import gherkin.formatter.model.ScenarioOutline;
 import gherkin.formatter.model.Step;
+import gherkin.formatter.model.Tag;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.Node;
@@ -34,8 +35,10 @@ import java.net.URL;
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
+import java.util.Set;
 
 class JUnitFormatter implements Formatter, Reporter, StrictAware {
     private final Writer out;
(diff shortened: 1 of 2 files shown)


mpkorstanje:
With cucumber 2.0.0 we have mode to a different api for formatters. As such I can not simply merge your pull request. If you have still need of this feature please feel free to create a new pull request.

jakub-bochenski:
you guys sure know how to encourage contributions

mpkorstanje:
Unfortunately cucumber-jvm is (or rather hasn't been) maintained by its community. As is this community currently consists of about 1.5 people. If you're willing to help us out and donate some time please hop into Slack.

https://cucumberbdd-slack-invite.herokuapp.com

brasmusson:
@jakub-bochenski On my day job we work a lot with xml files. The prime rule there is to never, ever read or write an xml file without validating it against its schema, and if the validation does not succeed, fail immediately.

Cucumber-JVM does not own the format, there is not actually an published schema, of the JUnit test result xml files. Adding new sub element would make the xml files produced by Cucumber-JVM illegal with respect to the format/schema of the JUnit test result xml files. There is a risk, maybe a very small risk, but still a risk that consumer that expects a "standard" JUnit test result xml file would fail on processing a xml file with new, unexpected sub elements.

No consumer that are designed process "standard" JUnit test result xml files in general would not how to process the new sub elements specific to Cucumber-JVM, so that data could only be used by a consumer written specifically to process the xml files from Cucumber-JVM. If Cucumber-JVM should produce xml files that includes elements for tags it seems better to define a new formatter, rather than add new sub elements, specific for Cucumber-JVM, to the JUnit formatter.

jakub-bochenski:
@brasmusson adding extra tags to XML is a normal thing to do. It's the eXtensible part of the name.

Not only in general, but in particular decorating the test report output is quite a popular thing to do. 
Example for pytest:
- https://github.com/pytest-dev/pytest/issues/2827
- https://stackoverflow.com/questions/29526054/

Talking of violating the JUnit XML schema makes no sense. The format originated in Ant, independently of JUnit. It has no official schema, and for that reason any input parser should be lenient (as most of them are).

PR #1016, opened 2016-06-18
Java parallel example

ManjuDevaraj:


mpkorstanje:
This example is a reproducer that shows a failure to execute multiple tests in parallel. Upgrading to the latest versions of `junit:4.12` and `maven-(failsafe|surefire)-plugin:2.20` and `cucumber-junit:1.2.5` should solve this.

Closing this PR as the reproducer is no longer needed.

PR #1018, opened 2016-06-21
Make Stats class public and add 'append' method into it

Illapikov:
To have a possibility to accumulate Stats through several runtimes.

### DIFF
--- a/core/src/main/java/cucumber/runtime/Runtime.java
+++ b/core/src/main/java/cucumber/runtime/Runtime.java
@@ -198,6 +198,10 @@ public class Runtime implements UnreportedStepExecutor {
         return glue;
     }
 
+    public Stats getStats() {
+        return stats;
+    }
+
     public void runBeforeHooks(Reporter reporter, Set<Tag> tags) {
         runHooks(glue.getBeforeHooks(), reporter, tags, true);
     }
--- a/core/src/main/java/cucumber/runtime/Stats.java
+++ b/core/src/main/java/cucumber/runtime/Stats.java
@@ -13,7 +13,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 
-class Stats {
+public class Stats {
     public static final long ONE_SECOND = 1000000000;
     public static final long ONE_MINUTE = 60 * ONE_SECOND;
     public static final String PENDING = "pending";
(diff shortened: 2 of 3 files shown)


aslakhellesoy:
Why?

Illapikov:
We are running separate runtimes in parallel using java ExecutorService and in the end we'd like to accumulate all stats from 'child' runtime into 'parent' runtime.

dimas:
Guys, it would be helpful to get your opinion on the subject. No offence but Cucumber is not very friendly to any extension attempt at the moment - lots of things sitting in private fields or in package-scoped classes without any access to them. We had to use reflections to work around these issues and that sucks big time.

As @Illapikov explained, we are trying to create multiple instances of Runtime, each running a single feature. In the end we need ability to collect individual feature stats from them (this is why public getStats() is needed) and build an aggregated stats (and this is the need for the public Stats.append(Stats stats))

These changes look very safe to me as we are not touching any existing functionality but please let us know if you think otherwise.

dimas:
Chaps, it would be nice to have some feedback rather than just being ignored :)

dimas:
@aslakhellesoy, is there a process we missed or something?
Your https://github.com/cucumber/cucumber-jvm/blob/master/CONTRIBUTING.md page says you love PRs and it looks to me this PR matches all the requirements listed there.... If something is missing - it would be nice to know...
Thanks

brasmusson:
Even if the `Stats` class is public it is still only part of the internal API, its the classes in `cucumber.api.` that forms the public API (that can be safely depended on, knowing that [SemVer](http://semver.org/) applies).

With the move to [Gherkin compiler and Pickles](https://github.com/cucumber/gherkin#compiler), I'm not sure that may instances of `Runtime` is the way to go with respect to parallel execution. Maybe parallel execution of Pickles coordinated by one `Runtime` - similar that is [envisioned for Cucumber-Ruby](https://github.com/cucumber/cucumber-ruby/issues/924#issuecomment-152604991) - is a better way.

I personally focus the time I have available to work on upgrading Cucumber-JVM to use Gherkin v4.0.0 (with compiler and Pickles), and am not keen on merging new changes before a Cucumber-JVM v2.0.0-SNAPSHOT, that uses Gherkin v4.0.0, is available.

mpkorstanje:
Once 2.0.0 is done you'll be able to use [Runtime.getEventBus](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/cucumber/runtime/Runtime.java#L166) to subscribe to test execution event stream. This will allow you to create your own aggregation. For an example see [Stats](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/cucumber/runtime/Stats.java).

PR #1019, opened 2016-06-21
[Core] Use UndefinedStepsTracker from provided RuntimeGlue 

Illapikov:
Passed RuntimeGlue and it's tracker is not connected with Runtime and it's tracker.

### DIFF
--- a/core/src/test/java/cucumber/runtime/RuntimeTest.java
+++ b/core/src/test/java/cucumber/runtime/RuntimeTest.java
@@ -622,12 +622,18 @@ public class RuntimeTest {
         ClassLoader classLoader = mock(ClassLoader.class);
         RuntimeOptions runtimeOptions = new RuntimeOptions(asList(runtimeArgs));
         Backend backend = mock(Backend.class);
-        RuntimeGlue glue = mock(RuntimeGlue.class);
+        RuntimeGlue glue = mockGlue();
         mockMatch(glue, match, isAmbiguous);
         mockHook(glue, hook, isBefore);
         Collection<Backend> backends = Arrays.asList(backend);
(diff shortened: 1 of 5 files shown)


core/src/test/java/cucumber/runtime/RuntimeTest.java
```diff
@@ -530,6 +531,21 @@ public void should_call_formatter_for_scenario_outline_with_two_examples_table_a
...
+        RuntimeOptions runtimeOptions = new RuntimeOptions(asList("-g", "anything", "--strict"));
+        Backend backend = mock(Backend.class);
+        Collection<Backend> backends = Arrays.asList(backend);
+        RuntimeGlue optionalGlue = new RuntimeGlue(new UndefinedStepsTracker(), new LocalizedXStreams(classLoader));
+        Runtime runtime = new Runtime(resourceLoader, classLoader, backends, runtimeOptions, optionalGlue);
+
+        optionalGlue.stepDefinitionMatch("", new Step(null, "Given ", "A", 1, null, null), ENGLISH);
+
+        assertEquals(0x1, runtime.exitStatus());
+    }
```
mpkorstanje: With the current code base no tests are executed so this won't pass.

mpkorstanje:
Nevermind. I just realized this was a PR from 2016! A belated thanks for your contribution.

brasmusson:
The exit code is now not connected at all to the `UndefinedStepsTracker`, but is solely determined by the result (status) of the test cases. So if the main issue to be fixed by this PR was to avoid getting wrong exit code when injecting a RuntimeGlue into the Runtime, that problem already gone.

PR #1021, opened 2016-06-26
Using JUnit categories to control Cucumber tests

pwielgolaski:
I'm trying to address issue from here https://groups.google.com/forum/#!topic/cukes/jVqbTdwoAPo
At the moment Cucumber JUnit runner does not work with categories as it pass filter to children like FeatureRunner, that are unaware of categories (annotation of parent)

My solution is to just filter based on Cucumber entry point runner and add or none of children tests.

### DIFF
--- a/junit/src/main/java/cucumber/api/junit/Cucumber.java
+++ b/junit/src/main/java/cucumber/api/junit/Cucumber.java
@@ -14,6 +14,8 @@ import cucumber.runtime.junit.JUnitOptions;
 import cucumber.runtime.junit.JUnitReporter;
 import cucumber.runtime.model.CucumberFeature;
 import org.junit.runner.Description;
+import org.junit.runner.manipulation.Filter;
+import org.junit.runner.manipulation.NoTestsRemainException;
 import org.junit.runner.notification.RunNotifier;
 import org.junit.runners.ParentRunner;
 import org.junit.runners.model.InitializationError;
@@ -88,6 +90,14 @@ public class Cucumber extends ParentRunner<FeatureRunner> {
     @Override
     protected Description describeChild(FeatureRunner child) {
         return child.getDescription();
+    }
+
+    @Override
+    public void filter(Filter filter) throws NoTestsRemainException {
+        // dont pass filter to children as ParentRunner does by default
+        if (!filter.shouldRun(getDescription())) {
+            children.clear();
+        }
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


pwielgolaski:
I would really love to see it in master.

roni-frantchi:
Thanks for this @pwielgolaski !

Would love to sett his in `master` myself.

@elijulian

neshkeev:
I would like to use it as well

maria-camenzuli:
This would be great.

SebastianDietrich:
We are as well waiting for this solution...

junit/src/main/java/cucumber/api/junit/Cucumber.java
```diff
@@ -91,6 +93,14 @@ protected Description describeChild(FeatureRunner child) {
     }
 
     @Override
+    public void filter(Filter filter) throws NoTestsRemainException {
+        // dont pass filter to children as ParentRunner does by default
```
mpkorstanje: This won't play well with the rerun functionality in surefire which uses filters to prune the successful tests. By not passing the filter on to the children these will not be pruned. Some more sophistication is needed.

junit/src/main/java/cucumber/api/junit/Cucumber.java
```diff
@@ -91,6 +93,14 @@ protected Description describeChild(FeatureRunner child) {
...
+        if (!filter.shouldRun(getDescription())) {
+            children.clear();
+        }
```
mpkorstanje: The `ParrentRunner` tracks a copy of the children and uses these in various methods. Simply clearing the children here is not satisfactory. Should the filter match the description the best soltuion would be to pass a filter that filters everything to super or otherwise pass the original filter.

This will seems ugly but this is an unfortunate side effect of trying to use categories in a manner they were not designed.

http://junit.org/junit4/javadoc/4.12/org/junit/experimental/categories/Categories.html
> Note that, for now, annotating suites with @Category has no effect. Categories must be annotated on the direct method or class.

mpkorstanje:
I am going to close this pull request. As mentioned above this approach is fundamentally broken.

ambition-consulting:
Maybe at least you can fix the README, since categories can **not** be expected to work.

mlvandijk:
Would you be able to provide a PR for that @ambition-consulting ?

PR #1029, opened 2016-07-25
JUnitOptions filename-compatible-names argument

bbindreiter:
Argument --no-filename-compatible-names was not working. Looking at the code revealed a possible copy&paste error


brasmusson:
Thanks.

PR #978, opened 2016-03-29
Double-check for directory exists in the ensureParentDirExists(File)

niberius:
Hi.

I have several JUnit runners which runs cucumber tests in parallel at the same time. Each runner generate JSON file with test results to the one directory, e.g. ./target/json-files/
```
@CucumberOptions(
        plugin = {"json:target/json-files/Test1.json"}, ...)
```
Sometimes I have an issue when one of runner is unable to create reports directory because another runner already done this. And it fails with error "Failed to create directory".
I think there is a make sense to double-check if reports directory is already exists (method cucumber.runtime.io.URLOutputStream#ensureParentDirExists).

PR #1032, opened 2016-07-29
Fix directory creation race condition

RichardBradley:
I'm running my Cucumber tests in parallel, to speed up my build, using the mvn failsafe plugin.

The following code throws a lot of errors at test startup. I think there's a race between the "exists" check at line 50 and the mkdirs call at line 51.


brasmusson:
Duplicate of #978.

RichardBradley:
Thanks, yep, the other PR fixes this and is better

PR #1035, opened 2016-08-04
Use Gherkin v4

brasmusson:
This PR updates Cucumber-JVM to use the Gherkin v4 parser and compiler.

Notes:
- Since this change contains non-backward compatible changes, the version is updated to 2.0.0-SNAPSHOT.
- Since Gherkin in v3/v4 has changed the group id from `info.cukes` to `io.cucumber`, the group id of Cucumber-JVM is also changed to `io.cucumber`.
- The formatter interface is re-worked completely (it had to it was previously defined in Gherkin v2), and is not based on events that are fired from inside Cucumber-JVM and that the formatters can register listeners for.
- Now when compiling features to pickles, and the to test cases, the step keyword is not available when running the test cases (basically the scenarios), there for the generated snippets always use the first step keyword of the language, and the keyword is not used when reporting the steps.
- When using notifications schemas for other testing frameworks (for instance JUnit), `Feature: <feature name>` have often been used as "class name", now the feature file path is used instead (as the feature AST is not available after compiling the feature to pickles).
- Previously the line filtering has been quite flexible, for instance if specifying the line of a step - then the scenario that the step is part of has been executed, or is specifying the line of an example table - then all scenarios from that table have been executed. Now the pickle only contain information of the line of the Scenario (or the line of the Scenario Outline and the example row that created the pickle), so that is the only line(s) that will execute the test case of the pickle. 
- The `Pickle` class in Gherkin does not have a getter for the tags of the pickle, therefore reflection has to be used to access the pickle tags.
- The `GherkinDialect` class in Gherkin does not have a getter for each keyword, therefore reflections is used to implement the `--i18n` option.
- To not lose the support to run Cucumber-JVM in OSGi containers, Gherkin need to define the Java jar as a bundle (cucumber/gherkin#221).

All in all, I do not see any blocker for merging this and creating a 2.0.0-SNAPSHOT. Before releasing a 2.0.0 version some more things (`pretty` and `html` formatters, OSGi support) are needed though.


aslakhellesoy:
You are a machine @brasmusson ! I'll do everything I can to cut a new release this week. Thanks for your awesome work.

mattwynne:
@aslakhellesoy are you the only person who can make releases? Is there anyone else with the experience / authority to merge this PR? Seems a shame for this to get stuck for so long just because you're busy.

junit/src/main/java/cucumber/runtime/junit/ExecutionUnitRunner.java
```diff
@@ -1,48 +1,43 @@
...
+    private final PickleEvent pickleEvent;
     private final JUnitReporter jUnitReporter;
+    private final Map<PickleStep, Description> stepDescriptions = new HashMap<PickleStep, Description>();
     private Description description;
-    private final Map<Step, Description> stepDescriptions = new HashMap<Step, Description>();
-    private final List<Step> runnerSteps = new ArrayList<Step>();
 
-    public ExecutionUnitRunner(Runtime runtime, CucumberScenario cucumberScenario, JUnitReporter jUnitReporter) throws InitializationError {
+    public ExecutionUnitRunner(Runner runner, PickleEvent pickleEvent, JUnitReporter jUnitReporter) throws InitializationError {
         super(ExecutionUnitRunner.class);
```
mpkorstanje: As far as I can tell this should be super(null). There is no reason to scan this class for annotations.

core/src/main/java/cucumber/api/TestCase.java
```diff
@@ -0,0 +1,63 @@
...
+        this.pickleEvent = pickleEvent;
+    }
+
+    public void run(EventBus bus) {
+        boolean skipNextStep = false;
+        bus.send(new TestCaseStarted(bus.getTime(), this));
+        ScenarioImpl scenarioResult = new ScenarioImpl(bus, pickleEvent.pickle);
+        for (TestStep step : testSteps) {
+            Result stepResult = step.run(bus, pickleEvent.pickle.getLanguage(), scenarioResult, skipNextStep);
+            if (stepResult.getStatus() != Result.PASSED) {
```
mpkorstanje: I guess you intended this to be an enum not a string?

junit/src/main/java/cucumber/api/junit/Cucumber.java
```diff
@@ -57,10 +62,10 @@ public Cucumber(Class clazz) throws InitializationError, IOException {
 
         ResourceLoader resourceLoader = new MultiLoader(classLoader);
         runtime = createRuntime(resourceLoader, classLoader, runtimeOptions);
-
+        formatter = runtimeOptions.formatter(classLoader);
```
mpkorstanje: I don't understand this. Is this needed? This also happens in Runner#buildBackendWorlds which calls  runtimeOptions.getPlugins();

core/src/main/java/cucumber/api/TestCase.java
mpkorstanje: This is also causing problems in TestNgReporter#result. It compares strings against objects for Result.UNDEFINED.

core/src/main/java/cucumber/runtime/formatter/TestSourcesModel.java
```diff
@@ -0,0 +1,181 @@
...
+import gherkin.ast.Node;
+import gherkin.ast.ScenarioDefinition;
+import gherkin.ast.ScenarioOutline;
+import gherkin.ast.Step;
+import gherkin.ast.TableRow;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class TestSourcesModel {
```
mpkorstanje: I don't think TestSourcesModel is part of the public API but I can't tell for sure. Whole thing should probably be package private.

core/src/main/java/cucumber/runtime/model/CucumberFeature.java
```diff
@@ -1,83 +1,99 @@
 package cucumber.runtime.model;
```
mpkorstanje: This package has become quite anemic. All closely tied to RuntimeOptions. Might want to consolidate this, RuntimeOptions and friends into a options package.

core/src/main/java/cucumber/runtime/table/DataTableDiff.java
```diff
@@ -0,0 +1,39 @@
...
+
+import cucumber.api.DataTable;
+import gherkin.pickles.PickleRow;
+import gherkin.pickles.PickleTable;
+
+import java.util.AbstractMap.SimpleEntry;
+import java.util.ArrayList;
+import java.util.List;
+
+public class DataTableDiff extends DataTable {
```
mpkorstanje: This and other classes are all public but don't need to be. They're not part of the public API as far as I can see.

core/src/main/java/cucumber/runtime/LinePredicate.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public LinePredicate(Map<String, List<Long>> lineFilters) {
+        this.lineFilters = lineFilters;
+    }
+
+    @Override
+    public boolean apply(PickleEvent pickleEvent) {
+        String picklePath = pickleEvent.uri;
+        if (!lineFilters.containsKey(picklePath)) {
+            return true;
```
mpkorstanje: I think this should be false. If a Pickle isn't referenced in the rerun file it shouldn't be added to the list of steps to execute.

mpkorstanje:
Reviewed the [Core] changes
 - aca512d468d96b6c120be36f3b47691dbdcf7614 
 - 3b226f65cf416007c689d11932084fdb8714143f
 - 15cf206912d711e987f1f62f4e4f20b8e2da952e
 - 6675ff8191e0de9fe518cae42cc111170252db21
 - f18c845930ec8b166c89f37cc890ebe1e9aecbc2
 - fe3a8bb1536765470da72ba8165c61ca09e705e1

Took me a while to do this but I found nothing major. This is mostly a straightforward rewrite against a changed API. There are some TODOs present and a few disabled tests but they don't seem to matter too much.

Pushed more more nitpicks to #1122. No point in writing them out.

brasmusson:
@mpkorstanje Your effort is much appreciated.

brasmusson:
> Is there a reason not to match the package structure to the groupId? E.g. io.cucumber.api rather then cucumber.api?

I do not know, when I got involved the groupId did correspond to the web site (`http://cukes.info/`), but the package structure was `cucumber.<x>`, `gherkin.<x>`. Now then the web site has moved to http://cucumber.io, the goupdId changes seems reasonable (and the newer gherkin release has changed groupId).

core/src/main/java/cucumber/api/TestCase.java
brasmusson: Now that the Result class has moved to this project from gherkin, there is an opportunity improve things like the definition of status codes. I'll look into that.

mattwynne:
> @mpkorstanje Your effort is much appreciated.

Amen!

testng/src/main/java/cucumber/api/testng/FeatureResultListener.java
```diff
@@ -1,52 +1,35 @@
...
 
-    @Override
-    public void before(Match match, Result result) {
-        collectError(result);
-        reporter.before(match, result);
-    }
-
-    @Override
-    public void embedding(String mimeType, byte[] data) {
-        reporter.embedding(mimeType, data);
```
mpkorstanje: I can not find where this method has been replaced with an event listener. From what I can see this is not needed as the Runtime takes care of this now and the TestNgReporter doesn't care about embeddings.

mpkorstanje:
Reviewed the [TestNG] changes
 - 9021844c58a3238d5ab77ed8e4e970c16aba727c

Nothing major. 

Pushed a fix for the string vs non-string comparison in the TestNgReporter to #1122.

Pushed more more nitpicks to #1122.

aslakhellesoy:
Hey @brasmusson and @mpkorstanje - awesome work on this and sorry it took so long!

core/src/main/java/cucumber/api/TestCase.java
brasmusson: By the introduction of an enum for the result statuses in #1133, this kind of problem should be eliminated.

brasmusson:
@mpkorstanje 
> I don't quite understand SanityChecker. Should probably be in test sources.

Yes, it is test related, but it is also used in the test of [picocontainer](https://github.com/cucumber/cucumber-jvm/blob/master/picocontainer/src/test/java/cucumber/runtime/java/picocontainer/SanityTest.java#L9). AFAIK it needs to be in the src/main part to be usable from the picocontainer module (on the other hand all of that testing maybe should be moved to the junit module).

core/src/main/java/cucumber/runtime/LinePredicate.java
brasmusson: I finally remembered why it should be true, see the (new) explaining comment in the [test case](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/test/java/cucumber/runtime/LinePredicateTest.java#L26-L28)

mpkorstanje: That makes sense. The runner presents each file[:line] arguments as having or semantics. But internally it first finds all feature files by looking at the file part and then reduces the set of tests with the line predicate.

Issue #1036, opened 2016-08-04
Guice InjectorSourceFactory call to Class#forName should use explicit ClassLoader


DPUkyle:
`c.r.j.guice.impl.InjectorSourceFactory` makes a call to `Class#forName(String)` to instantiate the InjectorSource named via the `cucumber.properties` file.

The call to `Class#forName` without the ClassLoader argument will use the calling class's ClassLoader to resolve the provided class name. However, if the class to be resolved is in a child ClassLoader, this call will fail to resolve the class. This fails in my enterprise app where the cucumber JARs are in the top-level ClassLoader, but the implementation of `c.r.j.guice.InjectorSource` is in a child ClassLoader.

Looking to the constructors of `c.r.j.JavaBackend` for inspiration, the typical way to "find" the ClassLoader when one is not explicitly provided is by calling `Thread.currentThread().getContextClassLoader()`.

I will issue a PR shortly with this small, one-line fix to `c.r.j.guice.impl.InjectorSourceFactory`.

This is a subset of issue #880.

aslakhellesoy:
Looks like you've done some work in your fork. Are you going to send a PR?

DPUkyle:
Hi @aslakhellesoy! Yes, please see PR #1037 .

Thanks very much.

- Kyle

mpkorstanje:
Merged. Thank you!

DPUkyle:
Thanks guys!

PR #1037, opened 2016-08-04
Issue #1036: Use Thread's context ClassLoader when looking for Inject…

DPUkyle:
…orSource impl.
- This mirrors JavaBackend's behavior when an explicit ClassLoader reference is not provided.


brendangibat:
I was looking for this exact functionality - has similar functionality been merged in to cucumber via some other path of configuration or cucumber object manipulation? 
This PR is well documented, has nice tests, and looks to be matching the cucumber-jvm style guidelines. @aslakhellesoy is there any update on reviewing this PR to get it in to a release?

mpkorstanje:
Yup. Looks good. Thanks for the review and the fix.

PR #1042, opened 2016-08-10
Fix issue in ComplexTypeWriter

RoyJacobs:
It can happen that in ComplexTypeWriter "startNode" is called more often than "setValue". In our case this happened when one of the fields was an empty Set, but perhaps there are other cases too.

In this case, startNode is called, but setValue is not. This causes the fieldNames and fieldValues lists to go out of alignment which causes all sorts of havoc when creating the DataTable.

This fix simply replaces the two separate lists with a Map, so that they are always aligned. One caveat of this approach is that if the same field name is serialized more than once, the previous value would be overwritten. However, seeing how the Writer is used, this does not seem like something that the current DataTables can handle anyway.

All existing tests pass.

### DIFF
--- a/core/src/main/java/cucumber/runtime/xstream/ComplexTypeWriter.java
+++ b/core/src/main/java/cucumber/runtime/xstream/ComplexTypeWriter.java
@@ -3,14 +3,16 @@ package cucumber.runtime.xstream;
 import cucumber.runtime.table.CamelCaseStringConverter;
 
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 
 import static java.util.Arrays.asList;
 
 public class ComplexTypeWriter extends CellWriter {
     private final List<String> columnNames;
-    private final List<String> fieldNames = new ArrayList<String>();
-    private final List<String> fieldValues = new ArrayList<String>();
+    private Map<String, String> fields = new LinkedHashMap<String, String>();
+    private String currentKey;
 
     private int nodeDepth = 0;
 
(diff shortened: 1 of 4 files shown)


JeffHoover:
I would really like to have this fix.

core/src/main/java/cucumber/runtime/xstream/ComplexTypeWriter.java
```diff
@@ -30,24 +32,24 @@ public ComplexTypeWriter(List<String> columnNames) {
...
-            return fieldValues;
+            return new ArrayList<String>(fields.values());
         }
     }
 
     @Override
     public void startNode(String name) {
         if (nodeDepth == 1) {
-            this.fieldNames.add(name);
+            currentKey = name;
```
mpkorstanje: This is not going to work with nested nodes.

mpkorstanje: Specifically when the nodes add values after the nested element. E.g.: `[value, [value, value], value]` because the current key will be null.

RoyJacobs: Ok, I can take a look at fixing this! Since the current tests all pass I think some more are needed. Would it be possible for you to add one or more failing tests that show the behaviour? I can try adding them myself but I don't have enough knowledge of all the various use cases to support.

mpkorstanje: I've send a PR on your PR.

RoyJacobs: Great, I hope to take a look when I'm back at work on Monday.

RoyJacobs:
@mpkorstanje I just merged your changes into my PR. It looks fine from my end, I can confirm that your fix works equally well as my fix for our use case.

mpkorstanje:
Cheers! Thanks for the setup.

> One caveat of this approach is that if the same field name is serialized more than once, the previous value would be overwritten. However, seeing how the Writer is used, this does not seem like something that the current DataTables can handle anyway.

And this fortunately is no longer a caveat. We are only including the fields of the root object and not their children. These are guaranteed to be unique.

PR #1045, opened 2016-08-13
Add unit tests to reproduce issue #1044 as well as proposed solution.

alayor:
This unit tests try to use picocontainer to resolve constructors for **String** and java primitive type wrapper classes (Integer, Double, Float, etc).

I propose to create a class called **ConstructorParameters** which can return a constructor parameters for these type of classes which can help picocontainer to identify the constructor that will be used.

This class can be used in the method **PicoFactory.start** to send those constructor parameters to the method **pico.addComponent()**.


aslakhellesoy:
This should be a PR to PicoContainer, not Cucumber. However, I'm pretty sure it would be rejected there too. Let's continue discussion on gitter.

PR #1048, opened 2016-08-20
Add unit tests for ConstantPoolTypeIntrospector class.

alayor:
This is related to changes made in PR #914.
I thought that it could be a good idea to add **unit tests** for those changes.
This allowed me to do a little refactor to method  **getLambdaTypeString()**.

I hope you can take a look at the changes.

Thanks.

### DIFF
--- a/java8/src/main/java/cucumber/runtime/java8/ConstantPoolTypeIntrospector.java
+++ b/java8/src/main/java/cucumber/runtime/java8/ConstantPoolTypeIntrospector.java
@@ -44,7 +44,7 @@ public class ConstantPoolTypeIntrospector implements TypeIntrospector {
 
     String getLambdaTypeString(ConstantPool constantPool) {
         int index = constantPool.getSize();
-        while (--index > 0) {
+        while (index-- > 0) {
             try {
                 return constantPool.getMemberRefInfoAt(index)[2];
             } catch (IllegalArgumentException | IndexOutOfBoundsException e) {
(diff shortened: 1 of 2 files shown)


java8/src/main/java/cucumber/runtime/java8/ConstantPoolTypeIntrospector.java
```diff
@@ -42,21 +42,16 @@
...
-    private String getLambdaTypeString(ConstantPool constantPool) {
-        int size = constantPool.getSize();
-        String[] memberRef = null;
-
-        // find last element in constantPool with valid memberRef
-        // - previously always at size-2 index but changed with JDK 1.8.0_60
-        for (int i = size - 1; i > -1; i--) {
+    String getLambdaTypeString(ConstantPool constantPool) {
+        int index = constantPool.getSize();
+        while (index-- > 0) {
```
brasmusson: Personally I find a `for` loop clearer when iterate using an index. I try to avoid side effects when checking conditions

java8/src/main/java/cucumber/runtime/java8/ConstantPoolTypeIntrospector.java
```diff
@@ -42,21 +42,16 @@
         return typeArguments;
     }
 
-    private String getLambdaTypeString(ConstantPool constantPool) {
-        int size = constantPool.getSize();
-        String[] memberRef = null;
-
-        // find last element in constantPool with valid memberRef
-        // - previously always at size-2 index but changed with JDK 1.8.0_60
-        for (int i = size - 1; i > -1; i--) {
```
brasmusson: Remove the comment? I'm not sure, the comment is dated since it refers to the but fix, but this code has not been released, should there be a problem, it might be useful to know why there is a loop here.

PR #1057, opened 2016-09-19
Add cucumber-metrics (init of project)

sgrillon14:
Cucumber Metrics report allows for the use of your methods ( java) Cucumber.

Cucumber Metrics use java annotations:

Sample with @Time

@Time
@Given("^me a hello, please. Best Regards '(.*)'.$")
public void hello(String name) throws InterruptedException, InstantiationException, IllegalAccessException {
logger.info("Hello " + name + "!");
}

Every time your scenarios using the method "hello" you get the following trace:

Cucumber Metrics TimedInterceptor invoke method public void your.package.your.class.hello(java.lang.String) throws java.lang.InterruptedException,java.lang.InstantiationException,java.lang.IllegalAccessException is called on your.package.your.class$$EnhancerByGuice$$9aa0e4ca@cda4919 with args [Ljava.lang.Object;@7f4d9395


sgrillon14:
documentation is available here: https://github.com/sgrillon14/cucumber-jvm/tree/master/metrics

sgrillon14:
I add a regulator in cucumber_metrics.

You can see documentation available here: https://github.com/sgrillon14/cucumber-jvm/tree/master/metrics

sgrillon14:
cucumber-metrics is stable and operational for production.

sgrillon14:
You need write this code in your project because you can use more one mudule. You can inject any injector by any modules.

Add guice.injector-source=com.yourcompany.metrics.regulator.injector.CucumberMetricsInjectorSource in src\test\resources\cucumber.properties (in your project)

```
public class CucumberMetricsInjectorSource implements InjectorSource {
    @Override
    public Injector getInjector() {
        return Guice.createInjector(Stage.PRODUCTION, CucumberModules.SCENARIO, new SpeedRegulatorModule(), new MyOtherModule());
    }
}
```

metrics/src/main/java/cucumber/metrics/regulator/module/SpeedRegulatorModule.java
```diff
@@ -0,0 +1,49 @@
+package cucumber.metrics.regulator.module;
```
aantono: Shouldn't this be placed in the `guice` sub-project as it is a Guice-specific functionality (I can see someone adding similar support for Spring, etc).

sgrillon14: I do not understand your PR, can you rephrase what you want, please? and what files?

aantono: I was saying that `SpeedRegulatorModule.java` probably belongs to `guice` sub-project under `cucumber-jvm` as it is Guice-specific, instead of being included in `metrics`.  If one were to add support for metrics for Spring IOC, how would that look like?  Spring also support AOP interceptors, so the same interceptor can be wired by `SpringFactory` from `spring` sub-project to add `metrics` support as well.

aantono: The point I'm trying to make is that the `metrics` sub-project is suppose to provide generic functionality to be able to meter the execution of the tests to get the time performance metrics.  The `SpeedRegulatorModule` class is using internally Google Guice specific functionality to inject the metrics components into the IOC.  Given the layout of the cucumber-jvm project, there are dedicated sub-projects to support integration of functions into various frameworks like Spring, Guice, Pico, etc, so IMHO the code that deals specifically with wiring things using Guice belongs in `guice` sub-project, to keep dependency separation clean.  This way `TimedInterceptor` can be used for all the different IOC containers such as Guice and Spring and Pico, etc to add AOP support for metric.

metrics/src/main/java/cucumber/metrics/jmx/PremierMBean.java
```diff
@@ -0,0 +1,13 @@
+package cucumber.metrics.jmx;
+
+public interface PremierMBean {
+
+    public String getNom();
```
aantono: Cucumber-JVM uses English as a standard for code naming, so probably, IMHO, best to rename to use English words.

sgrillon14: Yes, this part is not finished. I use english lang in final version.

sgrillon14: I commited tonight a stable JMX version on https://github.com/sgrillon14/cucumber-jvm/tree/metrics/metrics/src/main/java/cucumber/metrics/jmx

sgrillon14:
back to http://apache.claz.org/

sgrillon14:
I use it in production on the Orange telephone operator's servers. You can find all explication on https://github.com/sgrillon14/cucumber-jvm/tree/metrics/metrics

mpkorstanje:
While this project looks useful there are several reasons it will not be included:

1. We currently have insufficient maintainers as it is for the current code base. 
2. This project depends on a specific dependency injection mechanism. This makes it not generally useful.

It does appear however that you have very few dependencies on cucumber. As such I would recommend that you publish this module on your own. If you need any help with publishing maven artifacts feel free to hop into the slack channel.

PR #1058, opened 2016-09-21
Add support for the Nashorn Javascript engine

The-Alchemist:
- forked the Rhino code
- requires Java 8
- fixed tests


The-Alchemist:
Anything stopping this from getting merged?  It's very isolated and doesn't impact the rest of the build.

mpkorstanje:
At the moment we're a bit short handed. We might get to it later on.

PR #1059, opened 2016-09-24
Handle whitespace in rerun file

midopa:
If I write a rerun file manually using vi on a Mac and pass it to Cucumber as an option (`@rerun.txt`), it fails to run because it can't find the feature files. Even though the same text in a rerun file generated by Cucumber works.

The error is because there was a newline to the end of the rerun file. The rerun file generated by Cucumber does not have this.

It seems Cucumber shouldn't care whether or not lines have newlines or any other kind of whitespace.

### DIFF
--- a/core/src/main/java/cucumber/runtime/model/CucumberFeature.java
+++ b/core/src/main/java/cucumber/runtime/model/CucumberFeature.java
@@ -64,6 +64,10 @@ public class CucumberFeature {
             String source = builder.read(resource);
             if (!source.isEmpty()) {
                 for (String featurePath : source.split(" ")) {
+                    featurePath = featurePath.trim();
+                    if (featurePath.length() == 0) {
+                        continue;
+                    }
                     loadFromFileSystemOrClasspath(builder, resourceLoader, featurePath, filters);
                 }
             }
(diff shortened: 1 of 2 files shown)


midopa:
Sure thing, @haroon-sheikh. I've updated my remote branch with a unit test for this. Sorry for the delay.

midopa:
2/3 builds passed. The [OpenJDK env failed](https://travis-ci.org/cucumber/cucumber-jvm/jobs/219860478), but it appears to be unrelated to my change:

> [FATAL] Non-resolvable parent POM for info.cukes:pax-exam:[unknown-version]: Could not transfer artifact info.cukes:cucumber-jvm:pom:1.2.5-SNAPSHOT from/to codehaus-snapshots (https://nexus.codehaus.org/snapshots/): nexus.codehaus.org and 'parent.relativePath' points at wrong local POM @ line 4, column 13

mpkorstanje:
Using space as a separator in the rerun file is weird. Space is a valid character in file paths.

We should probably use new line as in [rerun_formatter.js](https://github.com/cucumber/cucumber-js/blob/master/src/formatter/rerun_formatter.js).

mpkorstanje:
Resolved by #1177.

You should now separate rerun paths by new lines.

Issue #893, opened 2015-08-02
Cucumber-JVM v1.2.3 (and v1.2.4) fails to build for Android


brasmusson:
As reported on the [mailing list](https://groups.google.com/forum/#!topic/cukes/tAwLqtWhjcI), Cucumber-JVM release v1.2.3 (and v1.2.4) fails to build for Android. The issue occurred when Cucumber-JVM-Deps was updated to use XStream v1.4.8, which contains two classes (com.thoughtworks.xstream.mapper.LambdaMapper.class and com.thoughtworks.xstream.converters.reflection.LambdaConverter.class) compiled to the byte code version of Java8 (major version 52), and classes of that byte code version cannot be included in an Android dex file.

cucumber/cucumber-jvm-deps#4 fixes the problem by filtering out those two classes from the Cucumber-JVM-Deps jar file.

aslakhellesoy:
I'll take a look at this. I wonder - is there a way this could have been discovered by an automated test you think?

brasmusson:
Building the integration test for cucumber-android, that is the project [examples/android/android-test](https://github.com/cucumber/cucumber-jvm/tree/master/examples/android/android-test) would have discovered this. The caveat is that to build it you need a Android SDK installed, and to run it you also need an Android emulator running (or an Android device connected). 

Now Travis have better [support for Android](http://docs.travis-ci.com/user/languages/android/) compared to when we used and Android SDK to build cucumber-android itself - but still beta they write.

I just rebased the PR where I worked on running the integration test (examples/android/android-test) and example (examples/android/cukeulator + examples/android/cukeulator-test) for Android, and as expected the issue is [disclosed on Travis](https://travis-ci.org/cucumber/cucumber-jvm/builds/73947085)

mikeholler:
@aslakhellesoy this can absolutely be discovered by an automated test. We use [Animal Sniffer](https://github.com/mojohaus/animal-sniffer) for all of our tests. It allows us to use the Java 7 language features while making sure we don't use any Java 7 libraries.

aslakhellesoy:
Animal sniffer doesn't even have a README. That smells like abandonware to me.

mikeholler:
It is not abandonware. The Codehaus website shutdown recently but the code
has been forked and is being maintained on GitHub. Unfortunately, I believe
the user doc was lost in the process. Plenty of popular projects still use
and depend on it, like square's OkHttp.

mikeholler:
[...]

PR #1060, opened 2016-09-24
Android example v1.2.5

dhoskins:
The Android example was not building or running correctly because since v1.2.3 Java 8 dependencies are included which will not work on Android.  See https://github.com/cucumber/cucumber-jvm/issues/893

This PR implements the suggestion at https://github.com/cucumber/cucumber-jvm/issues/893#issuecomment-157859872, and also updates to the latest version of Espresso.


brasmusson:
The cucumber-jvm-deps dependency is fixed by #1170.

Issue #1062, opened 2016-09-26
Unable to load java and groovy glue for the one feature file.


murphp15:
We are currently using features files which need to be mapped to some java glue files and some groovy glue files.
GroovyBackend uses the class FileResourceLoader to load in the glue and it expects the folder location as a parameter for where the glue is located.
Java on the other hand uses the ClasspathResourceLoader, it expects a package name as a parameter for where the glue is located. If I have groovy glue located at _tests-common/src/main/groovy/com/newbay/sal/test/cucumber_ and java glue located at _tests-common/src/main/java/com/newbay/sal/cucumber_, the path would look like.

```
com.newbay.sal.cucumber
tests-common/src/main/groovy/com/newbay/sal/test/cucumber
```

However the GroovyBackend does not understand the package deceleration com.newbay.sal.cucumber. It throws IllegalArgumentException("Not a file or directory: " ) from FileResourceIterator line 54. 

Solution: 
If Groovy backend can't process some glue, it simply ignores it as apposed to throwing an exception which terminates the cucumber tests. 
To achieve this we could put a try catch block within the for loop in loadGlue() in the class GroovyBackend.java. 
Example code of this can be seen below. 
This is blocking my team at the moment please let me know what you think of that idea. 

```

 @Override
    public void loadGlue(Glue glue, List<String> gluePaths) {
        this.glue = glue;
        final Binding context = shell.getContext();

        for (String gluePath : gluePaths) {
            try {
                // Load sources
                for (Resource resource : resourceLoader.resources(gluePath, ".groovy")) {
                    Script script = parse(resource);
                    runIfScript(context, script);
                }
                // Load compiled scripts
                for (Class<? extends Script> glueClass : classFinder.getDescendants(Script.class, packageName(gluePath))) {
                    try {
                        Script script = glueClass.getConstructor(Binding.class).newInstance(context);
                        runIfScript(context, script);
                    } catch (Exception e) {
                        throw new CucumberException(e);
                    }
                }
            }catch(Exception e){
                // log the exception 
                // don't re throw it. 
            }
        }
    }
```

brasmusson:
The Java Backend can all these different way to specify the glue package:
`com.newbay.sal.cucumber`
`com/newbay/sal/cucumber`
`classpath:com.newbay.sal.cucumber`
`classpath:com/newbay/sal/cucumber`

The two latter ones will also work with the Groovy Backend, so this is not a blocking issue.

murphp15:
[...]

PR #1063, opened 2016-09-26
    If groovyBackend doesn't understand glue it will no longer throw …

murphp15:
Please see the issue https://github.com/cucumber/cucumber-jvm/issues/1062


mpkorstanje:
I'm closing this issue because #1062 has also been closed.

Issue #1066, opened 2016-10-02
Add support for Hk2 Dependency Injection framework


tayloryork:
Hk2 is a popular dependency injection framework used by the popular JAX-RS implementation Jersey.

An Hk2 Object Factory could be added to cucumber-jvm to allow Cucumber Glue code to be injected via Hk2's service locator.

christianbellinaef:
Hi, any update on HK2 support? My team would really like to see this become a feature.

mpkorstanje:
Not at the moment. We're currently working getting 2.0.0 done. 

For now I'd recommend that you create your own jar.

It may be a while before we get to this issue.

mpkorstanje:
We currently have two different pull request for HK2 support. At a glance it looks like #1067 supports a only subset of the usecases of #1125 but I can't confirm this with certainty.

@tayloryork @christianbellinaef can you sort out which version should be merged?

And additionally to avoid problems like #724 could you add a `README.md` that minimally describes the possible configuration of the factory, the use of scopes if any and how it looks when used in a step definition.

mpkorstanje:
Feel free to restart this issue.

PR #1067, opened 2016-10-02
Add support for Hk2 Dependency Injection framework #1066

tayloryork:
Add support for Hk2 Dependency Injection framework #1066


mpkorstanje:
After about a year it appears this issue has gone stale. I regret that none has been able to look at this earlier. However as mentioned in #1066 there should be some consensus about which implementation is used.

Closing this to keep the tracker organized. Feel free to reopen this issue.

PR #1078, opened 2016-10-31
SnippetGenerator recognises parameters from Scenario Outline (<param>)

avokin:
## Summary
SnippetGenerator recognises outline step parameters

## Details

SnippetGenerator - added pattern for \<param\>
ArgumentPattern - added possibility to provide custom replacement string
JavaSnippetTest - test

## Motivation and Context

SnippetGenerator should suggest correct step definition for Scenario Outline Steps. At the moment cucumber-jvm creates step definition like this:

```
@Given("^Given a step with <param> from Scenario Outline$")
public void given_a_step() {
...
}
```

Fix makes the following step definition: 
```
@Given("^Given a step with (.*) from Scenario Outline$")
public void given_a_step(String arg0) {
...
}
```
We have related issue in tracker of IntelliJ IDEA: https://youtrack.jetbrains.com/issue/IDEA-160330

## How Has This Been Tested?
Fix tested like tested all other patterns for step parameters. See JavaSnippetTest

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/runtime/snippets/ArgumentPattern.java
+++ b/core/src/main/java/cucumber/runtime/snippets/ArgumentPattern.java
@@ -7,9 +7,15 @@ public class ArgumentPattern {
 
     private final Pattern pattern;
     private final Class<?> type;
+    private final String replacement;
 
     public ArgumentPattern(Pattern pattern, Class<?> type) {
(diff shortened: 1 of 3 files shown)


mpkorstanje:
Very nifty. Thank you!

PR #1081, opened 2016-11-17
Allow using Transformers in fields of complex Objects used in Data Ta…

adrianonobre:
## Summary

Allow using Transformers in fields of complex classes used in Data Tables that are unmarshalled as List<MyClass>.

## Details

This removes the need to manually call Transformers to convert each field of the complex Object from  String into the desired type. Instead, these fields could be declared using the desired type and annotated with the appropriate Transformer.

For example:

Given the following salaries:
      | Employee         | Salary        |
      | John Smith       | 100000.00 EUR |
      | John Doe         | 120000.00 CAD |

\@Given("^the following salaries:$")
public void withSalaries(List<Salary> salaries) {...}

private class Salary {
     String employeeName;
     \@Transform(MoneyTransformer.class)
     Money salary;
     ...
}

## Motivation and Context

Many applications have several domain specific data types. In Cucumber, Transformers allow for unmarshalling these data types in a reusable and simple way.

However, when these custom data types appear as columns in Data Tables they have to be represented as Strings in the corresponding fields of the DTO class. To be able to use existing Transformers in this scenario, the developer has to manually instantiate and invoke the "transform" method in the Transformer for each field.

The change here makes it possible to use Transformers in DTO fields in the same way you can them in method parameters.

## How Has This Been Tested?
Created a test TableConverterTest.converts_to_list_of_java_bean_with_custom_data_types

Also, built Cucumber JVM with the changes and applied it to different existing tests where the feature file has Data Tables with a few columns holding custom data types. With the change, I was able to annotate fields in the DTO class (representing a row in the data table) and remove the code that was manually invoking Transformers to convert the String representations into the desired type.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/api/Transform.java
+++ b/core/src/main/java/cucumber/api/Transform.java
@@ -12,7 +12,7 @@ import java.lang.annotation.Target;
  * @see Transformer
  */
 @Retention(RetentionPolicy.RUNTIME)
-@Target({ElementType.PARAMETER, ElementType.ANNOTATION_TYPE})
+@Target({ElementType.PARAMETER, ElementType.ANNOTATION_TYPE, ElementType.FIELD})
 @Documented
 public @interface Transform {
     Class<? extends Transformer<?>> value();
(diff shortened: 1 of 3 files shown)


adrianonobre:
It looks as though the pom files under examples/pax-exam are still referencing 1.2.5... I'll update that.

Issue #1087, opened 2016-12-21
NoSuchMethodError thrown with Scala 2.12


randomcoder:
## Summary
When running cucumber tests with `cucumber-scala` (1.2.5) using Scala 2.12.1 or 2.12.0 a `noSuchMethodError` is raised on startup

Here is a minimal project that reproduces the behaviour: https://github.com/randomcoder/cucumber-scala-2-12

## Expected Behavior
The test cases should run

## Current Behavior
The error below is thrown before any tests are executed

~~~
Exception in thread "main" java.lang.NoSuchMethodError: cucumber.api.scala.ScalaDsl.$init$(Lcucumber/api/scala/ScalaDsl;)V
	at steps.SimpleScenarioSteps.<init>(SimpleScenarioSteps.scala:6)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at cucumber.runtime.scala.ScalaBackend$$anonfun$6.apply(ScalaBackend.scala:62)
	at cucumber.runtime.scala.ScalaBackend$$anonfun$6.apply(ScalaBackend.scala:62)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:234)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:234)
	at scala.collection.AbstractTraversable.map(Traversable.scala:104)
	at cucumber.runtime.scala.ScalaBackend.loadGlue(ScalaBackend.scala:62)
	at cucumber.runtime.Runtime.<init>(Runtime.java:92)
	at cucumber.runtime.Runtime.<init>(Runtime.java:70)
	at cucumber.runtime.Runtime.<init>(Runtime.java:66)
	at cucumber.api.cli.Main.run(Main.java:35)
	at cucumber.api.cli.Main.main(Main.java:18)
~~~

## Steps to Reproduce (for bugs)

1. In the sample project execute `sbt "run --dry-run --glue steps classpath:feature"`

[...]

PR #1091, opened 2017-01-15
Scala 2.12 compatibility

paoloambrosio:
## Summary

Fix cucumber-scala compatibility with Scala 2.12.

## Details

* cucumber-scala_2.12 is compiled as Java 8 bytecode, along with scala-calculator (and cucumber-java8)
* Scala version has been upgraded to the latest 2.12.1, using `Manifest` instead of Java reflection
* @aslakhellesoy should probably verify the code, given that it's a variation on something that previously was found not to work (see his comment removed on `ScalaDsl`)

## Motivation and Context

Issue #1087.

## How Has This Been Tested?

Existing tests and Scala example passing.

## Types of changes

## Checklist:

### DIFF
--- a/examples/android/android-test/pom.xml
+++ b/examples/android/android-test/pom.xml
@@ -7,7 +7,7 @@
         <groupId>info.cukes.android-examples</groupId>
         <artifactId>android-examples</artifactId>
         <relativePath>../pom.xml</relativePath>
-        <version>1.2.5-SNAPSHOT</version>
+        <version>1.2.6-SNAPSHOT</version>
     </parent>
 
     <artifactId>cucumber-android-test</artifactId>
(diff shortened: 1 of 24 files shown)


zarthross:
It would appear as though functions of arity0 do not work if you add an empty argument.  These worked before this PR.
```
When("""^I multiply x \* y$""") { () =>    // <---- if you take out the '() =>' it works fine.
    z = x * y
}
```
in https://github.com/lewismj/cucumber/blob/master/cucumber-plugin-example/src/test/scala/com/waioeka/sbt/MultiplicationSteps.scala

The surprising behavior is that the step 'runs' according to the runner output, but the function isn't actually executed.

InspiredIdealist:
Any chance this could be resolved/merged soon?

mpkorstanje:
This PR has become a bit  messy. I'm going to clean this up from scratch.

PR #1103, opened 2017-02-25
Add a Junit 5 Jupiter extension for running scenarios.

tenwit:
## Summary

Classes with Junit 5 @TestFactory methods can now be annotated with @ExtendWith(CucumberExtension.class). This allows them to have a stream of Jupiter DynamicTest objects injected into them.
Returning this stream from the @TestFactory allows Jupiter to run them.

## Details

JunitJupiterReporter is a dumb map of Gherkin Scenarios to Gherkin Results (it should be changed to use a Jupiter extension store; I will do that when time allows). The flow within Jupiter then becomes:
- Jupiter notices that a Stream<DynamicTest> is a parameter to a TestFactory and asks CucumberExtension to build the parameter
- CucumberExtension uses the CucumberOptions on the TestFactory class to find scenarios
- It then builds a DynamicTest for each one
- The stream of DynamicTests is returned to Jupiter
- Jupiter injects it into the TestFactory method
- The (user-written) TestFactory method optionally modifies then returns the stream of DynamicTests
- Jupiter does its thing and runs each test
- Each DynamicTest is called and tells the JunitJupterReporter about the scenario
- It then calls scenario.run() with the JunitJupiterReporter, and run() in turn passes the Result to the reporter
- Each DynamicTest then pulls the Result out of the reporter and converts it to a Jupiter result, either by throwing the error/exception, by failing a Jupiter Assumption (which marks the test as disabled, equivalent to skipped), or by doing nothing (passing).

## Motivation and Context

I want all my unit tests, integration tests and executable specifications to be run from a single call to Junit.

## How Has This Been Tested?

I have a unit test for the new JunitJupiterReporter functionality (I don't test the functionality it delegates to the base reporter).
I have a simple integration test to prove that a feature file is converted to exactly the expected DyanmicTest.
More tests would be valuable, and I will get around to them in a week or two. I'm time-poor this week :(

I ran everything on my Ubuntu dev box. mvn clean install works both from the new maven module and from the base directory.

I am using it actively on two projects.

No other code is dependent on the new module.

## Types of changes

## Checklist:

This doesn't require a change any existing documentation but, if it's accepted, additional documentation should be added. Lifting the javadoc from CucumberExtension should be enough.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ matrix:
     after_success: mvn -q deploy --settings .travis-settings.xml -Dno.gem.deploy=true -DskipTests=true -Dmaven.javadoc.skip=true
   - jdk: oraclejdk8
     script: mvn -q install -P java8
-    after_success: mvn -q deploy -P java8 -pl java8 --settings .travis-settings.xml -DskipTests=true -Dmaven.javadoc.skip=true
+    after_success: mvn -q deploy -P java8 -pl java8,jupiter --settings .travis-settings.xml -DskipTests=true -Dmaven.javadoc.skip=true
 
 env:
   global:
--- a/examples/pax-exam/calculator-api/pom.xml
+++ b/examples/pax-exam/calculator-api/pom.xml
@@ -5,7 +5,7 @@
         <groupId>info.cukes</groupId>
         <artifactId>pax-exam</artifactId>
         <relativePath>../pom.xml</relativePath>
-        <version>1.2.5-SNAPSHOT</version>
+        <version>1.2.6-SNAPSHOT</version>
     </parent>
 
     <artifactId>pax-exam-calculator-api</artifactId>
--- a/examples/pax-exam/calculator-service/pom.xml
+++ b/examples/pax-exam/calculator-service/pom.xml
@@ -5,7 +5,7 @@
         <groupId>info.cukes</groupId>
         <artifactId>pax-exam</artifactId>
         <relativePath>../pom.xml</relativePath>
-        <version>1.2.5-SNAPSHOT</version>
+        <version>1.2.6-SNAPSHOT</version>
     </parent>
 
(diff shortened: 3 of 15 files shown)


tenwit:
Whoops, apparently I needed to add the new module to the java8 profile. Building locally now with that change, will add another commit to the PR when it's tested.

tenwit:
Hmm.  Can't figure out how to get Travis to not build the jupiter module in the two JDK7 containers. I've added jupiter to the profile, obviously I need to do something else. Just let me know what and I'll get onto it.

tenwit:
Okay I think I've got it. I need to remove jupiter from the list of modules, and rely on the java8 profile to add it. Let me built it locally and I'll update the PR if it works.

tenwit:
Pretty sure that last build failure is an artefact. It failed on one JDK7 build but passed on the other. And the error is in an unrelated area:
[FATAL] Non-resolvable parent POM for info.cukes:pax-exam:[unknown-version]: Could not transfer artifact info.cukes:cucumber-jvm:pom:1.2.5-SNAPSHOT from/to codehaus-snapshots (https://nexus.codehaus.org/snapshots/): nexus.codehaus.org and 'parent.relativePath' points at wrong local POM @ line 4, column 13

tenwit:
Fixed a few poms (unrelated to the PR) in order to get the build to pass. All good now.

reinhapa:
@tenwit  is it possible to merge in the current master? I would like to also take a look into this extension by myself

mpkorstanje:
@reinhapa not at the moment. We're  preparing things for 2.0 at the moment. But you could clone https://github.com/tenwit/cucumber-jvm and build locally.

mpkorstanje:
@reinhapa, @tenwit I've updated the PR for 2.0.0. Lets continue in #1258.

Issue #1106, opened 2017-03-16
Cucumber spring does not support multi-threaded scenarios


ishmeister:
## Summary

I've set up a prototype project using cucumber-jvm-parallel-plugin and maven-surefire-plugin. I've set surefire to run parallel test classes within the same JVM (threaded).

This set up is supposed to create a JUnit test class per feature (which it does) and run them in parallel threads (which it also does). The alternative to threaded mode is to run a fork of the JVM with each feature, which obviously requires more system resources but there is complete isolation between each test. The bug I describe here does not apply in the latter case.

## Expected Behavior

Before adding the cucumber-spring dependency to maven, I get (as I expect) one instance of the StepDef class per scenario even when the scenarios are multi-threaded:

~~~
10:46:05.924 [pool-1-thread-1] INFO  c.m.test.steps.WikipediaSteps - before scenario: Search for Badger
10:46:05.929 [pool-1-thread-1] INFO  c.m.test.steps.WikipediaSteps - i am object: 2132331383
10:46:13.993 [pool-1-thread-1] INFO  c.m.test.steps.WikipediaSteps - after scenario: Search for Badger
10:46:13.993 [pool-1-thread-1] INFO  c.m.test.steps.WikipediaSteps - i am object: 2132331383

10:46:05.924 [pool-1-thread-2] INFO  c.m.test.steps.WikipediaSteps - before scenario: Search for Llama
10:46:05.929 [pool-1-thread-2] INFO  c.m.test.steps.WikipediaSteps - i am object: 1544373948
10:46:12.365 [pool-1-thread-2] INFO  c.m.test.steps.WikipediaSteps - after scenario: Search for Llama
10:46:12.365 [pool-1-thread-2] INFO  c.m.test.steps.WikipediaSteps - i am object: 1544373948
~~~

There is no conflict between scenarios. Each scenario runs in a separate browser window.

## Current Behavior

However, when I add the cucumber-spring dependency (which changes how StepDef classes are instantiated) this isolation between scenarios breaks down:

The result is that the two scenarios are now sharing an StepDef instance and therefore the WebDriver, which is causing conflicts.

~~~
10:56:49.741 [pool-1-thread-1] INFO  c.m.test.steps.WikipediaSteps - before scenario: Search for Badger
10:56:49.745 [pool-1-thread-1] INFO  c.m.test.steps.WikipediaSteps - i am object: 616523439
[...]

PR #1107, opened 2017-03-16
ISSUE-1106: Added multi-threading support to cucumber-glue scope

ishmeister:
This fixes issue #1106.


mpkorstanje:
Added tests continuing in: 

https://github.com/cucumber/cucumber-jvm/pull/1148

Issue #1001, opened 2016-05-03
enhancement: allow Java8 lambda steps to throw checked Exceptions


RichardBradley:
Any steps declared with the Java8 lambda syntax may not throw checked Exceptions:

```
    // Compile error: Unhandled exception: IOException
    Given("^I have a file with some text$", () -> {
        Files.write("my setup", new File("example.txt"), Charsets.UTF_8);
    });
```

It would be more convenient for developers if they were declared as "throws Exception"; the runtime machinery all copes with it fine, and it avoids lots of "`catch(e) { throw new RuntimeException(e) }`" noise.

This should be a simple change to the interface declarations in [`StepdefBody`](https://github.com/cucumber/cucumber-jvm/blob/0aecc4d3e4c15fc34b63f848043deba2c230e35f/java/src/main/java/cucumber/api/java8/StepdefBody.java#L5)

leodutra:
Any temporary replacement for this? Any special RuntimeException to be thrown?

RichardBradley:
There is an easy workaround for now: catch all your checked exceptions and rethrow as `RuntimeException`:

    Given("^I have a file with some text$", () -> {
        try {
            Files.write("my setup", new File("example.txt"), Charsets.UTF_8);
        } catch(e) { 
            throw new RuntimeException(e) 
        }
    });

christianhujer:
I guess the solution is to add `throws Exception` to all `accept()` methods of the `A*` interfaces in `src/main/java/cucumber/api/java8/StepdefBody.java`. And I agree it would remove a hell lot of noise.

renannprado:
any news on this?

RichardBradley:
No-one has stepped up to implement it yet. I would hope that a well written PR fix would be accepted, but I can't speak for the project maintainers. I don't plan to implement it myself in the short / medium term.

RichardBradley:
Thanks!

PR #1110, opened 2017-04-03
Implement enhancement: allow Java8 lambda steps to throw checked Exce…

christianhujer:
…ptions #1001

## Summary

Implements enhancement: allow Java8 lambda steps to throw checked Exceptions #1001

## Details

* Add tests which use `throws Exception` in their `accept()` methods.
* Change the `accept()` methods in the nested interfaces in `StepdefBody` to declare `throws Exception`.

## Motivation and Context

This change solves the problem that `accept()` methods for cucumber-java8 had to wrap their code with boiler plate in case of checked exceptions. This is described in #1001

## How Has This Been Tested?

* I changed the `accept()` method in `AnonInnerClassStepdefs` to declare `throws Exception`.
* I added a new scenario to `java8.feature` along with the corresponding step definition, which uses a method reference that declares `throws Exception`.

I ran `mvn clean install` on the top-level of `cucumber-jvm` and everything still passes.

n/a

## Types of changes

The ticket speaks of a feature. However, since `cucumber-java` supported steps with `throws Exception`, I think that `cucumber-java8` didn't is actually not a missing feature but a bug.
I can't judge whether the change counts as breaking change. Adding `throws Exception` to the `accept()` methods in `StepdefBody` is potentially breaking. However, I think that code which uses cucumber should actually never call these methods but implement these methods, they are callbacks.

## Checklist:

No change to the documentation required. An optional change could be made to show an example in https://cucumber.io/docs/reference/jvm#lambda-expressions-java-8 where the lambda calls a method that declares a checked exception.

### DIFF
--- a/java/src/main/java/cucumber/api/java8/StepdefBody.java
+++ b/java/src/main/java/cucumber/api/java8/StepdefBody.java
@@ -2,42 +2,42 @@ package cucumber.api.java8;
 
 public interface StepdefBody {
     public static interface A0 extends StepdefBody {
-        void accept();
+        void accept() throws Exception;
     }
 
     public static interface A1<T1> extends StepdefBody {
-        void accept(T1 p1);
+        void accept(T1 p1) throws Exception;
     }
 
     public static interface A2<T1, T2> extends StepdefBody {
-        void accept(T1 p1, T2 p2);
+        void accept(T1 p1, T2 p2) throws Exception;
     }
 
     public static interface A3<T1, T2, T3> extends StepdefBody {
-        void accept(T1 p1, T2 p2, T3 p3);
+        void accept(T1 p1, T2 p2, T3 p3) throws Exception;
     }
 
     public static interface A4<T1, T2, T3, T4> extends StepdefBody {
-        void accept(T1 p1, T2 p2, T3 p3, T4 p4);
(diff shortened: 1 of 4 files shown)


christianhujer:
I've looked into the output of travis-ci. I do not think that the failling travis-ci is related to the change in any way.

dkowis:
It's not, apparently we've got something referencing `nexus.codehaus.org` which is dead and gone now.

dkowis:
https://github.com/cucumber/cucumber-jvm/issues/1111 Added this for reference.

java/src/main/java/cucumber/api/java8/StepdefBody.java
```diff
@@ -2,42 +2,42 @@
 
 public interface StepdefBody {
     public static interface A0 extends StepdefBody {
-        void accept();
+        void accept() throws Exception;
```
mpkorstanje: [Utils#invoke](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/cucumber/runtime/Utils.java#L45) accepts any Throwable. To remain consistent with the regular step defintions all methods should declare throw Throwable.

mpkorstanje:
Merged through #1138.

Issue #1113, opened 2017-04-17
Running cucumber scenarios using TestNG (Scenario level)


deepnirmal:
## Summary
When we run the  Cucumber-jvm using TestNG, by default it runs each and every feature file as a single Test case. What I want is basically it should give functionality to run each and every scenario with examples as single test case.

## Expected Behavior
I want to create screenshots of the test failure and I am currently using selenium with cucumber-testng. Selenium listener is used to capture the event and based on the event it captures the screenshots. But this happens only on a feature level, not at the scenario level. 
My expectation is I want each scenario to be passed in @Test such that if any scenarios fails. It should trigger the event for test fail such that my code would capture the event.

## Current Behavior
It takes all the scenarios as a single test case. So my code is generating the screenshots if any scenarios fails in the feature and a single snap is generated for a particular feature file.
It runs the scenarios one after the other and doesn't trigger any event for the failure.

~~~
~~~

## Possible Solution
By extending the feature of 		```testNGCucumberRunner.runCucumber(cucumberFeature.getCucumberFeature()); ```
such as 
```testNGCucumberRunner.runCucumberScenario(cucumberFeature.getFeatureScenario(scenarioname))```

## Steps to Reproduce (for bugs)

1. Functionality for getting all scenarios name and examples
2. Adding a method for running cucumber scenario by passing every scenario name
3. Throw exception if the scenario name is not present
4. Run every scenario as a @Test kind 

## Context & Motivation
This will help in achieving the scenario level runtime options for the QA for handling events such as creating screenshots or onSuccess things to be done.
Feature level Test helps in solving the issues on modular level. But Scenario level test will help in achieving the sub-functionalities and various other uses cases per module.

## Your Environment

* Operating System and version: Mac

mpkorstanje:
Nothing to worry about. The current build is passing. 

A commit was pushed to the the PR to include attribution. This started a new build. Then the PR was merged and its branch deleted. All before travis could actually fetch the Branch.

srilakshmi211:
Hello Team,

 I would like to know  if this was fixed -"Run each Scenario as a TestNG @Test".

In my automation framework I was using @BeforeTest and @AfterTest annotations but It executes before and after feature and not before and after each scenario.

Looking forward for your response.

Regards,
Sri

mlvandijk:
[...]

PR #1118, opened 2017-04-26
Run each Scenario as a TestNG @Test.

corinned:
## Summary

Currently, cucumber-testng provides the ability to execute each cucumber feature file as one TestNG Test.  This fix allows you to run each Scenario as a Test.

## Details

AbstractTestNGCucumberTests.class contains a Test method with groups="cucumber-scenarios".  The DataProvider for this Test pulls all Scenarios and Scenario Outlines in order to execute each one as its own TestNG test.

A sample implementation of cucumber-scenarios is:
```java
TestNG testNG = new TestNG();
testNG.setGroups("cucumber-scenarios");
testNG.setTestClasses(new Class[]{new AbstractTestNGCucumberTests() {}.getClass()});
testNG.run();
```
## Motivation and Context

#1113 

## How Has This Been Tested?

Tests added to the existing cucumber.api.testng tests.  Additional testing done from a separate java project utilizing the updated library.

## Types of changes

## Checklist:

### DIFF
--- a/testng/src/main/java/cucumber/api/testng/AbstractTestNGCucumberTests.java
+++ b/testng/src/main/java/cucumber/api/testng/AbstractTestNGCucumberTests.java
@@ -1,5 +1,6 @@
 package cucumber.api.testng;
 
+import cucumber.runtime.model.CucumberTagStatement;
 import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.DataProvider;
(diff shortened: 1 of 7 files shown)


brasmusson:
@corinned I appreciate your effort very much. But ... The reason I don't think this PR should be merged is that it is built on outdated code, that is `cucumber.runtime.model.*`. With the upgrade of Cucumber-JVM from Gherkin v2.12.2 (see #1035) all of `cucumber.runtime.model.*` will be removed its functionality replaced by the Gherkin compiler which produces [pickles](https://github.com/cucumber/cucumber/tree/master/gherkin#pickles), which are the bases of execution.

The upgrade of Cucumber-JVM from Gherkin v2.12.2 is already long overdue, considering the Cucumber-Ruby was upgraded to a new version of Gherkin in Cucumber-Ruby v2.1.0 - which was release September 11, 2015.

If you would like to implement the feature of this PR on top of #1035, I would be happy to review and work with you on that.

corinned:
Thanks, @brasmusson, I will take a look at #1035.

prasanta-biswas:
This will be really useful if we can just run each scenario as a separate Test. Can we do this with current version?

brasmusson:
@corinned Now that the master branch has been updated to use Gherkin v4, the current implementation of this PR is obsolete, however a re-implementation on the updated master is highly appreciated. Hint: Use [Runtime.compileFeature](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/cucumber/runtime/Runtime.java#L168-L174) to get the scenarios/pickles to be executed, and execute them like the [Android executor](https://github.com/cucumber/cucumber-jvm/blob/master/android/src/main/java/cucumber/runtime/android/CucumberExecutor.java#L111) does.

Issue #1120, opened 2017-05-10
Unable to use surefire maven plugin to rerun failing tests


mpkorstanje:
## Summary

Using [surefire to rerun failing tests ](https://maven.apache.org/surefire/maven-surefire-plugin/examples/rerun-failing-tests.html) results in a stack trace.

- cucumber-java8:1.2.5
- maven-failsafe-plugin:2.20
- maven-surefire-plugin:2.20 

## Expected Behavior

When using surefire with `mvn -Dfailsafe.rerunFailingTestsCount=2 test` I expect the failing tests to be rerun.

## Current Behavior

```
org.apache.maven.surefire.testset.TestSetFailedException: Unable to create test class 'Scenario: Fail when running'
        at org.apache.maven.surefire.junit4.JUnit4Provider.executeFailedMethod(JUnit4Provider.java:385)
        at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:292)
        at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:238)
        at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:161)
Caused by: java.lang.ClassNotFoundException: Scenario: Fail when running
        at org.apache.maven.surefire.junit4.JUnit4Provider.executeFailedMethod(JUnit4Provider.java:379)
        at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:292)
        at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:238)
        at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:161)
```

## Possible Solution

Surefire parses the test description for the class to rerun and expects a class name here rather then a human readable name. Either provide a class name or help convince the surefire folks to fix this this. :)

See: [SUREFIRE-1372](https://issues.apache.org/jira/browse/SUREFIRE-1372)

## Steps to Reproduce (for bugs)

1. Setup mvn project with a failing cucumber test
2. Add a test runner that uses `@RunWith(Cucumber.class)` and `@CucumberOptions(...)`
3. Run mvn -Dfailsafe.rerunFailingTestsCount=2 test
4. Observe stacktrace.

## Context & Motivation

[...]

PR #1121, opened 2017-05-11
Use deterministic unique ids in Descriptions

mpkorstanje:
## Summary

Use deterministic unique ids in Descriptions.

## Motivation and Context

 Rerunning failed tests with Surefire requires a method to identify the tests.
 Surefire currently uses a tuple of (Class, Method) to identify failed tests.
 Cucumber-jvm uses test suites which do not consist of Classes or methods.
 Therefore another method is needed.

 JUnit provides the option to select tests that match a Description.
 Descriptions are compared using  a unique Id. To identify tests between
 different runs we should provide Descriptions with a predictable unique
 id.

 The current implementation did not suffice. While the provided cucumber
 elements are unique, they are not predictable between runs making it
 impossible to use their descriptions to rerun a failed test.

 After this change it will be possible to update Surefire to use
 descriptions to rerun failed tests.

  Related Issues:
  - https://issues.apache.org/jira/browse/SUREFIRE-1372
  - https://github.com/cucumber/cucumber-jvm/issues/1120
  - https://github.com/temyers/cucumber-jvm-parallel-plugin/issues/31

## How Has This Been Tested?

Replaced `ExecutionUnitRunnerTest#shouldPopulateRunnerStepsWithStepsUsedInStepDescriptions` with `FeatureRunnerTest#shouldPopulateDescriptionsWithStableUniqueIds` that creates the same feature twice and tests the id's are unique and predictable.

The reproducer for Surefire can be found [in my fork of Surefire](
https://github.com/mpkorstanje/maven-surefire/tree/SUREFIRE-1372/surefire-integration-tests/src/test/resources/junit47-rerun-failing-tests-with-cucumber). Running with cucumber 1.2.5 results in failed tests, with the result of this PR as 1.2.6-SNAPSHOT results in 2 flakes. 

## Types of changes

## Checklist:

### DIFF
--- a/junit/src/main/java/cucumber/runtime/junit/IdProvider.java
+++ b/junit/src/main/java/cucumber/runtime/junit/IdProvider.java
@@ -1,0 +1,56 @@
+package cucumber.runtime.junit;
+
+import java.io.Serializable;
+
+/**
+ * JUnit Descriptions require an unique id. However this id should be identical between test
+ * re-runs. This allows descriptions to be used to determine which failed tests should be rerun.
+ *
+ * IdProvider generates predictable unique ids. The feature name is used as a base to disambiguate
+ * between different features.
+ */
+class IdProvider {
+
+    private long id = 0;
+    private final String base;
+
+    IdProvider(String base) {
+        this.base = base;
+    }
+
+    Serializable next() {
+        return new Id(base, id++);
+    }
+
+    private static final class Id implements Serializable {
+
+        private final String base;
+        private final long id;
+
+        private Id(String base, long id) {
+            this.base = base;
+            this.id = id;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+
+            Id id1 = (Id) o;
+            return id == id1.id && base.equals(id1.base);
+
+        }
+
+        @Override
+        public int hashCode() {
+            int result = base.hashCode();
+            result = 31 * result + (int) (id ^ (id >>> 32));
+            return result;
+        }
+    }
+}
+
+
+
+
--- a/junit/src/main/java/cucumber/runtime/junit/ExamplesRunner.java
+++ b/junit/src/main/java/cucumber/runtime/junit/ExamplesRunner.java
@@ -14,21 +14,23 @@ import java.util.List;
 
 public class ExamplesRunner extends Suite {
     private final CucumberExamples cucumberExamples;
+    private final IdProvider idProvider;
     private Description description;
     private JUnitReporter jUnitReporter;
 
-    protected ExamplesRunner(Runtime runtime, CucumberExamples cucumberExamples, JUnitReporter jUnitReporter) throws InitializationError {
-        super(ExamplesRunner.class, buildRunners(runtime, cucumberExamples, jUnitReporter));
+    protected ExamplesRunner(Runtime runtime, CucumberExamples cucumberExamples, JUnitReporter jUnitReporter, IdProvider idProvider) throws InitializationError {
+        super(ExamplesRunner.class, buildRunners(runtime, cucumberExamples, jUnitReporter, idProvider));
         this.cucumberExamples = cucumberExamples;
         this.jUnitReporter = jUnitReporter;
+        this.idProvider = idProvider;
     }
 
-    private static List<Runner> buildRunners(Runtime runtime, CucumberExamples cucumberExamples, JUnitReporter jUnitReporter) {
+    private static List<Runner> buildRunners(Runtime runtime, CucumberExamples cucumberExamples, JUnitReporter jUnitReporter, IdProvider idProvider) {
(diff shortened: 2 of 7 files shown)


mpkorstanje:
Build failure was present since 88649720b108626d4dcbdaf601152ba7594f697d. Does not appear to be related to my changes.

junit/src/main/java/cucumber/runtime/junit/IdProvider.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private long id = 0;
+    private final String base;
+
+    IdProvider(String base) {
+        this.base = base;
+    }
+
+    Serializable next() {
+        return new Id(base, id++);
```
brasmusson: Using a sequence counter as a base for the id-object makes them unpredictable. As I understand it using
```
cucumber/examples/java/calculator/basic_arithmetic.feature:7:12
```
and
```
cucumber/examples/java/calculator/basic_arithmetic.feature:12
```
would result in different ids for the test case from the scenario on line 12 (feature file from the [java-calculator](https://github.com/cucumber/cucumber-jvm/tree/master/examples/java-calculator) example in the repo).

The id should instead be base on the uri(/path) of the feature file and the line number of the feature/scenario/etc.

mpkorstanje: We're numbering the tree of Description elements in depth first order. Because the description elements are only created once and cached afterwards this is sufficient to identify which scenario needs to be rerun.

For example:

```feature
Feature: Basic Arithmetic
  Background: A Calculator
    Given a calculator I just turned on

  Scenario: Addition
    When I add 4 and 5
    Then the result is 9

  Scenario: Another Addition
    When I add 4 and 7
    Then the result is 11
```

Results in a tree of executors that looks like this:

```
FeatureRunner => Feature: Basic Arithmetic
  |- ScenarioRunner => Scenario: Addition
  |   |- ExecutionUnitRunner => 
  |       |- Background step: Given a calculator I just turned on
  |       |- When I add 4 and 5
  |       |- Then the result is 9
  |- ScenarioRunner => Scenario: Another Addition
    |- ExecutionUnitRunner => 
        |- Background step: Given a calculator I just turned on
        |- When I add 4 and 7
        |- Then the result is 11
```
This tree can then be numbered in depth first order:
```
0. FeatureRunner => Feature: Basic Arithmetic
1.   |- ScenarioRunner => Scenario: Addition
2.  |   |- ExecutionUnitRunner => 
2.  |       |- Background step: Given a calculator I just turned on
4.  |       |- When I add 4 and 5
5.  |       |- Then the result is 9
6.  |- ScenarioRunner => Scenario: Another Addition
7.    |- ExecutionUnitRunner => 
8.        |- Background step: Given a calculator I just turned on
9.        |- When I add 4 and 7
10.       |- Then the result is 11
```

Which means that for a given feature file creating the numbering of descriptions is entirely predictable. So I guess I don't quite understand your comment.

> The id should instead be base on the uri(/path) of the feature file and the line number of the feature/scenario/etc.

Using the feature uri and the line number alone is not feasible as background steps are repeated.  A failing background step in the execution of Scenario `Another Addition` would result in both `Another Addition` and `Addition` being rerun. Their descriptions both contain a child Description that uses line 3 .

It would be possible to use the uri and line number. But then we'd need some other other identifier to disambiguate between background steps in different scenarios. So as far as I can see numbering the tree by depth first order is sufficient and simple.

mpkorstanje: After reviewing the changes in Gherkin v4 I've changed my opinion. It is indeed much better to use a line based notation.

 - FeatureRunner: feature-uri
 - ExecutionRunner: feature-uri:scenario-line
 - Step: feature-uri-scenario-line:step-line

brasmusson:
Maybe it is an idea to add a `--[no-]step-notifications` option the [options of the JUnit module](https://github.com/cucumber/cucumber-jvm/pull/1002). That way the leafs of the Description tree will be things (scenarios) that can be rerun, instead of just parts (steps) of the test cases. When running tests from for instance Eclipse it can be nice to be able to see the tree down to each step, but when interacting with surefire, and trying to get surefire to rerun failing tests, I reckon it is less helpful.

mpkorstanje:
That would certainly fix surefire reporting the steps of a scenario as tests and some other reporting oddities. 

Doesn't really impact the rerunning failing tests though. Rerunning is done by creating a tree of suites with the tests as leaves. Any tests that don't need to be rerun (e.g. that don't match the description of the failed tests) are removed. Any suites that have become empty are also removed. Then the whole tree is executed again. Because the ExecutionUnitRunner overrides the execute method the whole scenario is executed and things work as expected.

Adding `--[no-]step-notifications` would require some refactoring. To make it work as is we'd have to override the filter method in the ExecutionUnitRunner and make output of getDescription conditional. In combination with the already overriden execute method I'm thinking these should be two different classes. 

This sort of conditional behavior smells

mpkorstanje:
Because #1035 has been merged I've reimplemented this in #1134.

PR #1128, opened 2017-05-25
Don't build on openjdk, only oraclejdk 7/8

aslakhellesoy:
Wicket has been failing on Openjdk7. We can add Openjdk back later if we can fix that, or maybe remove the Wicket module


aslakhellesoy:
Red herring - openjdk is not the culprit of the broken build

PR #1122, opened 2017-05-13
[Junit] update to Gherkin5. -- review nit picks

mpkorstanje:

PR #1131, opened 2017-05-25
Use gherkin4 review nitpicks

mpkorstanje:
Lets try #1122 again.

### DIFF
--- a/core/src/main/java/cucumber/api/DataTable.java
+++ b/core/src/main/java/cucumber/api/DataTable.java
@@ -257,12 +257,8 @@ public class DataTable {
     public boolean equals(Object o) {
         if (this == o) return true;
         if (!(o instanceof DataTable)) return false;
-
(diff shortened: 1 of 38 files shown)


mpkorstanje:
Owh the its are under a separate profile. Oops.

mpkorstanje:
@aslakhellesoy and @mattwynne build is green again except for open-jdk-7.

Reverted making Match package private. This made JSONFormatterTest fail because mockito was no longer able to mock on the remote server. Unsure why it works locally.

Also removed the fix for the EnumResult status because this will be properly solved in #1133.

brasmusson:
The class TestNGCucumberRunner needs to be public for the use cases where the runner class does not inherit from AbstractTestNGCucumberTests, see https://github.com/cucumber/cucumber-jvm/tree/master/examples/java-calculator-testng. I have fixed that.

brasmusson:
Merged in ac1b214.

PR #1134, opened 2017-05-25
Use deterministic unique ids in Descriptions

### LINKED ISSUE #1120: Unable to use surefire maven plugin to rerun failing tests

mpkorstanje:
## Summary

Improve surefire integration.

## Motivation and Context
 Rerunning failed tests with Surefire requires a method to identify the tests.
 Surefire currently uses a tuple of (Class, Method) to identify failed tests.
 Cucumber-jvm uses test suites which do not consist of Classes or methods.
 Therefore another method is needed.

 JUnit provides the option to select tests that match a Description.
 Descriptions are compared using  a unique Id. To identify tests between
 different runs we should provide Descriptions with a predictable unique
 id.

 The current implementation did not suffice. While the provided cucumber
 elements are unique, they are not predictable between runs making it
 impossible to use their descriptions to rerun a failed test.

 After this change it will be possible to update Surefire to use
 descriptions to rerun failed tests.

  Related Issues:
  - https://issues.apache.org/jira/browse/SUREFIRE-1372
  - https://github.com/cucumber/cucumber-jvm/issues/1120
  - https://github.com/temyers/cucumber-jvm-parallel-plugin/issues/31

Closes #1120.

## How Has This Been Tested?

Replaced `ExecutionUnitRunnerTest#shouldPopulateRunnerStepsWithStepsUsedInStepDescriptions` with `FeatureRunnerTest#shouldPopulateDescriptionsWithStableUniqueIds` that creates the same feature twice and tests the id's are unique and predictable.

The reproducer for Surefire can be found [in my fork of Surefire](
https://github.com/mpkorstanje/maven-surefire/tree/SUREFIRE-1372/surefire-integration-tests/src/test/resources/junit47-rerun-failing-tests-with-cucumber). Running with cucumber 1.2.5 results in failed tests, with the result of this PR as 2.0.0-SNAPSHOT results in 2 flakes. 

## Types of changes

## Checklist:

### DIFF
--- a/junit/src/main/java/cucumber/runtime/junit/ExecutionUnitRunner.java
+++ b/junit/src/main/java/cucumber/runtime/junit/ExecutionUnitRunner.java
@@ -131,10 +131,10 @@ class ExecutionUnitRunner extends ParentRunner<PickleStep> {
         private final int pickleLine;
         private final int pickleStepLine;
 
-        PickleStepId(PickleEvent pickleEvent, PickleStep pickleStepLine) {
+        PickleStepId(PickleEvent pickleEvent, PickleStep pickleStep) {
             this.uri = pickleEvent.uri;
             this.pickleLine = pickleEvent.pickle.getLocations().get(0).getLine();
-            this.pickleStepLine = pickleStepLine.getLocations().get(0).getLine();
+            this.pickleStepLine = pickleStep.getLocations().get(0).getLine();
         }
 
         @Override
--- a/junit/src/main/java/cucumber/runtime/junit/ExecutionUnitRunner.java
+++ b/junit/src/main/java/cucumber/runtime/junit/ExecutionUnitRunner.java
@@ -49,7 +49,7 @@ class ExecutionUnitRunner extends ParentRunner<PickleStep> {
     public Description getDescription() {
         if (description == null) {
             String nameForDescription = getName().isEmpty() ? "EMPTY_NAME" : getName();
-            description = Description.createSuiteDescription(nameForDescription, new PickleWrapper(pickleEvent));
+            description = Description.createSuiteDescription(nameForDescription, new PickleId(pickleEvent));
 
             for (PickleStep step : getChildren()) {
                 description.addChild(describeChild(step));
@@ -68,7 +68,7 @@ class ExecutionUnitRunner extends ParentRunner<PickleStep> {
             } else {
                 testName = step.getText();
             }
-            description = Description.createTestDescription(getName(), testName, new PickleStepWrapper(step));
+            description = Description.createTestDescription(getName(), testName, new PickleStepId(pickleEvent, step));
             stepDescriptions.put(step, description);
         }
         return description;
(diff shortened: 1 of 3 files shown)


junit/src/main/java/cucumber/runtime/junit/PickleRunners.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+
+    static PickleRunner withNoStepDescriptions(cucumber.runner.Runner runner, PickleEvent pickleEvent, JUnitReporter jUnitReporter) throws InitializationError {
+        return new NoStepDescriptions(runner, pickleEvent, jUnitReporter);
+    }
+
+
+    static class WithStepDescriptions extends ParentRunner<PickleStep> implements PickleRunner {
+        private final Runner runner;
```
mpkorstanje: The reason for two separate classes is because the semantics of ParentRunner somewhat expect there to be children. It can be hacked but doing so would make things fragile.

junit/src/test/java/cucumber/runtime/junit/CucumberTest.java
```diff
@@ -42,14 +42,14 @@ public void ensureOriginalDirectory() {
     @Test
     public void finds_features_based_on_implicit_package() throws IOException, InitializationError {
         Cucumber cucumber = new Cucumber(ImplicitFeatureAndGluePath.class);
-        assertEquals(3, cucumber.getChildren().size());
+        assertEquals(2, cucumber.getChildren().size());
         assertEquals("Feature: FA", cucumber.getChildren().get(0).getName());
```
mpkorstanje: Because I inlined feature_with_same_steps_in_different_scenarios.feature

junit/src/main/java/cucumber/runtime/junit/ExecutionUnitRunner.java
```diff
@@ -93,22 +93,69 @@ protected void runChild(PickleStep step, RunNotifier notifier) {
...
-    private PickleStep step;
+    private static final class PickleStepId implements Serializable {
+        private static final long serialVersionUID = 1L;
+        private final String uri;
+        private final int pickleLine;
+        private final int pickleStepLine;
 
-    PickleStepWrapper(PickleStep step) {
-        this.step = step;
+        PickleStepId(PickleEvent pickleEvent, PickleStep pickleStepLine) {
```
brasmusson: I assume it should be `pickleStep` instead of `pickleStepLine`, I fix that.

junit/src/test/java/cucumber/runtime/junit/FeatureRunnerTest.java
```diff
@@ -89,17 +93,67 @@ public void should_call_formatter_for_scenario_outline_with_two_examples_table_a
...
+            "    | D    |\n" +
+            "    | E    |\n"
+
+        );
+
+        FeatureRunner runner = createFeatureRunner(cucumberFeature);
+        FeatureRunner rerunner = createFeatureRunner(cucumberFeature);
+
+        Set<Description> descriptions = new HashSet<Description>();
+        assertDescriptionIsUnique(runner.getDescription(), descriptions);
```
brasmusson: It is a bit unusual for an assert-method to modify its arguments, but checking the uniqueness with a set makes sence - and then the set has been create and can be used subsequently.

junit/src/test/java/cucumber/runtime/junit/FeatureRunnerTest.java
```diff
@@ -89,17 +93,67 @@ public void should_call_formatter_for_scenario_outline_with_two_examples_table_a
...
-        FeatureRunner runner = new FeatureRunner(cucumberFeature, runtime, new JUnitReporter(runtime.getEventBus(), false, new JUnitOptions(Collections.<String>emptyList())));
-        RunNotifier notifier = mock(RunNotifier.class);
-        runner.run(notifier);
-        return notifier;
+        return new FeatureRunner(cucumberFeature, runtime, new JUnitReporter(runtime.getEventBus(), false, new JUnitOptions(Collections.<String>emptyList())));
+    }
+
+
+    @Test
+    public void shouldPopulateDescriptionsWithStableUniqueIds() throws Exception {
```
brasmusson: Most test method uses the non-idomatic form of using underscore, since some/many of us contributors find them easier to read (and in tests readability is even more important than elsewhere), see discussion at https://github.com/cucumber/cucumber-jvm/pull/302#issuecomment-5267175

junit/src/main/java/cucumber/runtime/junit/ExecutionUnitRunner.java
mpkorstanje: That is correct.

Issue #577, opened 2013-08-28
maven surefire test results are wrong (cucumber-java)


BertRobbenSony:
When running cukes using surefire, an xml file is generated with the results of the run. This xml file is wrong because:
- Each step is considered a testcase. A scenario corresponds to a test class. This is confusing, certainly in situations with many features.
- When using scenario outlines, the example line itself is also considered to be a testcase.
- When using scenario outlines, the example is used as classname. This results in unusable output.

You can easily see this by running "mvn clean install" on the cucumber-jvm/examples/java-helloworld sample, and then looking at target/surefire-reports/TEST-cucumber.examples.java.helloword.RunCukesTest.xml.

To also see the output for scenario outlines, change the first scenario in helloworld.feature into

Scenario Outline: Say hello
    Given I have a hello app with "<msg>"
    When I ask it to say hi
    Then it should answer with "<msg> World"

```
Examples:
| msg |
| one |
| two |
| three |
```

Then the xml output would be

``` xml
 <testcase name="Given I have a hello app with &quot;one&quot;" classname="| one |" time="0.001"/>
  <testcase name="When I ask it to say hi" classname="| one |" time="0"/>
  <testcase name="Then it should answer with &quot;one World&quot;" classname="| one |" time="0"/>
  <testcase name="| one |" classname="| one |" time="0.001"/>
  <testcase name="Given I have a hello app with &quot;two&quot;" classname="| two |" time="0"/>
  <testcase name="When I ask it to say hi" classname="| two |" time="0"/>
  <testcase name="Then it should answer with &quot;two World&quot;" classname="| two |" time="0"/>
  <testcase name="| two |" classname="| two |" time="0.002"/>
  <testcase name="Given I have a hello app with &quot;three&quot;" classname="| three |" time="0"/>
  <testcase name="When I ask it to say hi" classname="| three |" time="0"/>
  <testcase name="Then it should answer with &quot;three World&quot;" classname="| three |" time="0"/>
  <testcase name="| three |" classname="| three |" time="0"/>
  <testcase name="Given a shopping list:" classname="Scenario: Print my shopping list" time="0"/>
  <testcase name="When I print that list" classname="Scenario: Print my shopping list" time="0.006"/>
  <testcase name="Then it should look like:" classname="Scenario: Print my shopping list" time="0.001"/>
  <testcase name="Scenario: Print my shopping list" classname="Scenario: Print my shopping list" time="0.002"/>
[...]

Issue #263, opened 2012-03-19
Eclipse JUnit reports inaccurate run count


dgradl:
Using the eclipse JUnit feature cucumber junit is over-reporting the run count by seemingly double.  For example with 4 scenarios of 3 steps (with steps not written yet) each it reports Runs: 24/12.    In terms of the double counting it seems to be a matter of the way in which JUnitReporter is sending notifier events to the EachStepNotifier.  It doesn't seem to be consistent with JUnit itself, where if a test is ignored, it does a fireTestIgnored without a fireTestStarted.

I also question whether each individual cucumber step (GIVEN, WHEN, THEN) should be counted as a run as opposed to just the scenario or example itself.    I generally expect the run count to give me a number of tests invoked (in JUnit it is a reflection of number of @Tests).   However, its really only a count of steps.  After writing several scenarios with multiple GIVEN-AND and THEN-ANDs on a project I am working on it reported 390 runs, but its really only 90 some tests executed.

mayanksahai:
hi any update on this. recently moved from old cuke4duke to cucumber-jvm and like it quite a lot but this issue is still a bottleneck. i am using teamcity for build. any deadlines yet ?

proninja:
Are there any updates or solutions? This bug is a showstopper for me, because the number of tests is wrong, when they are run by the Team City or Jenkins.

ArthurBilyk:
Hi guys,
Is there a workaround for this problem?
It's very sad that this bug won't be fixed :(

mpkorstanje:
This will be fixed with #1135.

Issue #1007, opened 2016-05-25
JUnit runner does not distinguish between pending and ignored via Assume


RichardBradley:
JUnit tests can be marked as "ignored" via the [Assume](http://junit.sourceforge.net/javadoc/org/junit/Assume.html) API. If an assumption fails, then JUnit will mark the test as "ignored". Tests in this state do not count as failures.

Cucumber tests can be in "pending" state if they have undefined steps. Tests in this state should count as a failure if "strict" is set.

The `cucumber.api.junit.Cucumber` JUnit test runner doies not distinguish between these two cases; both are either failures or not according to the "strict" setting.

I believe that the "strict setting should make "pending" tests a failure, but not make "ignored" tests a failure.

RichardBradley:
See #359 -- I think it's wrong to treat "Assume" the same as "pending": the former means "ignore without failing" whereas the latter means "fail if strict or ignore if not strict".

mpkorstanje:
This will be partially resolved in version 2.0.0 once #1142 merged.

This won't cause tests to be marked as `SKIPPED` in Cucumber. They'll be marked as `PENDING` but when non-strict is enabled but these count as skipped in JUnit terms.

> I think it's wrong to treat "Assume" the same as "pending": the former means "ignore without failing" whereas the latter means "fail if strict or ignore if not strict".

@brasmusson how do you feel about mapping `AssumptionViolatedException` to `SKIPPED`?

PR #1135, opened 2017-05-27
[JUnit]  Add --[no-]step-notifications options to JunitOptions

brasmusson:
## Summary

Improve surefire integration.

## Motivation and Context

Because cucumber presents the steps as tests to junit surefire provides 
an incorrect test counts and reports strange names. By adding
`--[no-]step-notifications` to JunitOptions this can be avoided.

  Related Issues:
  - https://github.com/cucumber/cucumber-jvm/issues/263
  - https://github.com/cucumber/cucumber-jvm/issues/577

## How Has This Been Tested?

Added PickleRunnerWithNoStepDescriptionsTest and PickleRunnerWithStepDescriptionsTest to 
ensure both work as expected.

## Types of changes

## Checklist:
- [X] I have updated the documentation accordingly.


junit/src/main/java/cucumber/runtime/junit/JUnitReporter.java
```diff
@@ -6,10 +6,13 @@
 import cucumber.api.event.TestStepFinished;
 import cucumber.api.event.TestStepStarted;
 import cucumber.runner.EventBus;
+import cucumber.runtime.junit.PickleRunners.PickleRunner;
 import gherkin.pickles.PickleStep;
-import org.junit.internal.runners.model.EachTestNotifier;
+import org.junit.internal.AssumptionViolatedException;
```
brasmusson: AFAIK this is the JUnit <4.12 class for the assume feature, [JUnit 4.12 use another class](https://github.com/cucumber/cucumber-jvm/pull/961), as we handle both classes in other places we should do it here to (see comment on usage).

junit/src/main/java/cucumber/runtime/junit/JUnitReporter.java
```diff
@@ -141,6 +152,84 @@ private void addFailure(Result result) {
...
+
+        EachTestNotifier(RunNotifier notifier, Description description) {
+            this.notifier = notifier;
+            this.description = description;
+        }
+
+        public void addFailure(Throwable targetException) {
+            if (targetException instanceof MultipleFailureException) {
+                addMultipleFailureException((MultipleFailureException) targetException);
+            } else if (targetException instanceof AssumptionViolatedException) {
```
brasmusson: The different versions of `AssumtionViolatedException`s are kept in the array [`PENDING_EXCEPTIONS`](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/cucumber/runtime/Runtime.java#L36-L38), [`isPending()`](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/cucumber/runtime/Runtime.java#L239) searches the array, it could done here to (if also removing the cast of the argument to `addFailedAssumption` - as it uses its argument it could accept a `Throwable`)

As the [`PendingException`](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/cucumber/api/PendingException.java#L4-L5) now is identified with an annotation instead of as a member of `PENDING_EXCEPTIONS`, there is a case for renaming `PENDING_EXCEPTIONS` to something with `ASSUMPTION_VIOLATED` or something - but it does not directly has anything to do with this pull request.

mpkorstanje: Mmh. Good point. I'm not happy about using junit internal here either.

mpkorstanje:
@brasmusson Fixed and I've added tests for good measure.

junit/src/main/java/cucumber/runtime/junit/JUnitReporter.java
```diff
@@ -202,7 +203,7 @@ public void fireTestFinished() {
         public void addFailure(Throwable targetException) {
             if (targetException instanceof MultipleFailureException) {
                 addMultipleFailureException((MultipleFailureException) targetException);
-            } else if (isPending(targetException)) {
+            } else if (isAssumptionViolated(targetException)) {
```
mpkorstanje: Do you intend to count the pending exception as a failure? I'm not entirely sure how pending items should be reported? I guess this should depend on the strict setting.

mpkorstanje:
@brasmusson 

I found a bug in FeatureRunner. It is not properly notifying signaling the start of a test when steps are suppressed. I'll fix it later today.

junit/src/main/java/cucumber/runtime/junit/JUnitReporter.java
brasmusson: The Cucumber-JVM statuses `passed`, `skipped`, `undefined`, `pending`, `failed` are mapped to the JUnit statuses `passed`, `skipped`, `failure`, `error`, as:
* `passed` -> `passed`
* `skipped` -> `skipped`
* `failed` -> `failure`
* `undefined`, `pending` -> `skipped` unless in strict mode where they are mapped to `failure`
We do not end up here for a `pending` result unless in strict mode, if in non-strict mode we [go straight to `fireTestIgnored`](https://github.com/cucumber/cucumber-jvm/blob/junit-no-step-notification-option/junit/src/main/java/cucumber/runtime/junit/JUnitReporter.java#L135-L146)

With this PR there is a change in how `AssumptionViolatedException`s are notified, previously they would cause a `fireTestIgnored` in non-strict mode and `fireTestFailure` in strict mode, now they will cause a `fireTestIgnored` in non-strict mode and `fileTestAssumptionFailed` in strict mode. I intended to accept that and view it as a part of the fix to #1007.

Now that I know about the `fileTestAssumptionFailed` I think it should be used instead of `fireTestIgnored` to signal "skipped" for `passed`, `skipped`, `undefined` and `pending` statuses (the latter only in non-strict mode of course), but that is another PR.

mpkorstanje: Makese sense.

mpkorstanje:
@brasmusson 

It appears there are lots of implicit assumptions in junit4/surefire about the name of tests. Somebody already reported it as #865.

So I think we're good to go here. This problem needs its own PR.

longtimeago:
@mpkorstanje Could we release a new version with this fix? Rather annoying bug :(

mpkorstanje:
Hey @longtimeago we're currently working on releasing 2.0.0. There are a few issues we have to clean up first and then we'll be good to go. This may take some time as we're working with 3 people and spare time at the moment.

You can however get our snapshot builds by using:

```
<repository>
    <id>sonatype-snapshots</id>
    <url>https://oss.sonatype.org/content/repositories/snapshots</url>
    <snapshots>
        <enabled>true</enabled>
    </snapshots>
</repository>

<dependency>
    <version>2.0.0-SNAPSHOT</version>
    <scope>test</scope>
</dependency>
```

longtimeago:
@mpkorstanje Sure, no problem :) Just wanted to know some if there is a movement towards that :)
And it's not SO annoying to use snapshots :D

PR #1136, opened 2017-05-28
Implementing support for passing the response of a step to the subsequent step

paawak:
## Summary

Implementing support for passing the response of a step to the subsequent step

This would make step-definition files stateless, and true re-use of
step-definition across feature file. I have also added an example in
examples/java-calculator.

## Motivation and Context

I was writing a cucumber based test for a web-service. The problem was, how do I pass the response of my webservice from the WHEN step to the THEN step? I did not like the idea of storing the response as an instance variable in my step-definition class. There are 2 problems:
1. It becomes inherently thread-un-safe
2. It will not work across different feature files, across different step-definition classes, as it has now become stateful

The only solution imho was to pass the response of each step to the next one. This is an attempt to do that.

## How Has This Been Tested?

I have run mvn clean install on the projects core, java, java8, junit. Also, added an ene-to-end test of how the stateless thing is supposed to work in /examples/java-calculator/src/test/resources/cucumber/examples/java/calculator/stateless_calculator.feature

## Types of changes

- [ x] New feature (non-breaking change which adds functionality).

## Checklist:

- [ x] I've added tests for my code.

### DIFF
--- a/core/src/main/java/cucumber/api/PreviousStepState.java
+++ b/core/src/main/java/cucumber/api/PreviousStepState.java
@@ -1,0 +1,17 @@
+package cucumber.api;
+
+import java.util.Optional;
+
+public class PreviousStepState {
+	
+	private final Optional<Object> responseFromPreviousStep;
+
+	public PreviousStepState(Optional<Object> responseFromPreviousStep) {
+		this.responseFromPreviousStep = responseFromPreviousStep;
+	}
+
+	public Optional<Object> getResponseFromPreviousStep() {
+		return responseFromPreviousStep;
+	}
+
+}
--- a/core/src/main/java/cucumber/api/Result.java
+++ b/core/src/main/java/cucumber/api/Result.java
@@ -4,14 +4,18 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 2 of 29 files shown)


aslakhellesoy:
This is really cool!

I've been thinking about a similar pure functional API for step definitions, not just for Cucumber-JVM but for other implementations as well (Ruby, JavaScript). See cucumber/cucumber#214.

Depending on what comes out of that discussion we'll evaluate what to do with this PR.

Now regarding your two assertions:

> It becomes inherently thread-un-safe

Every scenario has a new instance of every stepdef class, so storing state on these objects does not prevent it from being thread-safe. Cucumber is single-threaded, but even if it were to become multi-threaded, storing state in classes would still be thread-safe as long as steps for a scenario run sequentially on the same thread (I can't imagine we'll change this).

> It will not work across different feature files, across different step-definition classes, as it has now become stateful

Whether or not there are different feature files involved is irrelevant. You're partially right about step definitions, but it is possible to share state using one of the many supported DI containers.

That said, I'm all for a stateful API. Let's discuss this over at cucumber/cucumber#214.

mpkorstanje:
Closing this issue. I suspect it might come around once https://github.com/cucumber/cucumber/issues/214 gains enough momentum.

Issue #865, opened 2015-05-13
surefire-junit47 reports results against incorrect test


joedj:
I originally raised this issue against the maven-surefire-plugin because the results are different when using the surefire-junit4 provider vs the surefire-junit47 provider, but was instructed to raise it as a cucumber issue: https://issues.apache.org/jira/browse/SUREFIRE-1156

There is a minimal non-working example here:

https://github.com/joedj/surefire-test.git

The project includes two tests - Test1, a simple junit test, and Test2, a cucumber test.  Test2 prints "Test2 Output".

When run with surefire-junit4, both the tests run, and the output is correct:

```
surefire-test$ mvn clean test -Pjunit4 | grep Running
Running Test1
Running Test2
surefire-test$ grep -rl 'Test2 Output' target/surefire-reports
target/surefire-reports/Test2-output.txt
```

When run with surefire-junit47, both the tests run, but the "Running Test2" line is never printed, and the Test2 output appears in the wrong file:

```
surefire-test$ mvn clean test -Pjunit47 | grep Running
Running Test1
surefire-test$ grep -rl 'Test2 Output' target/surefire-reports
target/surefire-reports/Test1-output.txt              # <--- wrong
```

joedj:
Tibor made the following comments on the surefire issue tracker:

    I have seen same issue in JUnit project, but finding those missing among 900 test was postponing the investigation on my side.
    AFAIK there was instantiation problem with Cucumber Runner as talked about it in JUnit group. 
    Isn't any exception in the reports ?

(there is no exception in the reports)

    Cucumber is running Test2 three times and Test1 once.
    Altogether 4.
    The Cucumber changes the description to "Scenario: Scenario:" instead of class name and now it looks like Cucumber does not call testStarted() on org.junit.runner.notification.RunListener.

(I'm not sure that Test2 is actually running three times - I think that's just the odd test-counting behaviour)

    Open bug in Cucumber. Feel free to reopen with new feedback from Cucumber.

mpkorstanje:
Surefire uses [NonConcurrentRunListener#describesNewTestSet](https://github.com/apache/maven-surefire/blob/master/surefire-providers/surefire-junit47/src/main/java/org/apache/maven/surefire/junitcore/NonConcurrentRunListener.java#L120) to determine if a new file needs to be made.

This relies on the implicit class name of the description.

mpkorstanje:
[...]

PR #1137, opened 2017-05-28
[JUnit] Provide class in description

mpkorstanje:
## Summary

Surefire 47 uses class names to group test reports. To ensure the tests results
end up in the proper report we provide the class of the runner. We can't fake
this because the class must be loadable by surefire.

## Details

Surefire uses [NonConcurrentRunListener#describesNewTestSet](https://github.com/apache/maven-surefire/blob/master/surefire-providers/surefire-junit47/src/main/java/org/apache/maven/surefire/junitcore/NonConcurrentRunListener.java#L120) to determine if a new file needs to be made.

The downside is that when running manually an IDE might show the class name. But as this only works with  `--no-step-notifications` enabled this should not be a major issue as `--no-step-notifications` is some what intended for machine use.

image

The other downside is that scenarios with the same name end up as duplicate entries in the report which makes tracking back harder. However I don't expect this to happen in practice.

## Motivation and Context

See: https://github.com/cucumber/cucumber-jvm/issues/865

## How Has This Been Tested?

Added unit tests for this specific scenario and executed a manual integration test with surefire.

Each test runner now has its own file with results.

image

## Types of changes

## Checklist:

### DIFF
--- a/junit/src/main/java/cucumber/runtime/junit/PickleRunners.java
+++ b/junit/src/main/java/cucumber/runtime/junit/PickleRunners.java
@@ -1,5 +1,7 @@
 package cucumber.runtime.junit;
 
+import static org.junit.runner.Description.createTestDescription;
+
 import cucumber.runner.Runner;
 import gherkin.events.PickleEvent;
 import gherkin.pickles.PickleStep;
@@ -31,8 +33,8 @@ class PickleRunners {
     }
 
 
-    static PickleRunner withNoStepDescriptions(cucumber.runner.Runner runner, PickleEvent pickleEvent, JUnitReporter jUnitReporter) throws InitializationError {
-        return new NoStepDescriptions(runner, pickleEvent, jUnitReporter);
+    static PickleRunner withNoStepDescriptions(Class featurePath, Runner runner, PickleEvent pickleEvent, JUnitReporter jUnitReporter) throws InitializationError {
+        return new NoStepDescriptions(featurePath, runner, pickleEvent, jUnitReporter);
     }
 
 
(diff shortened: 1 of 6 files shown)


junit/src/main/java/cucumber/runtime/junit/PickleRunners.java
```diff
@@ -31,8 +33,8 @@ static PickleRunner withStepDescriptions(cucumber.runner.Runner runner, PickleEv
     }
 
 
-    static PickleRunner withNoStepDescriptions(cucumber.runner.Runner runner, PickleEvent pickleEvent, JUnitReporter jUnitReporter) throws InitializationError {
-        return new NoStepDescriptions(runner, pickleEvent, jUnitReporter);
+    static PickleRunner withNoStepDescriptions(Class featurePath, Runner runner, PickleEvent pickleEvent, JUnitReporter jUnitReporter) throws InitializationError {
```
brasmusson: `featurePath` is misleading name for the runner class.

mpkorstanje:
I could also check with surefire to see if [NonConcurrentRunListener#describesNewTestSet](https://github.com/apache/maven-surefire/blob/master/surefire-providers/surefire-junit47/src/main/java/org/apache/maven/surefire/junitcore/NonConcurrentRunListener.java#L120) could use [Description#getClassName()](https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/runner/Description.java#L310) instead.

mpkorstanje:
Which won't work of-course because because we'd still have to provide a common identifier between tests. But now a feature name could also work.

PR #1142, opened 2017-06-06
[JUnit] Use AssumptionFailed to mark scenarios/steps as skipped

brasmusson:
## Summary

Send TestAssumptionFailed notifications instead of TestIgnored notifications to mark scenarios/steps as skipped. 

## Details

Since it is only after a step or scenario has been executed we can tell if it should be considered skipped in JUnit terms, TestAssumptionFailed notifications suits better than TestIgnored notifications.

By using TestAssumptionFailed to mark scenarios/steps as skipped, TestStarted can always be fired before the scenario/step is executed, therefore the option `--allow-stared-ignored` is not longer needed.

## Motivation and Context

By using TestIgnored notifications, the passing of the TestStarted had to be delayed until after the step has executed, which is both unexpected and an unnecessary complex solution. It also caused the need for the `--allow-stared-ignored` option (which adds more complexity).

## How Has This Been Tested?

In addition to the automatic tests of the junit module, I have experimented with the java-calculator example from within Eclipse and with maven.

## Types of changes

- [X] New feature (non-breaking change which adds functionality). Since the `--allow-started-ignored` option is only deprecated it is a non-breaking change.

## Checklist:

- [X] I have updated the documentation accordingly (the help for the junit options).

### DIFF
--- a/junit/src/main/java/cucumber/runtime/junit/SkippedThrowable.java
+++ b/junit/src/main/java/cucumber/runtime/junit/SkippedThrowable.java
@@ -1,0 +1,9 @@
+package cucumber.runtime.junit;
+
+class SkippedThrowable extends Throwable {
+    private static final long serialVersionUID = 1L;
+
+    public SkippedThrowable(NotificationLevel scenarioOrStep) {
+        super(String.format("This %s is skipped", scenarioOrStep.lowerCaseName()), null, false, false);
+    }
+}
--- a/junit/src/main/java/cucumber/runtime/junit/JUnitOptions.java
+++ b/junit/src/main/java/cucumber/runtime/junit/JUnitOptions.java
@@ -12,7 +12,6 @@ public class JUnitOptions {
     private static final String OPTIONS_RESOURCE = "/cucumber/api/junit/OPTIONS.txt";
     private static String optionsText;
 
-    private boolean allowStartedIgnored = false;
     private boolean filenameCompatibleNames = false;
     private boolean stepNotifications = true;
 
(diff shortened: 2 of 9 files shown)


brasmusson:
> However I don't quite understand why exceptions from the before and after hooks are not used? Should they not fail the test?

The exceptions from before and after hooks will fail the test, through the result of the Test Case Event, which holds the worst result ([created here](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/cucumber/api/TestCase.java#L35)) from all the steps and hooks (collected by [ScenarioImpl](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/cucumber/runtime/ScenarioImpl.java))

However, different from today only the worst exception will be notified on the pickleRunnerNotifier - maybe that should be reconsidered.

mpkorstanje:
Yes. I think we should reconsider that. Perhaps by wrapping all errors in something like junits `MultipleFailureException` construction. However looking at the scale of that change I think that is best done separately.

junit/src/main/java/cucumber/runtime/junit/SkippedThrowable.java
```diff
@@ -0,0 +1,13 @@
...
+
+class SkippedThrowable extends Throwable {
+    private static final long serialVersionUID = 1L;
+
+    public SkippedThrowable() {
+        this(NotificationLevel.STEP);
+    }
+
+    public SkippedThrowable(NotificationLevel scenarioOrStep) {
+        super(String.format("This %s is skipped", scenarioOrStep.lowerCaseName()), null, false, false);
```
brasmusson: Since v1.1.6 ["JDK7 is required to build everything. Built jars should still work on JDK6" (see History.md)](https://github.com/cucumber/cucumber-jvm/blob/master/History.md#116-2014-03-24). I would be a good idea to make it explicit in the pom.

brasmusson:
Now all step exceptions are collected and at the end of the test case notified on the `pickleRunnerNotifier` (similarly to using `ErrorCollector`s in JUnit). The type of notifications are controlled by the result of the test case, so if the first step failed and the second is undefined, both will be notified using `fileTestFailure` (also in non-strict mode) since the test case as a whole get the status failed. It is not obviously the best solution, but it seems reasonable.

In a way it is consistent to the behaviour of an `ErrorCollector` when I both had a failing assert and added a `AssumptionViolatedException` to it - both were notified using `fileTestFailure`, but it is unlikely that someone have expected using an `ErrorCollector` like that.

To get a clearer view of the end state of the JUnit module, I'm thinking of merging this after #1145 which also is affecting the notification logic.

PR #1147, opened 2017-06-12
[Build] Use 1.7 as the source and target level for compilation

mpkorstanje:
## Summary

Cucumber-jvm is using some library features that were introduced in Java 7.
Setting the compilation explicitly to Java 7 should avoid any unexpected
surprises.

## Motivation and Context

Java 6 has reached its end of life a while ago and we no longer run our tests against it. Supporting it has become impractical.

## How Has This Been Tested?

Check if the automated tests were still passing. 
Check manually if the Android studio example works

## Types of changes

## Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -724,8 +724,8 @@
                     <version>3.6.2</version>
                     <configuration>
                         <encoding>UTF-8</encoding>
-                        <source>1.6</source>
-                        <target>1.6</target>
+                        <source>1.7</source>
+                        <target>1.7</target>
                         <compilerArgument>-Werror</compilerArgument>
                     </configuration>
                 </plugin>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -732,11 +732,11 @@
...
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-compiler-plugin</artifactId>
-                    <version>3.5.1</version>
+                    <version>3.6.1</version>
                     <configuration>
                         <encoding>UTF-8</encoding>
-                        <source>1.6</source>
-                        <target>1.6</target>
+                        <source>1.7</source>
+                        <target>1.7</target>
```
brasmusson: The source version should definitely be 1.7, but I do not know about the target version. From Cucumber-JVM v1.1.6 (which was quite long ago now), the intention was require java7 to built, but only java6 to use Cucumber-JVM. For instance Android supported only java6 bytecode for a long time, even thought it looks like the current Android versions and tool tests also support java7. Maybe we should stay on 1.6 as the target version.

mpkorstanje: It is not possible to use source 1.7 and target 1.6. We can use source 1.6 and target 1.6 but then we'd also have to forgo using features that were introduced in 1.7.

Anyway starting with Android 4.4 we can use Java 7 features. Though I'm having a hard time finding a good reference for that. Most of the announcements seem to have rotted away.

https://developer.android.com/guide/platform/j8-jack.html
>Android supports all Java 7 language features and a subset of Java 8 language features that vary by platform version. This page describes the new language features you can use, how to properly configure your project to use them, and any known issues you may encounter.

PR #1151, opened 2017-06-14
PicoFactory: allow for configuring custom adapters

alb-i986:
## Summary

Add the method `#addAdapter` to the API of `PicoFactory`.

## Motivation and Context

Context: customizing the DI bindings in one's own BDD project using Cucumber.
See also https://cucumber.io/blog/2015/07/08/polymorphic-step-definitions

Motivation: to provide a means to configure the container with custom adapters.
In particular, my goal was to be able to **configure a custom factory method** by passing in a `FactoryInjector`.
 
 

## Alternative implementation

As an alternative, instead of adding this method, we could just make the field `pico` protected.
Subclasses would then be able to customise the DI bindings as much as they like.

But then we would lose some degree of encapsulation.

In this case I think that the tradeoff is acceptable, but I preferred to keep it private and add a setter so as to keep the implementation consistent with the existing `#addClass`.

## How Has This Been Tested?
Small change => no need for tests.
Plus, there are no tests for `#addClass` either.  

## Types of changes

- [ x] New feature (non-breaking change which adds functionality).

## Checklist:

PS: I think it would be nice if the article https://cucumber.io/blog/2015/07/08/polymorphic-step-definitions was mentioned in the class javadoc.
I can open a new PR with that if you like.

Thanks

### DIFF
--- a/picocontainer/src/main/java/cucumber/runtime/java/picocontainer/PicoFactory.java
+++ b/picocontainer/src/main/java/cucumber/runtime/java/picocontainer/PicoFactory.java
@@ -2,6 +2,7 @@ package cucumber.runtime.java.picocontainer;
 
 import cucumber.api.java.ObjectFactory;
 import cucumber.runtime.Utils;
+import org.picocontainer.ComponentAdapter;
 import org.picocontainer.MutablePicoContainer;
 import org.picocontainer.PicoBuilder;
 
@@ -12,6 +13,7 @@ import java.util.Set;
 public class PicoFactory implements ObjectFactory {
     private MutablePicoContainer pico;
     private final Set<Class<?>> classes = new HashSet<Class<?>>();
+    private final Set<ComponentAdapter<?>> adapters = new HashSet<ComponentAdapter<?>>();
 
     public void start() {
         pico = new PicoBuilder()
@@ -20,6 +22,9 @@ public class PicoFactory implements ObjectFactory {
             .build();
         for (Class<?> clazz : classes) {
             pico.addComponent(clazz);
+        }
+        for (ComponentAdapter<?> adapter : adapters) {
+            pico.addAdapter(adapter);
         }
         pico.start();
     }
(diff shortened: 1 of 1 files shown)


picocontainer/src/main/java/cucumber/runtime/java/picocontainer/PicoFactory.java
```diff
@@ -36,6 +41,18 @@ public boolean addClass(Class<?> clazz) {
...
+    /**
+     * Allows subclasses to configure {@link org.picocontainer.PicoContainer} with custom adapters.
+     * <p>
+     * For example, subclasses can use this method to configure custom factory methods
+     * by passing a {@link org.picocontainer.injectors.FactoryInjector}.
+     *
+     * @since 2.0.0
+     */
+    public void addAdapter(ComponentAdapter<?> componentAdapter) {
+        adapters.add(componentAdapter);
```
mpkorstanje: While I see the utility of being able to configure more of the pico container I don't think that adding methods PicoFactory to be used by extending classes is the correct way to do this. There is fairly wide range of potential customizations that be applied by using the PicoBuilder and MutablePicoContainer. Supporting all these by adding more methods would be cumbersome. Additionally most users would configure only a small part of the pico container and thus only use a small of these methods. We'd en up with a large code base with little utility that mostly duplicates the PicoBuilder and Container.

So I'd rather see that the creation of the MutablePicoContainer was extracted into two protected methods. One for the creation of the PicoBuilder with the minimal settings  and one for the creation of the MutablePicoContainer. Subclasses would then be able to override either method, call the parent method, apply their modifications such as adding components and return the result.

mpkorstanje:
Based on the reasoning developed in #1221 I have revised my opinion above.

In short: to ensure we don't break more then strictly necessary the use of packages other then `cucumber.api` is discouraged. As PicoFactory is not part of the public api it is recommend that you create and maintain your own implementation of an ObjectFactory interface that uses PicoFactory.

This runs counter to the example given in [Polymorphic Step Definitions](https://cucumber.io/blog/2015/07/08/polymorphic-step-definitions) but the example can be tweaked to accommodate a custom implementation.

It may also seem like duplication of code. Though in this scenario I'm weighing this against the cost of maintaining an extension point. It only seems right that those who benefit from a special use case also carry burden of maintaining it.

PR #1153, opened 2017-06-18
[Spring] Do not share ContextCache between threads

mpkorstanje:
The `TestContextManager(Class<?>)` uses a static ContextCache. This meant
 that step definitions from different threads could be registered in
 same context multiple times when running in parallel.

 This is resolved by creating a ThreadLocal ContextCache for each
 SpringFactory.

 Related issues:
  - https://travis-ci.org/cucumber/cucumber-jvm/jobs/244215644
  - https://github.com/cucumber/cucumber-jvm/pull/1148


mpkorstanje:
I merged this one a bit earlier then I'd liked. Spring kept failing the build otherwise.

PR #1086, opened 2016-12-14
Fix and re-enable Gosu

DPUkyle:
## Summary
Re-enable Gosu! Thanks to @kausnz for supplying the clever reflection solution.

## Details
This PR updates Gosu support and re-enables the cucumber-gosu module.

I have one concern - that being, "modern" Gosu requires Java 8 at runtime. So I only expect the tests to pass when the Travis matrix runs with `oraclejdk8`. ~~Can I have some help disabling test execution on the `cucumter-gosu` module unless the `java8` profile is active?~~ My second commit skips Gosu test execution unless the `java8` profile is active.

## Motivation and Context
Gosu has been broken for a while now, see #874 

## How Has This Been Tested?
Existing tests are passing again.

## Screenshots (if appropriate):
n/a

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue).
- [ ] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).

## Checklist:

- [x] I've added tests for my code.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Issue #874, opened 2015-06-07
Fix/Upgrade Gosu


aslakhellesoy:
See #781 for details. @sayewich - any luck so far?

sebrose:
@sayewich - did you have any luck bringing Gosu back into line?

PR #1154, opened 2017-06-18
[Examples] Fix java webbit websockets selenium

mpkorstanje:
Updated dependencies and added spiffy new webdriver factory that  in
 combination with the driver-binary-downloader-maven-plugin just works.

 Related issues:
  - #1033

This fixes #1033

PR #1155, opened 2017-06-18
[Gosu] Fix and re-enable Gosu for 2.0.0

mpkorstanje:
## Summary
Re-enable Gosu!

## Details
This PR updates Gosu support and re-enables the cucumber-gosu module.

## Motivation and Context
Gosu has been broken for a while now, see #874, #1086

## How Has This Been Tested?
Existing tests are passing again.

## Types of changes

## Checklist:


.travis.yml
```diff
@@ -19,13 +19,13 @@ cache:
 matrix:
   include:
   - jdk: openjdk7
-    script: mvn -q install -P examples
+    script: mvn -q install
```
brasmusson: I was more into building the examples in all jobs in the matrix ...

mpkorstanje: That will happen now. I've setup the pom in such a way that all modules and projects that can be build, will be build. 

Comes to mind I should probably push this change into #1154.

PR #1161, opened 2017-07-02
[Core] Use the test case duration in the JUnit Formatter.

brasmusson:
## Summary

Let the JUnit Formatter use the duration of the test case result.

## Details

Let the JUnit Formatter use the duration of the test case result instead of summing up the durations of the test step results.

Also refactor the TimeService to have the methods startTime(), currentTime(), and finishTime(). These methods actually do the same thing but having them all enables/simplifies the mocking of time in tests. For example when executing a test case with two test steps the following calls to the TimeService will be issued:
```
 startTime()  - for the time stamp of the TestCaseStarted Event
 startTime()  - for the time stamp of the first TestStepStarted Event
 finishTime() - for the time stamp of the first TestStepFinished Event
 startTime()  - for the time stamp of the second TestStepStarted Event
 finishTime() - for the time stamp of the second TestStepFinished Event
 finishTime() - for the time stamp of the TestCaseFinished Event
```
By communicating the use of the time stamp in the called method, it is possible for the TimeService Stub to behave so that the difference between the time stamps of the TestCaseFinished Event and TestCaseStarted Event equals the "test step duration" * "number of test steps in the test case".
## Motivation and Context

The JUnit Formatter only uses the duration of each test case, and since the test case duration is available in the result object of the TestCaseFinished event, it is unnecessary to calculated the duration in the JUnit Formatter from the durations of the test steps.

## How Has This Been Tested?

The automated test suite has been updated to verify this behaviour. 

## Types of changes

- [X] Internal refactoring.

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/runner/TimeService.java
+++ b/core/src/main/java/cucumber/runner/TimeService.java
@@ -10,20 +10,4 @@ public interface TimeService {
         }
     };
 
-    class Stub implements TimeService {
-        private final long duration;
-        private final ThreadLocal<Long> currentTime = new ThreadLocal<Long>();
-
-        public Stub(long duration) {
-            this.duration = duration;
-        }
-
-        @Override
-        public long time() {
-            Long result = currentTime.get();
-            result = result != null ? result : 0l;
-            currentTime.set(result + duration);
-            return result;
-        }
-    }
 }
--- a/core/src/test/java/cucumber/api/TestStepTest.java
+++ b/core/src/test/java/cucumber/api/TestStepTest.java
@@ -89,9 +89,13 @@ public class TestStepTest {
 
     @Test
     public void step_execution_time_is_measured() throws Throwable {
-        Long duration = new Long(1234);
+        Long duration = 1234L;
         TestStep step = new PickleTestStep("uri", mock(PickleStep.class), definitionMatch);
         when(bus.getTime()).thenReturn(0l, 1234l);
+
+        when(bus.getTime())
+            .thenReturn(0L)
+            .thenReturn(1234L);
 
         Result result = step.run(bus, language, scenario, false);
 
--- a/core/src/test/java/cucumber/runner/TimeServiceStub.java
+++ b/core/src/test/java/cucumber/runner/TimeServiceStub.java
@@ -1,0 +1,18 @@
+package cucumber.runner;
+
+public class TimeServiceStub implements TimeService {
+    private final long duration;
+    private final ThreadLocal<Long> currentTime = new ThreadLocal<Long>();
+
+    public TimeServiceStub(long duration) {
+        this.duration = duration;
+    }
+
+    @Override
+    public long time() {
+        Long result = currentTime.get();
+        result = result != null ? result : 0l;
+        currentTime.set(result + duration);
+        return result;
+    }
+}
--- a/android/src/main/java/cucumber/runtime/android/CucumberExecutor.java
+++ b/android/src/main/java/cucumber/runtime/android/CucumberExecutor.java
@@ -6,7 +6,6 @@ import android.util.Log;
 import cucumber.api.CucumberOptions;
 import cucumber.api.StepDefinitionReporter;
 import cucumber.api.event.TestRunFinished;
-import cucumber.api.formatter.Formatter;
 import cucumber.api.java.ObjectFactory;
 import cucumber.runtime.Backend;
 import cucumber.runtime.ClassFinder;
--- a/core/src/main/java/cucumber/runtime/Runtime.java
+++ b/core/src/main/java/cucumber/runtime/Runtime.java
@@ -8,7 +8,6 @@ import cucumber.api.event.EventHandler;
 import cucumber.api.event.TestCaseFinished;
 import cucumber.api.event.TestRunFinished;
 import cucumber.api.event.TestStepFinished;
-import cucumber.api.formatter.Formatter;
 import cucumber.runner.EventBus;
 import cucumber.runner.Runner;
 import cucumber.runner.TimeService;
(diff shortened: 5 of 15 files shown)


core/src/main/java/cucumber/runner/TimeService.java
```diff
@@ -1,28 +1,70 @@
 package cucumber.runner;
 
 public interface TimeService {
-    long time();
+    // The reason to have three method, which actually is doing the same thing,
+    // is to enable stubbing of the time service in tests. With the distinction
+    // between startTime() and finishTime() it is easy for the stub to make sure
+    // the the duration between the start time and finish time of the test case,
+    // is the sum of the durations of the test steps.
```
mpkorstanje: I'd like to think of a better way. Test concerns shouldn't leak into interface definitions.

brasmusson: I agree, but the hard part is how to write tests that make sense when they are read and still can set up the time service stub to both result in the expected test step duration, and test case duration, and is not disturbed by other events sent like write events, embed events etc.

core/src/test/java/cucumber/api/TestStepTest.java
```diff
@@ -89,9 +89,12 @@ public void result_is_pending_when_step_definition_throws_pending_exception() th
...
     @Test
     public void step_execution_time_is_measured() throws Throwable {
-        Long duration = new Long(1234);
+        Long duration = 1234L;
         TestStep step = new PickleTestStep("uri", mock(PickleStep.class), definitionMatch);
-        when(bus.getTime()).thenReturn(0l, 1234l);
+
+        when(bus.getTime())
+            .thenReturn(0L)
+            .thenReturn(1234L);
```
mpkorstanje: The solution was a neat little trick in the mockito API that allows calls be be chained. The first invocation will return the result of the first `thenReturn` the second that of the second and so on.

core/src/main/java/cucumber/runner/TimeService.java
```diff
@@ -10,20 +10,4 @@ public long time() {
         }
     };
 
-    class Stub implements TimeService {
-        private final long duration;
```
mpkorstanje: I've moved this class into the test packages. Doesn't belong in our public code.

core/src/test/java/cucumber/runner/TimeServiceStub.java
```diff
@@ -0,0 +1,18 @@
+package cucumber.runner;
+
+public class TimeServiceStub implements TimeService {
+    private final long duration;
+    private final ThreadLocal<Long> currentTime = new ThreadLocal<Long>();
```
mpkorstanje: The amount of ThreadLocals I'm seeing makes me itch. We'll have to look at that time time in the future.

mpkorstanje:
Tests are failing because I didn't fix the the formatter tests yet. Right now they're being tested against the wrong structure. For the time stuff they should be tested against a stream of events rather then an execution of a feature. And the the execution should be tested against that stream of events to ensure consistency.

brasmusson:
>  they should be tested against a stream of events

I disagree, test cases should be specified in the terms of the problem domain not the solution domain. The problem domain for formatter is reports from executing feature files. The event stream is an implementation detail.

One reason I feel strong about this is what happened somewhat before I started contributing to the Cucumber OSS projects. In Gherkin v2.11.x, which then implemented the Pretty Formatter and the Json Formatter, defined test from then on the API call level. Unfortunately the test used slightly different call sequences (the infamous `step(), step(), match(), result(), match(), result()` vs `step(), match(), result(), step(), match(), result()`) , therefore it was impossible for both to work in a Cucumber-JVM implementation. And sure enough, in Cucumber-JVM v1.1.2 the Json Formatter did not work, then the Cucumber-JVM implementation was change to "fix" this, so in Cucumber-JVM v1.1.3 the Pretty Formatter did not work (to be more specific Cucumber-JVM crashed when using the Pretty Formatter - as long as any scenario had more then one step).

In much I have developed much of the testing of the formatters and the ideas behind them - and am therefore certainly biased - however for me the choice of ["problem domain" oriented tests](https://github.com/cucumber/cucumber-jvm/blob/v1.2.5/core/src/test/java/cucumber/runtime/formatter/JUnitFormatterTest.java#L70-L97) vs ["implementation domain" oriented tests](https://github.com/cucumber/cucumber-jvm/blob/v1.2.5/core/src/test/java/cucumber/runtime/formatter/JUnitFormatterTest.java#L498-L527) is simple, the latter is fragile, hard to understand, and hard to maintain (IMHO).

Currently the formatter tests uses the actual runtime and runner to run the features, and mocks the glue. It is not necessary to do it like this, but it has the advantages to also perform integrations testing so that the runtime, runner and formatter works together - a main reason not to use the actual runtime and runner would be to speed up the tests, but these test does not seem to be so slow that it is a priority. The important thing is to keep the test at the level of the problem domain ("when this feature/scenario is executed, given some additional information of results and times, then the report is expected to look like this"), and defining the event stream directly in the tests is to leave the problem domain in which the tests should be formulated.

The connects to Nat Pryce's presentation at CukeUp! (but in somewhat the other way around). His theme was to use the same formulation of the tests regardless of on which level they should connect, and then use "inverse functions" to transform to the connected API. Here we need to go from a test formulated in terms of the feature to be executed, to the event stream of the formatter API - which in not an inverse function, but rather the same function that the runtime and the runner performs.

mpkorstanje:
Mmh. I agree that testing against the an event stream would make the test hard to understand. 

Would it be acceptable to decorate the formatter under test with a formatter that changes the TestCaseFinished events to contain predictable test durations?

brasmusson:
Thanks for pushing for a better solution than the first one. Now I have done a retake from scratch, and by using a new test version of the TimeService (which is also an event listener) I got the test passing without changes to the TimeService interface.

I kept the move of `TimeService.Stub` out of the `TimeService` java file that you did. A stub fits better in the test code, however when running with maven tests of other modules cannot use classes from the test code of the Core module, so the tests of the JUnit module need to implement their own (very simple) `TimeService` stub.

mpkorstanje:
This is a very elegant solution.

I could fix the build process to also produce a test-jar artifact.

mpkorstanje:
Mmh. It is not possible to both create a test-jar and exclude it from the artifacts that will be deployed. So we'll have a to suffer a bit of code duplication until there is enough to warrant the creation of a separate module with only test code.

PR looks good. I have no objections to you merging this.

PR #1162, opened 2017-07-02
[Core] Add TestRunStarted event, let Stats handle the exit code

brasmusson:
## Summary

Add TestRunStarted event, move the exit code handling to the Stats class.

## Details

Add the TestRunStarted event and present the duration between the TestRunStarted and TestRunFinished events in the summary print out.

Move the handling of exit code and errors to the summary print out to the Stats class.

## Motivation and Context

The duration of a test step is the duration between the TestStepStarted and TestStepFinished events. The duration of a test case is the duration between the TestCaseStarted and TestCaseFinished events. 
Therefore it is natural that the duration of the test run (presented in the summary print out) is the duration between the TestRunStarted and TestRunFinished events

## How Has This Been Tested?

The automated test suite has been updated to verify this behaviour.

## Types of changes

- [X] Internal refactoring

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/runtime/Runtime.java
+++ b/core/src/main/java/cucumber/runtime/Runtime.java
@@ -1,11 +1,7 @@
 package cucumber.runtime;
 
-import cucumber.api.Pending;
-import cucumber.api.Result;
 import cucumber.api.StepDefinitionReporter;
 import cucumber.api.SummaryPrinter;
-import cucumber.api.event.EventHandler;
-import cucumber.api.event.TestCaseFinished;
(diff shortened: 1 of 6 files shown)


mpkorstanje:
Rebased to resolve conflicts

core/src/main/java/cucumber/runtime/Runtime.java
```diff
@@ -32,50 +27,17 @@
  */
 public class Runtime {
 
-    private static final String[] ASSUMPTION_VIOLATED_EXCEPTIONS = {
```
mpkorstanje: Good riddance!

PR #1164, opened 2017-07-05
Adapting the code for the optional parameters (for cucumber V2.0.0 or more)

sgrillon14:
I tested this development with the java cucumber code following:

```
    @Given("^me a hello, please. Best Regards '(.*)'.$")
    public void hello(String name, List<GherkinCondition> conditions) {
        int i = 0;
        for (GherkinCondition gherkinCondition : conditions) {
            i++;
            logger.info(String.format("  expected contition N°%d=%s", i, gherkinCondition.getExpected()));
            logger.info(String.format("  actual   contition N°%d=%s", i, gherkinCondition.getActual()));
        }
        logger.info("Hello " + name + "!");
    }

    @Given("^me a cat, please. Best Regards '(.*)'.$")
    public void hello(String name) {
        logger.info("Take my cat " + name + "!");
    }

    @Given("^me a bye, please. Best Regards '(.*)'.$")
    public void bye(String name, Map<String, String> params) {
        int i = 0;
        for (Map.Entry<String, String> entry : params.entrySet()) {
            i++;
            logger.info(String.format("  Key N°%d: %s   Value:%s", i, entry.getKey(), entry.getValue()));
        }
        logger.info("Bye " + name + "!");
    }
```


aslakhellesoy:
I'm a bit confused about this PR. Can you link to an existing issue or mailing list thread where this has been discussed and described in more detail @sgrillon14?

sgrillon14:
is it the same PR of https://github.com/cucumber/cucumber-jvm/pull/1056 but the code is very different in V2.0.0

sgrillon14:
I do not understand why there is this conflict?
```
<<<<<<< optionalparametersV2
            assertEquals("Arity mismatch: Step Definition 'withThreeParams' with pattern [some pattern] is declared with 4 parameters. However, the gherkin step has 2 arguments [4, Table:[]]. \n" +
=======
            assertEquals("Arity mismatch: Step Definition 'withTwoParams' with pattern [some pattern] is declared with 3 parameters. However, the gherkin step has 2 arguments [4, Table:[]]. \n" +
>>>>>>> master
```

aslakhellesoy:
Are there any prior discussions about why you want to add this feature? Have you explained anywhere what problem this is solving? Has anyone provided any feedback?

It's hard to understand what this is by just reading the sparse PR description.

sgrillon14:
We use cucumber-jvm in the [NoraUi](https://github.com/NoraUi/NoraUi) framework. We have any generic steps (I click on '....', I update text on '....' with '....', ...). 

**In feature file,** you can use 3 possibles syntax:

```
Given me a hello, please. Best Regards '<author>'.

Given me a hello, please. Best Regards '<author>'?
        |key|expected|actual|

Given me a hello, please. Best Regards '<author>'?
        |key|expected|actual|
        |zip|35000|<zip>|
        |city|Rennes|<city>|
```

**in java,** you finish your regex by [\\.|\\?] and The cucumber-jvm framework tolerates having one more argument if it is of type "java.util.List" or "java.util.Map" (no mandatory parameter). In our case we have created a [@Conditionned annotation](https://github.com/NoraUi/NoraUi/tree/master/src/main/java/noraui/cucumber/annotation), but the additional argument can also be used directly in the java code (it fails to test if it is null, empty, ...)

Example of our google guice interceptor (ConditionedInterceptor): https://github.com/NoraUi/NoraUi/blob/master/src/main/java/noraui/cucumber/interceptor/ConditionedInterceptor.java

```
@Conditioned
@Etantdonné("j'ai un bonjour, s'il vous plaît. Cordialement '(.*)'[\\.|\\?]")
@Given("me a hello, please. Best Regards '(.*)'[\\.|\\?]")
public void hello(String name, List<GherkinStepCondition> conditions) {
    logger.info("Hello " + name + "!");
}
```


**With this evolution we have the ability to do step with parameters that are not mandatory.**

aslakhellesoy:
I appreciate your contribution, but I don't think this is a feature we want to include in Cucumber. I don't see how it would be generally useful, and I think it is rather confusing. 

You've explained what the code does, and I understand how it works. What I don't understand is *why* you think this would be generally useful. What *problem* does this solve?

sgrillon14:
In a big company, it is useful to build a Gherkin-Cucumber common foundation. With this new feature, all peoples can use generic Gherkin step (used with or without parameters). Example:
click always on a button:
`When I click on 'myapp.MyAppHomePage-submit'.`
click on a button only if data is ADMIN:
```
When I click on 'myapp.MyAppHomePage-submit'?
    |key|expected|actual|
    |action|^\\[\\{\"user\":\"ADMIN\".*|<DataJson>|
```

in Gherkin-Cucumber common foundation, a **functionnal people (no java developer)**:
When I click on 'xxxx'.     or       When I click on 'xxxx'?

This solution is increasingly being used: 

oss-sonatype-downloads

aslakhellesoy:
A generic set of step definitions designed to be reused across projects is something I strongly discourage.

Such stepdefs always end up as UI-centric. As a result you end up with scenarios that are slow, brittle and expensive to maintain. They are also exceptionally poor as documentation. This is something we've been saying for years.

Reusable stepdefs may *seem* useful in big companies, but the reality is that it perpetuates bad practices.

I've helped dozens of big companies get themselves out of the kind of mess they end up after a year or two producing tons of UI tests.

I've been saying this for years, but it's hard to be heard among all the misinformation and bad practices out there.

Step definitions should be agnostic of the UI and describe the business process, not the UI. That's the only way to have fast, reliable and maintainable Cucumber scenarios.

tfcardoso:
too bad :( having a step like this:

    Then shares multiple texts
      | Hi there  |
      | Hello     |

would be nice to authorize this

    Then shares multiple texts

instead of doing:

    Then shares multiple texts
      | |

it would be more intuitive for testers that need to populate the steps with data.

PR #1167, opened 2017-07-06
Use the feature name as class name in reports.

brasmusson:
## Summary

Change both the JUnit Formatter and the Android Instrumentation Reporter to use the feature name as the class name in the reports.

## Details

Change both the JUnit Formatter and the Android Instrumentation Reporter to use the feature name as the class name reports. This is the behaviour in v1.2.5, but to avoid having to extract the feature name from the TestSourceRead event it was changed in the upgrade to Gherkin v4.

## Motivation and Context

To avoid a regression between v1.2.5 and v2.0.0.

## How Has This Been Tested?

The automated test suite has been update to verify this behaviour.

## Types of changes

## Checklist:

### DIFF
--- a/android/src/main/java/cucumber/runtime/android/AndroidInstrumentationReporter.java
+++ b/android/src/main/java/cucumber/runtime/android/AndroidInstrumentationReporter.java
@@ -62,7 +62,7 @@ public class AndroidInstrumentationReporter implements Formatter {
      * The collected TestSourceRead events.
      */
     private final TestSourcesModel testSources = new TestSourcesModel();
-    
+
     /**
      * The current cucumber runtime.
      */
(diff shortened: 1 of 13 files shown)


mpkorstanje:
Looks good. You might want to check your IDE settings. You're using tabs for indentation.

brasmusson:
Some of the android changes did I with Emacs (instead of Eclipse as I normally uses) while trying to run the android examples, so it seems like I have some things to do with my Emacs setup.

brasmusson:
A missed to update the TestNGFormatter, fixed in https://github.com/cucumber/cucumber-jvm/commit/2f2269fac15c5799078da7d9ae078437560c3c5d

PR #1174, opened 2017-07-08
Run a seperate TestNG test per scenario

### LINKED ISSUE #1113: Running cucumber scenarios using TestNG (Scenario level)

lucianovdveekens:
## Summary

Currently, each feature runs as one test in TestNG. This PR changes that to one test for each scenario in a feature file.

Another PR #1118 already exists, but it's based on old libraries.

## Motivation and Context

Fixes #1113 & supersedes and closes #1118

## How Has This Been Tested?

I've updated existing unit tests.

## Types of changes

## Checklist:


core/src/main/java/cucumber/runtime/Runtime.java
```diff
@@ -121,6 +121,12 @@ public void runFeature(CucumberFeature feature) {
         }
     }
 
+    public void runPickle(PickleEvent pickle) {
```
brasmusson: There should be no `runPickle` on the `Runtime`, instead the `TestNGCucumberRunner` should do what for instance the [Android executor](https://github.com/cucumber/cucumber-jvm/blob/master/android/src/main/java/cucumber/runtime/android/CucumberExecutor.java#L113) and call `runPickle` on the `Runner`.

core/src/main/java/cucumber/runtime/Runtime.java
```diff
@@ -121,6 +121,12 @@ public void runFeature(CucumberFeature feature) {
...
+        if (matchesFilters(pickle)) {
```
brasmusson: Already the data provider `provideScenarios` should filter the `PickleEvent`, otherwise it will look like they were run and passed, when in fact they were not selected for execution.

testng/src/main/java/cucumber/api/testng/CucumberFeatureWrapperImpl.java
```diff
@@ -22,6 +22,6 @@ public CucumberFeature getCucumberFeature() {
 
     @Override
     public String toString() {
-        return cucumberFeature.getGherkinFeature().getFeature().getName();
+        return cucumberFeature.getPath();
```
brasmusson: I think it still should be the feature name here (possibly surrounded by "") for the sake of consistency, we have tried to standardise on the feature name as the "classname" see #1167.

testng/src/main/java/cucumber/api/testng/TestNGCucumberRunner.java
```diff
@@ -58,7 +59,7 @@ public void runCukes() {
         }
     }
 
-    public void runCucumber(CucumberFeature cucumberFeature) {
+    public void runFeature(CucumberFeature cucumberFeature) {
```
brasmusson: Do not change this method name, it makes the PR contain breaking changes for those people that use the structure of the RunCukesByFeatureAndCompositionTest example runner in the java-calculator-testng example.

testng/src/main/java/cucumber/api/testng/TestNGCucumberRunner.java
```diff
@@ -68,6 +69,10 @@ public void runCucumber(CucumberFeature cucumberFeature) {
         }
     }
 
+    public void runScenario(PickleEvent pickle) throws Throwable {
+        runtime.runPickle(pickle);
```
brasmusson: Not using a listener as in `runCucumber` make Ambiguous and Undefined result to be missed (TestNG seems to detect exceptions from the step definitions, but not the exceptions for Ambiguous and Undefined result which are thrown internally in the runner). I'll provide a Listener class (similar to the `FeatureResultListener`) which I had to make to get to the bottom of this behaviour.

testng/src/main/java/cucumber/api/testng/TestNGCucumberRunner.java
```diff
@@ -97,4 +102,25 @@ public void finish() {
...
+    public Object[][] provideScenarios() {
+        try {
+            List<Object[]> scenarios = new ArrayList<Object[]>();
+
+            List<CucumberFeature> features = getFeatures();
+            for (CucumberFeature feature : features) {
+                List<PickleEvent> pickles = runtime.compileFeature(feature);
+
+                for (PickleEvent pickle : pickles) {
+                    scenarios.add(new Object[]{new PickleEventWrapper(pickle), new CucumberFeatureWrapperImpl(feature)});
```
brasmusson: It is here that `runtime.matchesFilters(pickle)` should be used to select the pickles to run.

testng/src/test/java/cucumber/api/testng/TestNGCucumberRunnerTest.java
```diff
@@ -30,7 +30,7 @@ public void runCukesStrict() throws Exception {
     @Test
     public void parse_error_propagated_to_testng_test_execution() throws Exception {
         testNGCucumberRunner = new ParseErrorCucumberRunner(RunCukesTest.class);
-        Object[][] features = testNGCucumberRunner.provideFeatures(); // provideFeatures() throws CucumberException
+        Object[][] features = testNGCucumberRunner.provideScenarios(); // provideFeatures() throws CucumberException
```
brasmusson: Also the comment needs to updated.

testng/src/main/java/cucumber/api/testng/TestNGCucumberRunner.java
brasmusson: So I did succeed in pushing a commit onto your pull request branch with the result listener and the update of `runScenario`, very convenient.

lucianovdveekens: I totally missed that, thanks!

lucianovdveekens: I've added a test that verifies the working of the new listener when a scenario contains undefined steps and the 'strict' flag is set.

Issue #957, opened 2016-01-28
Allow multiple step patterns on the same method


mikeholler:
I've been working with Cucumber JVM for the past couple of months and have been loving it, but there's one feature I would love to see and I think others might as well.

Basically, I'd like to do something like this:

``` java
@When(
        value = {
                "^I perform the calculation (\\d+) plus (\\d+)$",
                "^I add the numbers (\\d+) and (\\d+)$"
        }
)
void performXPlusY(int x, int y) {
    // do the step
}
```

I know it's technically possible to do this via regex, but it's very ugly, uses extra capture groups, and allows for more than the two desired patterns:

``` java
@When("^(I perform the calculation|I add the numbers) (\\d+) (plus|and) (\\d+)$")
void performXPlusY(String ignored1, int x, String ignored2, int y) {
    // do the step
}
```

What does everyone think about this? Is it something that would be easy enough to add?

lvinniel:
Plus 1!

moreau-nicolas:
I think this feature would be nice to have too!

Regarding the implementation I'd rather let the annotations be @Repeatable.

Your solution might be friendlier to non-Java 8 users, though.

mikeholler:
@moreau-nicolas it would be possible to make this annotation repeatable on JVM versions that support it. You can have it so that people using Java 8 can write:

```
@When("^I perform the calculation (\\d+) plus (\\d+)$")
@When("^I add the numbers (\\d+) and (\\d+)$")
void performXPlusY(int x, int y) {
    // do the step
}
```

And people using Java < 8 could write:

```
@When(
        value = {
                "^I perform the calculation (\\d+) plus (\\d+)$",
                "^I add the numbers (\\d+) and (\\d+)$"
        }
)
void performXPlusY(int x, int y) {
    // do the step
}
```

Notice that both examples work in Java 8, but only the latter works in Java < 8.

dkowis:
I don't believe this is a feature in any of the other Cucumber implementations. You are able to delegate to a method call, having two annotated methods, 3 methods total. 

Help me understand how will this feature improve the quality of your Scenarios? How will this feature improve the communication between stakeholders?

mikeholler:
Cucumber relies on the natural language capabilities of Gherkin to allow tests to be written that are supposedly easy for both managers and programmers to understand. However, it's not always one person that writes specs (Gherkin .feature files) and even if there is one person it's very easy to write two steps that mean the same thing in syntactically different ways.

[...]

Issue #937, opened 2015-11-30
Enhancement: make Java8 lambda steps reflection portable and use stable javac APIs


RichardBradley:
See:
- #912: a JDK upgrade which changed class file layout broke Java8 lambda steps
- #916: closing over local variables breaks Java8 lambda steps
  - (currently fixed on branch https://github.com/cucumber/cucumber-jvm/tree/java8-fixes , using the existing Constant Pool inspector)
- #936: Java 8 Lambda steps don't support Map or List arguments
  - (may not be fixable using the the existing Constant Pool inspector)

Currently, for Java8 lambda steps to work, the step runner needs to know the declared type of all arguments to the step body (see `Java8StepDefinition.getParameterInfos`).

The declared type can be any runtime type, including primitives like `int`, normal classes like `String` or generic types like `List<String>`. Users can register converters to allow any type for these step params (see [`cucumber.runtime.ParameterInfo.convert`](https://github.com/cucumber/cucumber-jvm/issues/912#issuecomment-160616578)), including perhaps `Object` (although that would be a pathological case, I expect it might need special handling by the reflector and it does seem to be allowed).

It does look to me like we could replace the class ConstantPool inspector with a compile-time Compiler Annotation Processor which records the declared generic types.
See e.g. http://stackoverflow.com/questions/29922668/get-typeelement-from-generic-typeparameterelement-for-java-annotation-processor

This would mean that cucumber-jvm would be using a public supported javac API, instead of hacking about with trying to interpret the class files internals (which might change again).

We could either require Cucumber java8 users to annotate their Step files to give cucumber-jvm access to edit the source, or we could do a full [project lombok](https://github.com/rzwitserloot/lombok/blob/9fbc226e090b1ca1efef5e10fd1bd13c4a686cfd/src/core/lombok/core/AnnotationProcessor.java) and register a compiler plugin for everyone who compiles anything with cucumber-java8 on the classpath and scan for (and fixup) all calls to En.Given etc.

I'll see if I can get this to work and submit a POC.

This will need careful testing, and possibly deserve a major release version, as it will greatly change how `java8` steps are compiled. 
We'll need to test different compilers and different build tools (Maven, ant, manual javac) and check that they all execute the CAP correctly.
First though, I'll write a proof of concept to see if this can be made to work at all, and what the code would look like with this new approach.

RichardBradley:
OK, here's what I have been able to find.
[...]

PR #1178, opened 2017-07-09
[Java8] Replace ConstantPoolTypeIntrospector

mpkorstanje:
## Summary

Replace ConstantPoolTypeIntrospector with [TypeResolver](https://github.com/jhalterman/typetools) and refactor all the Java8 related stuff out of cucumber-java.

## Motivation and Context

To determine which argument types a lambda function requires we created a [ConstantPoolTypeIntrospector](https://github.com/cucumber/cucumber-jvm/blob/aa4c9a012d623cbbedeaa0e54af7c30f32825928/java8/src/main/java/cucumber/runtime/java8/ConstantPoolTypeIntrospector.java). However it has never functioned quite correctly (#937, #1140, [#957](https://github.com/cucumber/cucumber-jvm/issues/957#issuecomment-315467655)), was prone to breaking (#912, #914)  and hasn't been tested much (#1048).

It is important to understand that while we will get a properly functioning and tested replacement, TypeResolver uses the same ConstantPool and thus has the same potential to break. However because TypeResolver is used by a much larger audience I expect these problems to be shallow.

Because this changes the interface of Java8StepDefinition it made sense to refactor all the Java8 related stuff out of cucumber-java. This will make it easier in the future to add things like KotlinStepDefintions without creating a separate KotlinBackend.

## How Has This Been Tested?

Check if the automated tests were still passing. 

## Types of changes
 
     -  Method references that use primitives are correctly understood.
     -  When using anonymous innerclasses with Java8 Stepdefinitions it was possible to use typed lists and maps as an argument. This is no longer possible as TypeResolver only returns classes, not the full type.
    - In fact it will not be possible to use anonymous inner classes as step definitions without java8 at all.

## Checklist:


java/pom.xml
```diff
@@ -17,16 +17,6 @@
...
         </dependency>
-        <dependency>
-            <groupId>info.cukes</groupId>
-            <artifactId>cucumber-jvm-deps</artifactId>
-            <scope>provided</scope>
-        </dependency>
-        <dependency>
-            <groupId>io.cucumber</groupId>
-            <artifactId>gherkin</artifactId>
-            <scope>provided</scope>
```
mpkorstanje: Actually both are transitive dependency through cucumber-core.

java/src/main/java/cucumber/runtime/java/JavaBackend.java
```diff
@@ -59,24 +60,25 @@ private Snippet createSnippet() {
...
+                    currentThread().getContextClassLoader()),
+                currentThread().getContextClassLoader()));
     }
 
     public JavaBackend(ObjectFactory objectFactory, ClassFinder classFinder) {
-        this.objectFactory = objectFactory;
         this.classFinder = classFinder;
-        methodScanner = new MethodScanner(classFinder);
+        this.objectFactory = objectFactory;
+        this.methodScanner = new MethodScanner(classFinder);
```
mpkorstanje: Rewrote the constructors to be idiomatic java.

java/src/main/java/cucumber/runtime/java/JavaBackend.java
```diff
@@ -157,44 +159,30 @@ void addStepDefinition(Annotation annotation, Method method) {
         }
     }
 
-    public void addStepDefinition(String regexp, long timeoutMillis, StepdefBody body, TypeIntrospector typeIntrospector) {
-        try {
-            glue.addStepDefinition(new Java8StepDefinition(Pattern.compile(regexp), timeoutMillis, body, typeIntrospector));
```
mpkorstanje: This has moved into the generated code.

java/src/main/java/cucumber/runtime/java/JavaBackend.java
```diff
@@ -157,44 +159,30 @@ void addStepDefinition(Annotation annotation, Method method) {
...
                 String[] tagExpressions = ((After) annotation).value();
                 long timeout = ((After) annotation).timeout();
-                glue.addAfterHook(new JavaHookDefinition(method, tagExpressions, ((After) annotation).order(), timeout, objectFactory));
+                addAfterHookDefinition(new JavaHookDefinition(method, tagExpressions, ((After) annotation).order(), timeout, objectFactory));
             }
         }
     }
 
-    public void addBeforeHookDefinition(String[] tagExpressions, long timeoutMillis, int order, HookBody body) {
-        glue.addBeforeHook(new Java8HookDefinition(tagExpressions, order, timeoutMillis, body));
```
mpkorstanje: This has moved into the generated code.

java/src/main/java/cucumber/runtime/java/MethodScanner.java
```diff
@@ -18,7 +18,7 @@
 
     private final ClassFinder classFinder;
 
-    public MethodScanner(ClassFinder classFinder) {
+    MethodScanner(ClassFinder classFinder) {
```
mpkorstanje: Unrelated cleanup.

java/src/test/java/cucumber/runtime/java/java8test/AnonInnerClassStepdefs.java
```diff
@@ -1,20 +0,0 @@
-package cucumber.runtime.java.java8test;
-
-import cucumber.api.java8.GlueBase;
-import cucumber.api.java8.StepdefBody;
-import cucumber.runtime.java.JavaBackend;
-
-import static org.junit.Assert.assertEquals;
-
-public class AnonInnerClassStepdefs implements GlueBase {
```
mpkorstanje: moved to cucumber.runtime.java8.test.AnonInnerClassStepdefs

java8/src/main/code_generator/I18n.java8.txt
```diff
@@ -1,28 +1,37 @@
...
-        JavaBackend.INSTANCE.get().addStepDefinition(regexp, 0, body, ConstantPoolTypeIntrospector.INSTANCE);
+        JavaBackend.INSTANCE.get().addStepDefinition(
+            new Java8StepDefinition(regexp, 0, StepdefBody.A0.class, body)
+        );
     }
 
     default void ${java.text.Normalizer.normalize(kw.replaceAll("[\\s',!]", ""), java.text.Normalizer.Form.NFC)}(final String regexp, final long timeoutMillis, final StepdefBody.A0 body) {
-        JavaBackend.INSTANCE.get().addStepDefinition(regexp, timeoutMillis, body, ConstantPoolTypeIntrospector.INSTANCE);
+        JavaBackend.INSTANCE.get().addStepDefinition(
+            new Java8StepDefinition(regexp, timeoutMillis, StepdefBody.A0.class, body)
```
mpkorstanje: Typetools resolves the arguments of the the generic type (StepdefBody.A1) using the type variable information from body (the implementing class) so we pass these in here.

java8/src/test/java/cucumber/runtime/java8/test/LambdaStepdefs.java
```diff
@@ -65,19 +66,24 @@ public LambdaStepdefs() {
...
     }
 
     private void methodThatDeclaresException() throws Throwable {
     }
     private void methodWithAnArgument(Integer cuckes) throws Throwable {
         assertEquals(42, cuckes.intValue());
     }
+    private void methodWithAnIntArgument(int cuckes) throws Throwable {
+        assertEquals(42, cuckes);
+    }
```
mpkorstanje: Fixed by using TypeResolver

java/src/main/java/cucumber/runtime/java/JavaBackend.java
```diff
@@ -59,24 +60,25 @@ private Snippet createSnippet() {
...
 
     public JavaBackend(ObjectFactory objectFactory) {
-        ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
-        ResourceLoader resourceLoader = new MultiLoader(classLoader);
-        classFinder = new ResourceLoaderClassFinder(resourceLoader, classLoader);
-        methodScanner = new MethodScanner(classFinder);
-        this.objectFactory = objectFactory;
+        this(objectFactory,
+            new ResourceLoaderClassFinder(
+                new MultiLoader(
```
brasmusson: I'm not sure all these line breaks improve readability, especially 
`new MultiLoader(currentThread().getContextClassLoader())` seems to fit better on one line.

java8/src/main/java/cucumber/runtime/java8/Java8HookDefinition.java
```diff
@@ -1,17 +1,17 @@
-package cucumber.runtime.java;
+package cucumber.runtime.java8;
+
+import static java.util.Arrays.asList;
```
brasmusson: I've found that the most common ordering of import in the code base is static imports after the other imports, so I've added that ordering to the Eclipse configuration I use for Cucumber-JVM, but you seems to have another configuration.

mpkorstanje:
I am interested in your opinion about separating out the java8 bits from cucumber-java. Moving it out of the cucumber-java, out of the JavaBackend, means that any information the Java8StepDefintion uses will have to be either locally available in the generated steps globally accessible.

With the upcoming CucumberExpressions this might be a problem. CucumberExpressions need the the TypeRegistry to be instantiated.

java8/src/main/java/cucumber/runtime/java8/Java8HookDefinition.java
mpkorstanje: I've been using the default that comes with IDEA. What does Eclipse do by default? I'd prefer a minimal amount of configuration if possible.

mpkorstanje: Looks like I have the ability to quickly format the imports of the entire project. We can pick a standard now and adhere to it.

https://www.jetbrains.com/help/idea/optimizing-imports.html

java/src/main/java/cucumber/runtime/java/JavaBackend.java
mpkorstanje: You are right. I was missing a constructor `JavaBackend(ObjectFactory, ResourceLoader)`. This made look into the root cause, why do we have these constructors in the first place. It turns out they're only used by tests which needed them because they were poorly written. 

I've added a commit which improves these test somewhat.

mpkorstanje:
I have a solution for the CucumberExpressions in mind. Rather then adding the Java8StepDefintion directly to the backend:

```java
JavaBackend.INSTANCE.get().addStepDefinition(
     new Java8StepDefinition(regexp, timeoutMillis, StepdefBody.A0.class, body)
);
```
We'd provide a Function<TypeRegistry,StepDefinition> (or the java7 equivalent) that will instantiate one:

```java
JavaBackend.INSTANCE.get().addStepDefinition((typeRegistry) -> 
     new Java8StepDefinition(typeRegistry, regexp, timeoutMillis, StepdefBody.A0.class, body)
);
```

PR #1182, opened 2017-07-16
Run Android Studio Examples on Travis

mpkorstanje:
Run Android Studio Examples on Travis

Todo:
 *  [ ] Use snapshot snapshot builds in Android studio

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,18 @@
 language: java
 sudo: false # Use container based builds. They are faster and we do not need sudo.
 
+before_cache:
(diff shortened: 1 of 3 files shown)


brasmusson:
I think that adding 
```
    maven { url 'https://oss.sonatype.org/content/repositories/snapshots' }
``` 
to the list of repositories in the [`build.gradle`](https://github.com/cucumber/cucumber-jvm/blob/master/examples/android/android-studio/Cukeulator/build.gradle) file would enable the use of snapshots in the gradle build.

mpkorstanje:
Obsoleted by  #1410

PR #1187, opened 2017-07-24
[Core] Use cucumber-ruby rerun file specification

mpkorstanje:
Implemented the rerun file specification from cucumber ruby[1]. It
handles spaces, new lines as a separator as well as no separator at all.

References:
 [1] https://github.com/cucumber/cucumber-ruby/blame/master/lib/cucumber/cli/rerun_file.rb

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/runtime/model/CucumberFeature.java
+++ b/core/src/main/java/cucumber/runtime/model/CucumberFeature.java
@@ -14,10 +14,11 @@ import java.io.IOException;
 import java.io.PrintStream;
 import java.io.Serializable;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public class CucumberFeature implements Serializable {
     private static final long serialVersionUID = 1L;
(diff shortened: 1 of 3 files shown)


core/src/main/java/cucumber/runtime/model/CucumberFeature.java
```diff
@@ -14,17 +14,19 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public class CucumberFeature implements Serializable {
     private static final long serialVersionUID = 1L;
     private final String uri;
     private String language;
     private GherkinDocument gherkinDocument;
     private String gherkinSource;
+    public static final Pattern RERUN_PATH_SPECIFICATION = Pattern.compile("(?m:^| |)(.*?\\.feature(?:(?::\\d+)*))");
```
mpkorstanje: `?m: ` for multi line or this input:

```
/some/path/foo.feature
[space]/some/path/foo.feature
```
Will yield: `["/some/path/foo.feature", "/some/path/bar.feature"]`. 

Not sure if this actually a bad thing though.

core/src/main/java/cucumber/runtime/model/CucumberFeature.java
```diff
@@ -71,7 +66,10 @@ private static void loadFromRerunFile(FeatureBuilder builder, ResourceLoader res
         for (Resource resource : resources) {
             String source = read(resource);
             if (!source.isEmpty()) {
-                featurePaths.addAll(Arrays.asList(source.split(" ")));
```
mpkorstanje: Not sure how I missed this the first time round.

brasmusson: With Gherkin v2 the filters were passed to the parser and the filtering was done there, so it was straightforward to add the line filters from the rerun file to the filters passed through the `load` method. Now with Gherkin v4 the filtering is instead done after the compiling of the features to pickles, so we are actually reading the rerun file twice, once to find the feature files to be parsed, and once to get the line filters to the pickle filter chain. This duplication was probably caused by that those two used was fixed at different times in the work or upgrading to Gherkin v4 (and the extract method refactoring missed).

PR #1190, opened 2017-08-01
[JUnit] Invoke (Before|After)Class and TestRules around Cucumber execution

mpkorstanje:
Junit invokes several framework methods around the test execution.
The cucumber runner should finish its whole execution inside these.

To be precises these two should be equivalent:

```
@BeforeClass
public static void before {
   // Do stuff
}

@Test
public void test(){
    cucumber.api.cli.Main.main(some arguments....)
}

@AfterClass
public static void after {
   // Do stuff
}

```

```
@RunWith(Cucumber.class)
public class Test {
    @BeforeClass
    public static void before {
       // Do stuff
    }

    @AfterClass
    public static void after {
     // Do stuff
    }
}
```

By firing the TestRunFinished event and printing the summary directly
after invoking the children rather then after the completion of the
whole test run these are made equivalent again.

Firing the TestRunFinished event before the JUnit execution 
is done might look a bit odd but semantically this is a cucumber event
not a junit event and cucumber has finished its execution at this point.

## Context

https://cucumberbdd.slack.com/archives/C590XDQQH/p1501620355467970

## Types of changes

## Checklist:

- [ ] Awaiting confirmation from toothpicks131

I've manually placed breakpoints in the examples above to verify the methods are called in the correct order.

### DIFF
--- a/junit/src/main/java/cucumber/api/junit/Cucumber.java
+++ b/junit/src/main/java/cucumber/api/junit/Cucumber.java
@@ -19,6 +19,7 @@ import org.junit.runner.Description;
 import org.junit.runner.notification.RunNotifier;
 import org.junit.runners.ParentRunner;
 import org.junit.runners.model.InitializationError;
+import org.junit.runners.model.Statement;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -75,7 +76,7 @@ public class Cucumber extends ParentRunner<FeatureRunner> {
      * @param runtimeOptions configuration
      * @return a new runtime
      * @throws InitializationError if a JUnit error occurred
-     * @throws IOException if a class or resource could not be loaded
+     * @throws IOException         if a class or resource could not be loaded
      */
     protected Runtime createRuntime(ResourceLoader resourceLoader, ClassLoader classLoader,
                                     RuntimeOptions runtimeOptions) throws InitializationError, IOException {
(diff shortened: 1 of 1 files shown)


brasmusson:
I can see how this makes sense. There are however some interesting differences in the fine details of execution depending on which runner that is used, for instance:
* The JUnit and TestNG runners will establish the test count to be executed before any scenario is executed (because it needs to be communicated to the JUnit/TestNG infrastructure), but the CLI runner will compile and execute each feature file in sequence (so the test count from the later feature files are not known when the scenarios from the first feature file is executed).
* The TestNG runner will finish the Cucumber execution (send the TestRunFinished event and print the summary) in a [`@AfterClass` method](https://github.com/cucumber/cucumber-jvm/blob/master/testng/src/main/java/cucumber/api/testng/AbstractTestNGCucumberTests.java#L36), whereas the JUnit runner with this PR will have done that before any `@AfterClass` method has been executed.

From the Slack thread triggering this PR:
> The challenge, though, is that the cucumber reports are not available when junit's AfterClass hook executes.

In this Slack thread the proposed `@beforeall`/`@afterall` are also discussed. However since `@beforeall`/`@afterall` (if implemented) will be part of the Cucumber execution, they would not be equivalent to JUnit `@BeforeClass`/`@AfterClass` methods. Specifically the Started/Finished events for these hooks should be available to the formatters and event listeners ([which I have argued for Cucumber-Ruby](https://github.com/cucumber/cucumber-ruby/pull/1011#issuecomment-244625343), so the cucumber report cannot be available when a `@afterall` hook is executed.

Issue #619, opened 2013-11-01
[JVM] Pretty Formatter error


TarekSaid:
Hi,

I am getting a StringIndexOutOfBoundsException when using the pretty formatter. When I remove the formatter, all my scenarios pass.

UPDATE:
it seems this problem has to do with the regular expressions I use for my feature steps.

I have the following feature file, in portuguese (UPDATED):

```
# language: pt
# encoding: iso-8859-1
@copia-ncm
Funcionalidade: copiar o NCM do produto para cada empresa do grupo
  Como uma empresa cadastrando um novo produto 
  Para gerar corretamente as tributações para as minhas outras empresas do grupo
  Eu quero que o NCM do produto seja copiado

  Contexto: Dado que eu estou logado

  @copia-ncm-existente
  Cenário: cópia de NCM existente
    Dado que eu estou cadastrando o produto "12341" com NCM iniciado por "9999.99.99"
    E a empresa 1 possui esse NCM
    Quando eu salvo o produto
    Então eu deveria ver o ncm no produto dessa empresa com código "9999.99.99-01"

  @copia-ncm-enquad-igual
  Cenário: cópia de NCM inexistente para mesmo enquadramento tributário
    Dado que eu estou cadastrando o produto "12343" com NCM iniciado por "8888.88.88-60"
    E a empresa 16 não possui esse NCM e tem mesmo enquadramento tributário
    Quando eu salvo o produto
    Então eu deveria ver o ncm no produto dessa empresa com código "8888.88.88-01" e a mesma situação tributária

  @copia-ncm-enquad-diferente
  Cenário: cópia de NCM inexistente para enquadramento tributário diferente
    Dado que eu estou cadastrando o produto "12342" com NCM iniciado por "1234.56.78"
    E a empresa 1 não possui esse NCM e não tem mesmo enquadramento tributário
    Quando eu salvo o produto
    Então eu deveria ver o ncm no produto dessa empresa com código "1234.56.78-01" e a mesma situação tributária

  @copia-ncm-importacao-existente
  Cenário: cópia de NCM existente durante a importação de dados
    Dado que eu estou importando o produto "999994" com NCM iniciado por "9999.99.99"
    E que eu tenho o arquivo "test/resources/0200-ncm-existente.txt"
    E que eu quero importar "Produtos"
    E a empresa 1 possui esse NCM
    Quando eu importo o arquivo
    Então eu deveria ver o ncm no produto dessa empresa com código "9999.99.99-01"

[...]

PR #1197, opened 2017-08-12
Fix PrettyFormatter Exception for nested capture groups

mlvandijk:
## Summary

Do not throw StringIndexoutOfBoundsException in PrettyFormatter for nested capture groups.
Format nested arguments as part of already formatted enclosing argument.

## Details

Reproduced issue #619 with (now removed) test that expects a StringIndexOutOfBoundsException.
Added 2 tests to verify formatting of nested capture groups and added some checks on variables passed to substring() to prevent StringIndexOutOfBoundsException.
Additionally, minor refactor (renamed these variables) for clarity.
(Additionally, fixed a typo in PrettyFormatterTest)

## Motivation and Context
Nested capture groups could not be handled by the PrettyFormatter. Since each capture groups maps to an argument, this sort of gives an argument in an argument. A nested argument starts before the enclosing argument ends, throwing a StringIndexOutOfBoundsException on stepText.substring() - thus preventing additional features from running.

Detailed explanation:
public String substring(int beginIndex, int endIndex)
throws: IndexOutOfBoundsException - if the beginIndex is negative, or endIndex is larger than the length of this String object, or beginIndex is larger than endIndex.

The expected value of argument.offset is negative for nested capture groups, since a nested argument starts before the previous (=enclosing) argument ends. Therefore:
ad 1. beginIndex is negative => when using a nested argument, as the end of the nested argument (argument offset + length of argument) is before the end of the previous (=enclosing) argument -> Added check to see if beginIndex >= 0
ad 2. endIndex is larger than the length of this String object => should not happen, as endIndex = argument offset + length of argument, which should always be inside the String stepText.
ad 3. beginIndex is larger than endIndex => for a nested argument, the nested argument starts before the previous (=enclosing) argument ends, and the endIndex (=value of argument.offset) is negative -> Added check check if argumentOffset < beginIndex
Fixed to ignore formatting for nested capture groups (arguments), as they are already formatted as an argument as part of the enclosing capture group (argument), thus no longer throwing StringIndexOutOfBoundsException.

## How Has This Been Tested?
1. Reproduce original bug (expect Exception)
2. Test current function with nested argument(s) -> nested argument is formatted as part of the enclosing argument (and no Exception is thrown
3. Clean/install cucumber-jvm core (i.e. checking all other tests still pass)

## Types of changes

## Checklist:

May need update to documentation, which I have not yet made. I suggest using explanation of pretty formatter by @mpkorstanje

### DIFF
--- a/core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
+++ b/core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
@@ -214,10 +214,13 @@ class PrettyFormatter implements Formatter, ColorAware {
             // can be null if the argument is missing.
             if (argument.getOffset() != null) {
                 int textEnd = argument.getOffset();
-                if (textStart <= textEnd ) { // textStart can be > textEnd with nested capture groups
+                if (textStart > textEnd ) { // textStart can be > textEnd with nested capture groups
+                    continue;
+                } else {
                     String text = stepText.substring(textStart, textEnd);
                     result.append(textFormat.text(text));
                 }
+
             }
             // val can be null if the argument isn't there, for example @And("(it )?has something")
             if (argument.getVal() != null) {
@@ -225,7 +228,7 @@ class PrettyFormatter implements Formatter, ColorAware {
                 textStart = argument.getOffset() + argument.getVal().length();
             }
         }
-        if (textStart != stepText.length() && textStart >= 0) { // textStart can be < 0 with nested capture groups
+        if (textStart != stepText.length() && textStart >= 0 && textStart < stepText.length()) { // textStart can be < 0 with nested capture groups
             String text = stepText.substring(textStart, stepText.length());
             result.append(textFormat.text(text));
         }
--- a/core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
+++ b/core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
@@ -213,8 +213,11 @@ class PrettyFormatter implements Formatter, ColorAware {
         for (Argument argument : arguments) {
             // can be null if the argument is missing.
             if (argument.getOffset() != null) {
-                String text = stepText.substring(textStart, argument.getOffset());
-                result.append(textFormat.text(text));
+                int textEnd = argument.getOffset();
+                if (textStart <= textEnd ) { // textStart can be > textEnd with nested capture groups
+                    String text = stepText.substring(textStart, textEnd);
+                    result.append(textFormat.text(text));
+                }
             }
             // val can be null if the argument isn't there, for example @And("(it )?has something")
             if (argument.getVal() != null) {
@@ -222,7 +225,7 @@ class PrettyFormatter implements Formatter, ColorAware {
                 textStart = argument.getOffset() + argument.getVal().length();
             }
         }
-        if (textStart != stepText.length()) {
+        if (textStart != stepText.length() && textStart >= 0) { // textStart can be < 0 with nested capture groups
             String text = stepText.substring(textStart, stepText.length());
             result.append(textFormat.text(text));
         }
--- a/core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
+++ b/core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
@@ -208,28 +208,27 @@ class PrettyFormatter implements Formatter, ColorAware {
     }
 
     String formatStepText(String keyword, String stepText, Format textFormat, Format argFormat, List<Argument> arguments) {
-        int textStart = 0;
+        int beginIndex = 0;
         StringBuilder result = new StringBuilder(textFormat.text(keyword));
         for (Argument argument : arguments) {
             // can be null if the argument is missing.
             if (argument.getOffset() != null) {
-                int textEnd = argument.getOffset();
-                if (textStart > textEnd ) { // textStart can be > textEnd with nested capture groups
+                int argumentOffset = argument.getOffset();
+                if (argumentOffset < beginIndex ) { // a nested argument starts before the enclosing argument ends; ignore it when formatting
                     continue;
-                } else {
-                    String text = stepText.substring(textStart, textEnd);
-                    result.append(textFormat.text(text));
                 }
-
+                String text = stepText.substring(beginIndex, argumentOffset);
+                result.append(textFormat.text(text));
             }
             // val can be null if the argument isn't there, for example @And("(it )?has something")
             if (argument.getVal() != null) {
                 result.append(argFormat.text(argument.getVal()));
-                textStart = argument.getOffset() + argument.getVal().length();
+                int argumentEnd = argument.getOffset() + argument.getVal().length(); // end of current argument; can be < 0 for nested argument
+                beginIndex = argumentEnd;
             }
         }
-        if (textStart != stepText.length() && textStart >= 0 && textStart < stepText.length()) { // textStart can be < 0 with nested capture groups
-            String text = stepText.substring(textStart, stepText.length());
+        if (beginIndex != stepText.length() && beginIndex >= 0) {
+            String text = stepText.substring(beginIndex, stepText.length());
             result.append(textFormat.text(text));
         }
         return result.toString();
--- a/core/src/test/java/cucumber/runtime/formatter/PrettyFormatterTest.java
+++ b/core/src/test/java/cucumber/runtime/formatter/PrettyFormatterTest.java
@@ -21,6 +21,7 @@ import static cucumber.runtime.TestHelper.result;
 import static java.util.Arrays.asList;
 import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.equalTo;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThat;
 
 public class PrettyFormatterTest {
@@ -50,7 +51,7 @@ public class PrettyFormatterTest {
     }
 
     @Test
-    public void should_handle_backgound() throws Throwable {
+    public void should_handle_background() throws Throwable {
         CucumberFeature feature = feature("path/test.feature", "" +
                 "Feature: feature name\n" +
                 "  Background: background name\n" +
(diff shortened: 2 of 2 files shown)


core/src/test/java/cucumber/runtime/formatter/PrettyFormatterTest.java
```diff
@@ -381,6 +382,37 @@ public void should_mark_arguments_in_steps() throws Throwable {
...
+    public void nested_capture_groups_cannot_be_formatted(){
+        Formats formats = new AnsiFormats();
+        Argument fullArg = new Argument(0, "this is the nested argument full argument");
+        Argument nestedArg = new Argument(-29, "nested argument");
+        PrettyFormatter prettyFormatter = new PrettyFormatter(null);
+
+        try {
+            String formattedText = prettyFormatter.formatStepText("Given ", "this is the (nested argument) full argument", formats.get("passed"), formats.get("passed_arg"), asList(fullArg, nestedArg));
+        } catch (StringIndexOutOfBoundsException e) {
+            assertEquals("String index out of range: -70", e.getMessage());
```
mpkorstanje: Right now this test is asserting that an exception which it doesn't expect to be thrown has the right format. I would expect a test that checks if the formatted text is properly formatted. For example see: `should_mark_arguments_in_steps`

core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
```diff
@@ -213,16 +213,19 @@ String formatStepText(String keyword, String stepText, Format textFormat, Format
...
+                int textEnd = argument.getOffset();
+                if (textStart <= textEnd ) { // textStart can be > textEnd with nested capture groups
+                    String text = stepText.substring(textStart, textEnd);
+                    result.append(textFormat.text(text));
+                }
             }
             // val can be null if the argument isn't there, for example @And("(it )?has something")
             if (argument.getVal() != null) {
                 result.append(argFormat.text(argument.getVal()));
                 textStart = argument.getOffset() + argument.getVal().length();
```
mpkorstanje: I think you're missing a bounds check here.

Though I regret suggesting bounds checking in the first place as the solution. Having looked at the problem a bit more closely I think it would be more elegant to skip nested arguments all together. You can skip processing an element in an array using the `continue` statement. 

https://docs.oracle.com/javase/tutorial/java/nutsandbolts/branch.html
https://stackoverflow.com/questions/389741/what-is-the-continue-keyword-and-how-does-it-work-in-java

core/src/test/java/cucumber/runtime/formatter/PrettyFormatterTest.java
```diff
@@ -381,6 +382,37 @@ public void should_mark_arguments_in_steps() throws Throwable {
                                           AnsiEscapes.GREEN + AnsiEscapes.INTENSITY_BOLD + "arg2"  + AnsiEscapes.RESET));
     }
 
+    // Reproduce issue #619: This test should now fail!
+    @Test(expected = StringIndexOutOfBoundsException.class)
+    public void nested_capture_groups_cannot_be_handled(){
```
mlvandijk: Will remove this test. I only kept it as the "guidelines for contributing" mentioned something about having a test to reproduce a bug...

mpkorstanje: Ah. That means you have to write a test that would fail without your fix.

For example in your test you assert that it produces a certain output -the one you expected it to produce- but instead you get an exception. Then you fix the bug so the test passes.

mlvandijk: Ofcourse that's what it means...

core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
mlvandijk: Yes, fixed that too. And thanks for specifying intended behaviour! New PR on the way!

mlvandijk:
@plaindocs I think above explanation of PrettyFormatter should be in the docs (if it isn't already)?

core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
```diff
@@ -213,16 +213,22 @@ String formatStepText(String keyword, String stepText, Format textFormat, Format
         for (Argument argument : arguments) {
             // can be null if the argument is missing.
             if (argument.getOffset() != null) {
-                String text = stepText.substring(textStart, argument.getOffset());
-                result.append(textFormat.text(text));
+                int textEnd = argument.getOffset();
+                if (textStart > textEnd ) { // textStart can be > textEnd with nested capture groups
```
mpkorstanje: What is the expected range of offset values for nested capture groups?

core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
```diff
@@ -213,16 +213,22 @@ String formatStepText(String keyword, String stepText, Format textFormat, Format
...
+                    continue;
+                } else {
```
mpkorstanje: After a branching statement (`break`, `continue` and `return`) you can omit the else branch. It is implied.

core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
```diff
@@ -213,16 +213,22 @@ String formatStepText(String keyword, String stepText, Format textFormat, Format
...
+
             }
             // val can be null if the argument isn't there, for example @And("(it )?has something")
             if (argument.getVal() != null) {
                 result.append(argFormat.text(argument.getVal()));
                 textStart = argument.getOffset() + argument.getVal().length();
             }
         }
-        if (textStart != stepText.length()) {
+        if (textStart != stepText.length() && textStart >= 0 && textStart < stepText.length()) { // textStart can be < 0 with nested capture groups
```
mpkorstanje: Can you explain (or even better write  a test) for the scenario where `textStart` becomes less then zero or longer then the stepText now that we are skipping nested arguments?

core/src/test/java/cucumber/runtime/formatter/PrettyFormatterTest.java
```diff
@@ -381,6 +382,20 @@ public void should_mark_arguments_in_steps() throws Throwable {
                                           AnsiEscapes.GREEN + AnsiEscapes.INTENSITY_BOLD + "arg2"  + AnsiEscapes.RESET));
     }
 
+    @Test
+    public void should_mark_nested_argument_as_part_of_full_argument(){
```
mpkorstanje: Looks good!

core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
mpkorstanje: Or alternatively, if you can't find such a scenario, can you proof that it can't happen?

core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
mlvandijk: Ofcourse. Removed the else.

core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
mlvandijk: Working on it...

core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
mlvandijk: @mpkorstanje Quick question to make sure I really understand: In your example, the second (nested) argument starts 17 positions before the end of the first (enclosing) argument. The offset would therefore be -17, right?

mpkorstanje:
That would be correct. In general I believe that the position of the first
nested argument is before its parent (a negative offset value).

But I am not sure how it works for a second nested argument. The offset
could be relative to the parent (also negative) or to the previous in which
case it could be positive. This is something that would have to be sorted
out.

I think it be best to proceed with the assumption that there is only one
nested argument. Then once complete figure out how multiple nested
arguments work.

On Aug 15, 2017 22:04, "Marit van Dijk" <notifications@github.com> wrote:

*@mlvandijk* commented on this pull request.
------------------------------

In core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
<https://github.com/cucumber/cucumber-jvm/pull/1197#discussion_r133288720>:

> @@ -213,16 +213,19 @@ String formatStepText(String keyword, String stepText, Format textFormat, Format
         for (Argument argument : arguments) {
             // can be null if the argument is missing.
             if (argument.getOffset() != null) {
-                String text = stepText.substring(textStart,
argument.getOffset());
-                result.append(textFormat.text(text));
+                int textEnd = argument.getOffset();
+                if (textStart <= textEnd ) { // textStart can be >
textEnd with nested capture groups
+                    String text = stepText.substring(textStart, textEnd);
+                    result.append(textFormat.text(text));
+                }
             }
             // val can be null if the argument isn't there, for
example @And("(it )?has something")
             if (argument.getVal() != null) {
                 result.append(argFormat.text(argument.getVal()));
                 textStart = argument.getOffset() + argument.getVal().length();

@mpkorstanje <https://github.com/mpkorstanje> Quick question to make sure I
really understand: In your example, the second (nested) argument starts 17
positions before the end of the first (enclosing) argument. The offset
would therefore be -17, right?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/cucumber/cucumber-jvm/pull/1197#discussion_r133288720>,
or mute the thread

aslakhellesoy:
I have only skimmed the discussion, and I haven't looked at the code. I just wanted to mention that Cucumber Expressions, which we plan to integrate in Cucumber-JVM soon (it's in Cucumber.js and Cucumber-Ruby already), has a nice (IMHO) API for accessing nested capture groups).

The [TreeRegexp#match](https://github.com/cucumber/cucumber/blob/c413e7ccb608070292b949cf399fb4e118748bc6/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/TreeRegexp.java#L57) method returns a `Group` which is a tree structure representing the nested capture groups in the regexp.

Do you think it will be easier to fix this issue when that's in place? Any changes we'd need to make to Cucumber Expressions?

mlvandijk:
@mpkorstanje Thanks! It looks like the original implementation assumes multiple arguments in a row only; no nested arguments. Hence the exception on  stepText.substring(textStart, textEnd); when textEnd is before textStart. I'd also say the naming here is confusing because textStart / textEnd are (IMHO) correct for the substring (from, to) but are in fact the textEnd is the argument.getOffset() so start of the argument and textStart is argument.getOffset() + argument.getVal().length() so the end of the argument. I can do some testing around this to see exactly when the exception is thrown and make sure all situations are covered.
Also, I would assume any nested argument inside nested argument would be handled since we're just ignoring the nested arguments with continue, but I'll check that also.
@aslak looks like the cucumber expressions do explicitly take nested arguments into account, so that might solve the problem altogether? Or would the PrettyFormatter still be used (in which case I'll finish this issue to keep from throwing the StringIndexOutOfBoundsException)

mlvandijk:
Please see new PR #1200

Issue #318, opened 2012-05-16
Add 'throws Exception' to generated Java stepdef snippets


pettermahlen:
The convention (or my convention at least - I think it is general) when writing unit tests is to have method signatures like this:

``` java
@Test
public void shouldDoSomethingSomeWay() throws Exception {
   // test code
}
```

This means you don't need to worry about catching any checked exceptions, and is a statement that 'any exceptions thrown out of this test means the test should fail'.

I think it is a good idea to do the same with step definitions. Would it make sense to add 'throws Exception' to the method signatures in Java snippets? I'll be happy to submit some code if so.

PR #1208, opened 2017-08-28
Reduce Throwable to Exception in JavaSnippet

### LINKED ISSUE #1207: Checkstyle warnings on throwing Throwables in Step Definitions
## Summary

If a given Cucumber Java project uses [CheckStyle](http://checkstyle.sourceforge.net/), one of the warnings raised is the `IllegalThrowsCheck` for Step Definitions, because, by default, they throw a Throwable.

## Expected Behaviour

We could try to improve the default throws of step definitions to ensure that types (Like Error or Throwable) are not declared to be thrown and are not allowed.

## Current Behavior
The current behaviour is that by default, we allow throwing Throwables.

## Context & Motivation
Like mentioned above, if a given Cucumber Java project uses [CheckStyle](http://checkstyle.sourceforge.net/) and depends on its validations, a common failure is the throwing of Throwables.

mpkorstanje:
## Summary

Tests are allowed to fail by throwing an(y) exception. As such #318
added `throws Throwable` to the JavaSnippet. Aside from being a bad
practice this results in unnecessary checkstyle warnings. As such the
generated snippet should not explicitly declare `Throwable`.

Reducing `Throwable` to `Exception` should be sufficient.

Fixes #1207
## How Has This Been Tested?

Updated the Unit Tests.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CO
 
 ## [2.0.0-SNAPSHOT](https://github.com/cucumber/cucumber-jvm/compare/v1.2.5...master) (In Git)
 
+* [Java] Reduce Throwable to Exception in JavaSnippet ([#1207](https://github.com/cucumber/cucumber-jvm/issues/1207), [#1208](https://github.com/cucumber/cucumber-jvm/pull/1208) M.P. Korstanje)
 * [Core] Update the cucumber-html dependency to version 0.2.6 (Björn Rasmusson)
 * [Core] Fix PrettyFormatter exception on nested arguments ([#1200](https://github.com/cucumber/cucumber-jvm/pull/1200) Marit van Dijk, M.P. Korstanje) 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@ Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CO
 
 ## [2.0.0-SNAPSHOT](https://github.com/cucumber/cucumber-jvm/compare/v1.2.5...master) (In Git)
 
+* [Java] Reduce Throwable to Exception in JavaSnippet ([#1207](https://github.com/cucumber/cucumber-jvm/issues/1207), [#1208](https://github.com/cucumber/cucumber-jvm/pull/1208) M.P. Korstanje)
```
brasmusson: AFAIK, we discourage updating the Changelog in PR:s, and instead put that task on the person merging the PR.

mpkorstanje: Planned on doing that myself. Then realized @mlvandijk could do it as a practice.

mlvandijk: But I was too slow ;)

brasmusson: Unfortunately, the same day I was up to release Cucumber-JVM 2.0.0 was the wrong day for such practice ...

mpkorstanje: Such is life. :D

PR #1211, opened 2017-08-29
Include javadoc in the build

### LINKED ISSUE #1209: Cannot generate javadoc under java 8
## Summary
Cannot generate javadoc under java 8

## Expected Behavior
Should be able to generate javadoc under java 8

## Current Behavior
Generating javadoc using command 'mvn javadoc:aggregate' under java 8 fails if the javadoc tags are incomplete, apparently due to a new DocLint feature introduced in java8. 

## Possible Solution

This can be 'fixed' by ignoring the linting for java 8 (https://stackoverflow.com/questions/15886209/maven-is-not-working-in-java-8-when-javadoc-tags-are-incomplete/34809831#34809831)

Note: It might be better to "fix"/improve the javadoc by adding the missing tags, but am not sure that doable...

Generating javadoc using command 'mvn javadoc:aggregate' under java 8

## Context & Motivation
Trying to fix issue #796

## Your Environment

* Version used: Current master
* Operating System and version: OS X, Sierra 10.12.6
* SDK 1.8 (1.8.0_121)

mpkorstanje:
Generating javadoc using command 'mvn javadoc:aggregate' under java 8
fails if the javadoc tags are incomplete. To ensure the javadoc remains
complete we include it in the build.

Fixes #1209

### DIFF
--- a/testng/src/main/java/cucumber/api/testng/AbstractTestNGCucumberTests.java
+++ b/testng/src/main/java/cucumber/api/testng/AbstractTestNGCucumberTests.java
@@ -23,8 +23,9 @@ public abstract class AbstractTestNGCucumberTests {
     }
 
     /**
-     * @return returns a two dimensional array of {@link PickleEventWrapper} scenarios
-     * with their associated {@link CucumberFeatureWrapper) feature.
+     * Returns two dimensional array of PickleEventWrapper scenarios with their associated CucumberFeatureWrapper feature.
+     *
+     * @return a two dimensional array of scenarios features.
      */
     @DataProvider
     public Object[][] scenarios() {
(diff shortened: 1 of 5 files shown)


testng/src/main/java/cucumber/api/testng/AbstractTestNGCucumberTests.java
```diff
@@ -23,8 +23,9 @@ public void runScenario(PickleEventWrapper pickleWrapper, CucumberFeatureWrapper
     }
 
     /**
-     * @return returns a two dimensional array of {@link PickleEventWrapper} scenarios
```
mpkorstanje: The links broke the javadoc for some reason. Can't seem to fix it.

mlvandijk:
Verified that generating javadoc using command 'mvn javadoc:aggregate' under java 8 no longer fails due to missing tags (as it was before). 
Looks good!

Issue #729, opened 2014-06-10
Test fails : HTMLFormatterTest#writes_valid_report_js


yoelb:
If host JVM Locale is different of en or us, this test fail.
It seemes that Locale is initialised without parameter provided in parent pom cucumber-jvm.xml:  
         <plugin>
                    <groupId>org.apache.maven.plugins</groupId>
                    <artifactId>maven-surefire-plugin</artifactId>
                    <version>2.17</version>
                    <configuration>
                        <argLine>-Duser.language=en</argLine>
                        <argLine>-Xmx1024m</argLine>
                        <argLine>-XX:MaxPermSize=256m</argLine>
                        <argLine>-Dfile.encoding=UTF-8</argLine>
                        <useFile>false</useFile>  
                    </configuration>
                </plugin>

A dirty workaround is to redefine  <argLine>-Duser.language=en</argLine> in cucumber-core.xml.

If someone has an better and elegant solution, please submit it.

mpkorstanje:
@mlvandijk I'm guessing we're not passing a locale properly somewhere. But this issue is rather old. The first step should be reproducing it.

mlvandijk:
## Root cause:
The test fails because the expected error message is in English. If Locale is set to different language (using Locale.setDefault()), the actual error message is in the language set as default.

## Steps to reproduce:
Version: Current master branch (commit a6dff1d)

Set a different Locale at the beginning of the test, for instance:
` Locale.setDefault(new Locale("fr", "FR"));`

Expected: Error message starts with: "ReferenceError: \"document\" is not defined."
Actual error message starts with: "ReferenceError: "document" n'est pas défini" (i.e. in French, not English)

Sidenote: I could not reproduce this by setting Language to Dutch on OS X and restarting (which I expect would have worked on Windows)

## Possible solutions:
1. Set Locale to new Locale("en", "US") explicitly at the start of this test. 
> After you set the default locale with this method, subsequent calls to Locale.getDefault() will return the newly set locale. (http://www.oracle.com/technetwork/articles/javase/locale-140624.html#using)

It could be reset to the original Locale using a finally block, which is expected to always run.
Creating PR for this solution to illustrate; if OK then this solution can be used.

2. An easier solution might be to limit the check on expected error message to check for startsWith(""ReferenceError: \"document\""), but I'm not sure if the test would then still have the same value.

mpkorstanje:
[...]

PR #1218, opened 2017-09-06
Rename and add test for  HTMLFormatter

mlvandijk:
## Summary

Rename and add test for HTMLFormatter

## Details

Trying to duplicate / fix https://github.com/cucumber/cucumber-jvm/issues/729 

Cannot reproduce this bug by setting Locale for my laptop, or running the test with VM options: -Duser.language=fr (or nl, for that matter). Only by explicitly setting the Locale in the test (see previous PR).

## Motivation and Context

The original test does not validate that report.js is valid; it validates that an empty report throws a particular exception. I've renamed this test.
As suggested, I've added a test that does check for valid Javascript. I'm not sure this adds a lot of value, as other tests are also checking for valid js.

## How Has This Been Tested?
'mvn clean install' => Build Success

N/A
## Types of changes

## Checklist:

### DIFF
--- a/core/src/test/java/cucumber/runtime/formatter/HTMLFormatterTest.java
+++ b/core/src/test/java/cucumber/runtime/formatter/HTMLFormatterTest.java
@@ -12,9 +12,6 @@ import org.jsoup.nodes.Document;
 import org.jsoup.nodes.Element;
 import org.junit.Test;
 import org.mockito.stubbing.Answer;
-import org.mozilla.javascript.Context;
-import org.mozilla.javascript.EcmaError;
-import org.mozilla.javascript.tools.shell.Global;
 
 import java.io.File;
 import java.io.InputStreamReader;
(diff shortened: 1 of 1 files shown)


core/src/test/java/cucumber/runtime/formatter/HTMLFormatterTest.java
```diff
@@ -58,8 +53,8 @@ public void writes_index_html() throws Throwable {
         assertEquals("", reportElement.text());
     }
 
-    @Test
-    public void writes_valid_report_js() throws Throwable {
+    @Test(expected = EcmaError.class)
+    public void empty_report_js_throws_error() throws Throwable {
```
mpkorstanje: Wasn't this this the test that caused the problems with the locale because the EcmaError contained a localized message?

mlvandijk: Yes it was. But I cannot reproduce that either by setting my laptops locale or passing a VM argument (see PR description).
I can only reproduce it by explicitly setting the Locale in the test and the resetting after.

mlvandijk: As discussed, removed the test that checks for error and replaced with test that verifies content of reportJs.

core/src/test/java/cucumber/runtime/formatter/HTMLFormatterTest.java
```diff
@@ -61,15 +58,38 @@ public void writes_index_html() throws Throwable {
...
-        Context cx = Context.enter();
-        Global scope = new Global(cx);
-        try {
-            cx.evaluateReader(scope, new InputStreamReader(reportJs.openStream(), "UTF-8"), reportJs.getFile(), 1, null);
-            fail("Should have failed");
-        } catch (EcmaError expected) {
-            assertTrue(expected.getMessage().startsWith("ReferenceError: \"document\" is not defined."));
-        }
+        String reportJs = FixJava.readReader(new InputStreamReader(new URL(outputDir, "report.js").openStream(), "UTF-8"));
+        assertEquals("$(document).ready(function() {var formatter = new CucumberHTML.DOMFormatter($('.cucumber-report'));formatter.uri(\"some\\\\windows\\\\path\\\\some.feature\");\n" +
```
brasmusson: @mlvandijk @mpkorstanje assertEquals cannot be used to compare things with json data structures in them, since json data structures should be considered equal also if the attribute order differs. Thus we got a build failure when this PR was merged: https://travis-ci.org/cucumber/cucumber-jvm/jobs/273568042.
It is now fixed in https://github.com/cucumber/cucumber-jvm/commit/fb1595232f4a7fddfde33c9c3c63928fd906707b

PR #1220, opened 2017-09-09
Skip test step execution if --dry-run is specified

adrian-baker:
## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/api/TestCase.java
+++ b/core/src/main/java/cucumber/api/TestCase.java
@@ -13,14 +13,16 @@ import java.util.List;
 public class TestCase {
     private final PickleEvent pickleEvent;
     private final List<TestStep> testSteps;
+    private final boolean dryRun;
 
-    public TestCase(List<TestStep> testSteps, PickleEvent pickleEvent) {
+    public TestCase(List<TestStep> testSteps, PickleEvent pickleEvent, boolean dryRun) {
         this.testSteps = testSteps;
         this.pickleEvent = pickleEvent;
+        this.dryRun = dryRun;
     }
 
-    public void run(EventBus bus, boolean dryRun) {
-        boolean skipNextStep = dryRun;
+    public void run(EventBus bus) {
+        boolean skipNextStep = this.dryRun;
         Long startTime = bus.getTime();
         bus.send(new TestCaseStarted(startTime, this));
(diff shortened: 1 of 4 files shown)


core/src/test/java/cucumber/runner/RunnerTest.java
```diff
@@ -84,6 +84,25 @@ public void hooks_execute_also_after_failure() throws Throwable {
     }
 
     @Test
+    public void steps_are_not_executed_on_dry_run() throws Throwable {
```
brasmusson: This test verifies two things. I think it should make into two tests.

core/src/main/java/cucumber/api/TestCase.java
```diff
@@ -13,14 +13,16 @@
 public class TestCase {
     private final PickleEvent pickleEvent;
     private final List<TestStep> testSteps;
+    private final boolean dryRun;
 
-    public TestCase(List<TestStep> testSteps, PickleEvent pickleEvent) {
```
brasmusson: Because this is an `api` package, should the ctor with the current signature be kept for backward compatibility? @mpkorstanje WDYT?

mpkorstanje: Along with adding a deprecation tag that would be the correct course of of action.

In the long run (and I'm only just starting to think about this) I'd like TestCase to be an interface so we can exclude the constructors and run method from the public api. These methods aren't needed as part of the event bus protocol.

mpkorstanje: Can we push this as a extra commit on this feature branch so we don't have to bother Adrian with this?

mpkorstanje: I can't. Permissions have apparently not been set.

mpkorstanje: I've pushed my suggested changes to https://github.com/cucumber/cucumber-jvm/tree/adrian-baker/master. I reckon we can finish up this discussion the PR for that branch. I've got no further objections to this branch being merged.

adrian-baker: Great, thanks. Yes wasn't comfortable with having to change an api package, but in the current state didn't seem avoidable.

aslakhellesoy:
Hi @adrian-baker,

Aslak Hellesøy
Creator of Cucumber

PR #1221, opened 2017-09-09
Allow RuntimeOptions to be created by subclasses

a1dutch:
## Summary

Allow creation of `RuntimeOptions` without use of the `CucumberOptions` annotation.

## Details

Having tried to configure plugins in an overridden `createRuntime` method, these plugins did not get applied.  This plugin extracts the existing logic into a method, and allows for it to be overridden.

## Motivation and Context

To allow `RuntimeOptions` to be overridden by subclasses of `Cucumber` allowing programatic configuration.

## How Has This Been Tested?

Existing test suite passes.

## Types of changes

## Checklist:

### DIFF
--- a/junit/src/main/java/cucumber/api/junit/Cucumber.java
+++ b/junit/src/main/java/cucumber/api/junit/Cucumber.java
@@ -56,8 +56,7 @@ public class Cucumber extends ParentRunner<FeatureRunner> {
         ClassLoader classLoader = clazz.getClassLoader();
         Assertions.assertNoCucumberAnnotatedMethods(clazz);
 
-        RuntimeOptionsFactory runtimeOptionsFactory = new RuntimeOptionsFactory(clazz);
-        RuntimeOptions runtimeOptions = runtimeOptionsFactory.create();
+        RuntimeOptions runtimeOptions = createRuntimeOptions(clazz);
 
         ResourceLoader resourceLoader = new MultiLoader(classLoader);
         runtime = createRuntime(resourceLoader, classLoader, runtimeOptions);
@@ -66,6 +65,17 @@ public class Cucumber extends ParentRunner<FeatureRunner> {
         final List<CucumberFeature> cucumberFeatures = runtimeOptions.cucumberFeatures(resourceLoader, runtime.getEventBus());
         jUnitReporter = new JUnitReporter(runtime.getEventBus(), runtimeOptions.isStrict(), junitOptions);
         addChildren(cucumberFeatures);
+    }
+
+    /**
+     * Create the runtime options. Can be overridden to customize the runtime options.
+     * 
+     * @param clazz
+     *            the class with the @RunWith annotation.
+     * @return runtime options
+     */
+    protected RuntimeOptions createRuntimeOptions(Class clazz) {
+        return new RuntimeOptionsFactory(clazz).create();
     }
 
     /**


mpkorstanje:
What is your usecase for extending the Cucumber.class rather then duplicating it?

a1dutch:
Duplicating the class means i have to maintain the whole class, rather than the one method to configure the options, the class has an existing `createRuntime` method so looks like it is meant to be subclassed by extensions.

mpkorstanje:
Yes, but what is your usecase? What is the need that drives your wish to customize the RuntimeOptions?

a1dutch:
We are using the cucumber/runner per-package to get round the problems with duplicate names in step definitions, what i'm using a custom cucumber for is to configure the output paths for plugins, the tags based on class name and in the future add some common steps to the glue.

Here's my `Cucumber` extension i used with this pull request.

`package my.cucumber;

import cucumber.api.junit.Cucumber;
import cucumber.runtime.RuntimeOptions;
import java.io.IOException;
import java.util.Arrays;
import org.junit.runners.model.InitializationError;

public class CustomisedCucumber extends Cucumber {

    public CustomisedCucumber(Class clazz) throws InitializationError, IOException {
        super(clazz);
    }

    @Override
    protected RuntimeOptions createRuntimeOptions(Class clazz) {
        return newRuntimeOptions();
    }

    private RuntimeOptions newRuntimeOptions() {
        return new RuntimeOptions(
            Arrays.asList("--tags ", "@" + findTag(),
                          "--plugin", "pretty", 
                "--plugin", "json:build/cucumber-results/" + findTestName() + ".json",
                          "--strict",
                          "--snippets", "UNDERSCORE",
                          "classpath:" + getTestClass().getJavaClass().getPackage().getName().replaceAll("\\.", "/"),
                          "--glue", "classpath:" + getTestClass().getJavaClass().getPackage().getName().replaceAll("\\.", "/")));
    }

    private String findTestName() {
        String testClass = getTestClass().getJavaClass().getPackage().getName();
        testClass = testClass.substring(testClass.indexOf("tests.") + 6);
        return testClass + "-" + findTag().toLowerCase();
    }

    private String findTag() {
        String testClass = getTestClass().getJavaClass().getSimpleName();
        return testClass.substring(0, testClass.indexOf("Test"));
    }
}
`

mpkorstanje:
My concern with opening up a class for extension, especially one that is also part of the public API is that this will make it harder to improve `cucumber-jvm` in the future without introducing breaking changes.

As is I'm not happy with the existing extension point, it exposes parts of the internal API and allows the class to be altered in a manner that does not satisfy any reasonable "instance of" relation. The same reasoning applies to your use case. It is effectively subverting the established method of configuring cucumber and substituting it its own.

On the other-hand  [Cucumber.java](https://github.com/cucumber/cucumber-jvm/blob/master/junit/src/main/java/cucumber/api/junit/Cucumber.java) contains about 60 lines of code. Most of it rather uncomplicated boiler plate. 

As such I have to weigh my desire to avoid a potential future headache over the improper use of inheritance against the general utility of your use case. I find this hard to judge but it does not seem to be a generic use case. So as it currently stands I think it best that you duplicate the runner. The code is not complicated and it seems right that those who benefit from a special use case also carry burden of maintaining it.

@aslakhellesoy @brasmusson how would you judge the general utility of this use case?

brasmusson:
Currently we have the mechanism of "inheriting" `@CucumberOptions` annotations for reusing option settings. As in:
```
@CucumberOptions(strict=true, snippets=UNDERSCORE);
class RunnerBase {}
```
```
@CucumberOptions(glue={"<package>"}, features={"classpath:<path>"});
@RunWith(Cucumber.class);
class OneRunner extends RunnerBase {}
```
It is also possible to override the options from the `@CucumberOptions` annotation using the cucumber.options property (as in `mvn -Dcucumber.options="<options>" test`)

Before introducing a new mechanism, we should thoroughly evaluate whether the current mechanisms can support this situation.

I do not know the intended use case for the current extension point of the `Cucumber` class (the protected `createRuntime` method).

a1dutch:
i've just copied the cucumber code as you suggest, thanks

PR #1226, opened 2017-09-20
[Spring] Throw when step definitions have the Component stereotype

### LINKED ISSUE #1225: [Spring] Exception is thrown complaining about multiple matching beans with version 1.2.5
In the process of transitioning from version Cucumber 1.1.8 to 1.2.5 I have ran into an issue where the below exception is thrown:

`CucumberException No qualifying bean of type x is defined: expected single matching bean but found 2`

I tried to dig a bit deeper and I see that the `<T> T getInstance(final Class<T> type)` method in `SpringFactory` class had changed and apparently is somehow responsible for that behavior.

I was able to replicate this behavior in a simple project which I uploaded to github and the link is below.

Cucumber Version: 1.2.5
Operating System and version: Windows 7
Link to your project: https://github.com/EugeneS30/cucumber-bean-defect-verification.git

mpkorstanje:
When step definitions are annotated with `@Component` or other related
annotations they can be picked up by springs class path scanning. This
conflicts with cucumbers class path scanning and may result in
multiple bean definitions for the same class.

This problem is hard to understand and hard to trace. By making the
problem explicit and providing a clear instruction on how to resolve
this we can hopefully avoid future confusion.

The current implementation only checks the a subset of all annotations.
This will hopefully be sufficient.

Closes #1225

## Types of changes

## Checklist:

### DIFF
--- a/gosu/pom.xml
+++ b/gosu/pom.xml
@@ -53,10 +53,5 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.mockito</groupId>
-            <artifactId>mockito-all</artifactId>
-            <scope>test</scope>
-        </dependency>
     </dependencies>
 </project>
--- a/spring/src/main/java/cucumber/runtime/java/spring/SpringFactory.java
+++ b/spring/src/main/java/cucumber/runtime/java/spring/SpringFactory.java
@@ -13,6 +13,7 @@ import org.springframework.beans.factory.support.BeanDefinitionRegistry;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.support.ClassPathXmlApplicationContext;
 import org.springframework.context.support.GenericApplicationContext;
+import org.springframework.stereotype.Component;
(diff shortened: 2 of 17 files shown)


spring/src/main/java/cucumber/runtime/java/spring/SpringFactory.java
```diff
@@ -73,6 +87,20 @@ public boolean addClass(final Class<?> stepClass) {
         return true;
     }
 
+    private void checkNoComponentAnnotations(Class<?> type) {
```
mpkorstanje: This is not going to work. This should recursively search the annotations annotations for `@Component` taking care not to get stuck in a loop.

gosu/pom.xml
```diff
@@ -53,10 +53,5 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.mockito</groupId>
-            <artifactId>mockito-all</artifactId>
```
mpkorstanje: mockito-all packages an older version hamcrest which conflicts with the one provided by junit and should not be used.

PR #1238, opened 2017-10-04
[Core] Optimize MethodScanner

lsuski:
## Summary

Optimized MethodScanner.scan method to not unnecessarily check for all cucumber annotations

## Details
MethodScanner scans each method for existence of cucumber annotation (there are many). It is redundant and slow because only methods with Before/After or StepDefAnnotation are added to backend. 

## Motivation and Context

On Android < 7.0 it is extremely slow, and starting with debugger is almost impossible. I've measured it and optimized method is about 150 times faster on Android <7.0 and about 10 times faster on Android >=7.0

## How Has This Been Tested?

MethodScannerTest passes. 
I've also tested it on my Android project and it works (and it is much more faster)

## Types of changes

### DIFF
--- a/java/src/main/java/cucumber/runtime/java/MethodScanner.java
+++ b/java/src/main/java/cucumber/runtime/java/MethodScanner.java
@@ -2,25 +2,22 @@ package cucumber.runtime.java;
 
 import cucumber.api.java.After;
 import cucumber.api.java.Before;
+import cucumber.runtime.ClassFinder;
 import cucumber.runtime.CucumberException;
 import cucumber.runtime.Utils;
-import cucumber.runtime.ClassFinder;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
-import java.util.Collection;
 import java.util.List;
 
 import static cucumber.runtime.io.MultiLoader.packageName;
 
 class MethodScanner {
-    private final Collection<Class<? extends Annotation>> cucumberAnnotationClasses;
 
     private final ClassFinder classFinder;
 
     public MethodScanner(ClassFinder classFinder) {
         this.classFinder = classFinder;
-        cucumberAnnotationClasses = findCucumberAnnotationClasses();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


java/src/main/java/cucumber/runtime/java/MethodScanner.java
```diff
@@ -56,26 +53,25 @@ public void scan(JavaBackend javaBackend, List<String> gluePaths) {
...
-                if (!glueCodeClass.equals(method.getDeclaringClass())) {
-                    throw new CucumberException(String.format("You're not allowed to extend classes that define Step Definitions or hooks. %s extends %s", glueCodeClass, method.getDeclaringClass()));
-                }
-                if (isHookAnnotation(annotation)) {
-                    javaBackend.addHook(annotation, method);
-                } else if (isStepdefAnnotation(annotation)) {
-                    javaBackend.addStepDefinition(annotation, method);
-                }
+        Annotation[] methodAnnotations = method.getAnnotations();
+        for (Annotation annotation : methodAnnotations) {
```
mpkorstanje: Might be even more efficient to use Method.getAnnotation(Class<T> annotationClass) rather then checking all annotations. Then we can replace this for loop with two null checks.

lsuski: I'm not sure about it. Could you clarify? Based on previous logic one method can be annotated with Before/After and any number of step annotations  - we can't achieve it without loop. Second thing is that we are not able to get step annotation e.g. `@Given` using `Method.getAnnotation(Class annotationClass)` because we don't know what annotation is used (actually cucumber does not care if it is `@Given` or `@Then` it just should be annotation that itself is annotated with `@StepDefAnnotation` ), eventually this code won't work: `method.getAnnotation(StepDefAnnotation.class)`

mpkorstanje: Derp. You are correct. I didn't have hierarchy of annotations properly in my mind.

aslakhellesoy:
Hi @lsuski,

Aslak Hellesøy
Creator of Cucumber

PR #1246, opened 2017-10-11
[Spring] Limit context configuration to a single class

### LINKED ISSUE #1240: [Spring, Kotlin] Kotlin Metadata annotations breaks SpringFactory
## Summary

kotlin-spring does not work together with Kotlin Step classes

## Expected Behavior

*Given* Two different  Kotlin Steps definition classes like 
```
@ContextConfiguration(classes = arrayOf(TestConfig::class))
class SomeSteps1 {
    // some step definitions
}

@ContextConfiguration(classes = arrayOf(TestConfig::class))
class SomeSteps2 {
    // some step definitions
}
```
*When* A scenario is executed using this step class as glue code
*Then* The scenario is executed without an error 
## Current Behavior

An exception is thrown: 
` cucumber.runtime.CucumberException: Annotations differs on glue classes found: com.xxx.SomeSteps1, com.xxx.SomeSteps2`

Background: Kotlin adds an additonal Annotation `kotlin.Metadata` to each class. These Metadata differ for each class, therefore the check in SpringFactory.checkAnnotationsEqual() fails. 

## Possible Solution

Before comparing the annotations, remove annotations of type `kotlin.Metadata` (can be done by its name so there is no dependency to Kotlin)

(see above in Expected behaviour)

## Context & Motivation

We use in our project Cucumber with Spring and Java for a while and I tried out to add a Kotlin step definition class, which did not work out that good...

## Your Environment

* Version used: 1.2.5
* Operating System and version: Windows 10
* Link to your project: -

mpkorstanje:
## Summary

Fixes #1240

## Types of changes

## Checklist:


mpkorstanje:
Documentation can be updated after #1245 is merged

PR #1248, opened 2017-10-12
[Core] Implement cucumber expressions

### LINKED ISSUE #1041: Add support for Cucumber Expressions
## Summary

[Cucumber Expressions](http://docs.cucumber.io/cucumber-expressions/) is a new alternative to
Regular Expressions, which is easier to read and write, and less daunting.
## Expected Behavior

Step Defintions that don't start with a `^` or end with a `$` should be interpreted as a Cucumber Expression:

``` Java
@Given("I have {n} cukes in my belly")
```

Alternatively we can decide to use regular expressions if the string uses `/` delimiters:

``` Java
@Given("/I have (\\d+) cukes in my belly/")
```

Whatever mechanism we choose to differentiate is likely to break existing step definitions, so we'd have to make a new major release and document how to migrate.

Snippets for undefined steps should use Cucumber Expressions, and not Regular Expressions. We want users to start migrating, and if we can avoid it we'll not generate snippets with regular expressions. (Well, we might if enough people want it). The `cucumber-expression` library can already generate expressions from text.

Finally, the current `Transform` logic should be deprecated and the logic built into the `cucumber-expression` library should be used instead.
## Context & Motivation

Having to use Regular Expressions is one of the most common complaints against Cucumber. Other clones of Cucumber who support something similar to Cucumber Expressions seem to have happier users!

mpkorstanje:
## Summary

* Implements [Cucumber expressions](https://docs.cucumber.io/cucumber-expressions/) 
* Removes XStream
* Adds DataTables from https://github.com/cucumber/cucumber/pull/291
* Introduces the concept of StepExpressions which combines a CucumberExpression and DataTable expression. Currently data table expressions are limited to a name or java type. Currently only the implicit type derived from the method argument is used.

For a working example check the calculator examples.

## How to build

Install (`mvn install`) `datatables/java` from `cucumber/cucumber`
Then install this branch. If it fails you may need to remove `pax-exam` from `examples/pom,xml`. This is okay.

## Todo 
 * [x] Fix android examples
 ~~* [ ] Add link to configuration docs in exceptions~~
 * [x] Test SnippetGenerators and StepdefGenerator icm Idea
 ~~* [ ] Update documentation~~
 * [x] Write blog post
 ~~* [ ] File IDEA bug about "Pattern expected" warning~~
 * [x] Undefined parameter type error should be similarly friendly as missing table type

This closes  #1041


examples/android/android-studio/Cukeulator/app/src/androidTest/java/cucumber/cukeulator/test/ParameterTypes.java
```diff
@@ -0,0 +1,44 @@
...
+import static java.util.Locale.ENGLISH;
+
+public class ParameterTypes implements Configuration {
+
+    @Override
+    public ParameterTypeRegistry createParameterTypeRegistry() {
+        ParameterTypeRegistry parameterTypeRegistry = new ParameterTypeRegistry(ENGLISH);
+        parameterTypeRegistry.defineParameterType(new ParameterType<Integer>(
+            "digit",
+            "([0-9])",
```
aslakhellesoy: You don't need a capture group here. Only use capture group if you want to capture more than one thing.

aslakhellesoy:
cucumber-expressions 5.0.2 is released, and that fixes the pax stuff!

aslakhellesoy:
> What do we do with @Format, @Transform and @Delimiter

We get rid of them - Cucumber Expression offers similar (and better IMHO) functionality. When we write a migration guide we'll get a better feel for how hard it will be for users to migrate. If it seems hard, perhaps we'll add them back, ideally delegating functionality to Cucumber Expressions (and deprecating them). We should only do this if we think migrating is *really* hard for people.

mpkorstanje:
@aslakhellesoy I added a table converter implementation in the spike https://github.com/cucumber/cucumber/pull/291.

rjwittams:
Things that would be nice given that Transforms are being reworked: 
     - Initialize the Transformer through the DI container. This way it can be quite easy to e.g. look things up defined is a previous steps data table, which can make steps cleaner.
     - Allow the transformers to be registered globally for the whole test run rather than needing to annotate each parameter. 
Not sure if these points are already covered. Thanks

mpkorstanje:
@rjwittams
>  - Initialize the Transformer through the DI container. This way it can be quite easy to e.g. look things up defined is a previous steps data table, which can make steps cleaner.

I don't think we'll do that in this version. This change is big enough as it is already. Once we're done it should be perfectly possible to create a system that register transformations in the same way we register step definitions. Like so:

```java
    MaConfigurationFrançaise(TrafficLightDatabase database){

       ExpressionDeConcombre("vert", "fue (rouge|jaune|vert)",
           groupValues -> database.createTrafficLight(groupValues);

   }
```

> - Allow the transformers to be registered globally for the whole test run rather than needing to annotate each parameter.

This is the current implementation. All transformers are global. See https://docs.cucumber.io/cucumber-expressions/ for some explanation.

mlvandijk:
Partial review:
1. Using a double in a step does not generate correct snippet:
The generated snippet using a double has two arguments, both ints (I'm assuming it's the numbers before and after the decimal)
Running it leads to cucumber.runtime.CucumberException: Arity mismatch: Step Definition 'expressions.StepDefs.i_have_cucumbers_in_my_belly(int,int)
Example:
```
    @Given("I have {double} cucumbers in my belly")
    public void i_have_cucumbers_in_my_belly(int arg1, int arg2) {
        // Write code here that turns the phrase above into concrete actions
        throw new PendingException();
    }
```
2. Expression {string} is not recognized?
Replacing multiple words in a generated stepdef with {string} is not recognized; I still get a message saying "You can implement missing steps with the snippets below:"

with the following snippet:
```
@Then("my test should very surely pass")
public void my_test_should_very_surely_pass() {
    // Write code here that turns the phrase above into concrete actions
    throw new PendingException();
}
```

Also, several things don't work yet with the IntelliJ plugin, which is to be expected. It'll need an update.

For more details, see [cucumber-expressions test project on GitHub](https://github.com/mlvandijk/cucumber-expression-test), specifically the [notes in StepDefs.java](https://github.com/mlvandijk/cucumber-expression-test/blob/master/src/test/java/expressions/StepDefs.java)

mpkorstanje:
@mlvandijk can you tell me what your default locale is? `Locale.getDefault()`

mpkorstanje:
Thanks! I've added some in detailed comments in your test project.

1. Fixed.
2. That's how it is supposed to work. `{string}` expects the contents to be delimited by qoutes.

mlvandijk:
1. Thnx, will try it out
2. Wasn't aware of the quotes, will try that (and update docs as needed ;) )

mlvandijk:
**Test overview:**
Parameter types:
- [x] {int}
- [ ] {byte} **does not work** (see above)
- [x] {short}
- [x] {long}
- [x] {double} *fixed*
- [x] {float}
- [x] {word}
- [x] {string} **Note:** update doc to clarify needing to use double (`"`) or single (`'`) quotation marks
- [ ] {bigint} - TODO
- [ ] {bigdec} - TODO
- [ ] Custom parameter types - TODO
- [x] Optional text works. e.g. cucumber(s)
- [x] Alternative text works, e.g. belly/stomach
- [ ] Escaping - TODO
- [ ] Regular Expressions - TODO what/how to test?
- [ ] Preferential parameter types- TODO: how?

In addition, I'd like to try updating several work projects to this version to see if that works, or whether I find anything that doesn't.

mlvandijk:
Updated needed to IntelliJ Cucumber plugin:
- IntelliJ gives an error on {int} or {double} or other number type saying "Number expected"
- IntelliJ also gives an error on {word} or {string} saying "Number expected"
- Steps using cucumber expressions are not recognized by IntelliJ

mlvandijk:
Updated test results for {bigint} and {bigdec} (NOK - see above)

mpkorstanje:
> Updated test results for {bigint} and {bigdec} (NOK - see above)

Snippets with big integers and big decimals are intentionally not generated by default. Their pattern is the same as that of an integer. 

>  {byte} NOK (see above)

Snippets with bytes are also intentionally not generated. I can't think of a good technical reason for this though.

mpkorstanje:
> In addition, I'd like to try updating several work projects to this version to see if that works, or whether I find anything that doesn't work as expected.

Awesome!

mpkorstanje:
I've updated the build instructions. Aslak merged the data table PR.

mpkorstanje:
>  Note: replacing a {double} with a {bigdec} throws io.cucumber.cucumberexpressions.UndefinedParameterTypeException: Undefined parameter type {bigdec}

The name is `{bigdecimal}`.  You can find the list of pre-configured types in the [ParameterTypeRegistry](https://github.com/cucumber/cucumber/blob/master/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ParameterTypeRegistry.java#L29-L92)

mpkorstanje:
> Note: by default generates a step using {int} which then throws a NumberFormatException. It does work when replacing {int} with {bigint} and argument int with BigInteger. Still, this is not as nice as it could be?

Good point. 

Unfortunately fixing this by improving the regex to only match the possible values for `int32` would also cause the common `/\d+/` pattern to not be recognized as the pattern for a number (in this case int) any more.

mlvandijk:
@mpkorstanje Thanks for the additional information. Note that I've tested based on the documentation, so any misunderstandings on my part should maybe be clarified in the docs. I can do a PR for that :)

mlvandijk:
Updated test overview (see above).
1. Ok: {bigint} and {bigdecimal} - although they could be consistent, either full name or abbreviated for both?
2. byte is NOK; a step def with {byte} is not recognized even though my expectation based on docs is it should work just like {short} does.
I'll do some work project now ;)

aslakhellesoy:
@mlvandijk I'll make a change in cucumber-expressions to use `{biginteger}` and `{bigdecimal}`.

mlvandijk:
@aslakhellesoy Beats me, but imho either it should be clear how it works OR just remove it if not needed).

aslakhellesoy:
byte, short, int and long are all parsed with tjeir boxed type’s `#decode(String)` method. Would that be sufficient for docs?

I think we should support all number types for the sake of completeness/consistency, even if people will use just int 90% of the time, long or double 9% of the time and the other types 1% of the time.

mpkorstanje:
The work flow Marit is using is 1) write the step with the proper values,  2) run the test 3) copy past the snippets. Then it fails to work for numbers that are too large and doesn't generate a snippet when bytes are written in HEX. 

I think we can turn on snippet generation for bytes. But I don't have a good solution yet that works for big numbers.

mpkorstanje:
Also  big decimal fails for work for decimal numbers. It's not using the `FLOAT_REGEXPS`. Derp.

aslakhellesoy:
I'll add some more tests to cucumber expressions to expose those number issues, and hopefully add some fixes.

mpkorstanje:
Eh. I may have fixed them already: https://github.com/cucumber/cucumber/pull/354

mpkorstanje:
```
Given I want to generate a snippet that uses a big integer
When I define a step that uses a number that is larger then an integer 1234567890123456789012345678901234567890
Then I expect the cucumber expression to contain `{bigdecimal}`
And  I expect the method parameter  type to be `BigDecimal`
```

aslakhellesoy:
@mlvandijk `{byte}` uses a different pattern from `{int}`, `{short}` and `{long}`:

https://github.com/cucumber/cucumber/blob/c63637985f396302bcb9db69506cf438774ab055/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ParameterTypeRegistry.java#L23

You might as well replace the `{int}` with `{colour}` if you know what I mean - they've got different patterns.

mlvandijk:
@aslakhellesoy Yeah, I remember now. Thnx. Will try with byte pattern tomorrow. :)

aslakhellesoy:
I took 3.0.0.SNAPSHOT for a spin. My overall experience was great! I'm really excited to get this out the door now :-)

Existing regular expression stepdefs worked flawlessly.

I have some suggestions for improving the DX. The smoother we can get this, the more people will upgrade without needing help.

Existing table stepdef:

```java
@Given("^people are located at:$")
public void people_are_located_at(List<Whereabout> whereabouts) {}
``` 

This gave me the following error:

```
io.cucumber.datatable.CucumberDataTableException: Can't convert DataTable to List<class shouty.LocationSteps$Whereabout>. Please register a DataTableType with a TableEntryTransformer or TableRowTransformer for class shouty.LocationSteps$Whereabout
```

I think `List<class shouty.LocationSteps$Whereabout>` should be `List<shouty.LocationSteps$Whereabout>` (without `class `).

It wasn't obvious to me how I would register a `DataTableType`. After poking around the source code I found out there is a `cucumber.api.Configuration` interface I can implement. Most users won't be able to figure this out on their own. We'll document it, but the error message should provide enough information to help users get past it. It would be better if the error message said something like:

```
Define a class that implements `cucumber.api.Configuration` and register a DataTableType with a TableEntryTransformer or TableRowTransformer for class shouty.LocationSteps$Whereabout.
See https://docs.cucumber.io/cucumber/#data-tables for more details.
```

(The docs aren't there yet, but I've picked them up again this week).

I then realised that `TypeRegistry#parameterTypeRegistry()` and `TypeRegistry#dataTableTypeRegistry()`
are public. I think it would be more user friendly if those accessors were not available.

It was really helpful that I was asked to define one of `TableEntryTransformer` or `TableRowTransformer`,
because the `DataTableType` constructor also allows `TableTransformer` and `TableCellTransformer`.
I guess it's smart enough to understand that the latter won't work for a `List`!

However, I wasn't sure which one to use, so I just picked `TableRowTransformer` at first. That gave
me a method accepting a `List<String>` which I assumed was wrong, so I went with the other one
and ended up with this:

```java
typeRegistry.defineDataTableType(new DataTableType(
                "whatever",
                LocationSteps.Whereabout.class,
                (TableEntryTransformer<LocationSteps.Whereabout>) entry -> new LocationSteps.Whereabout(
                        entry.get("personName"),
                        Integer.parseInt(entry.get("x")),
                        Integer.parseInt(entry.get("y"))
                )));
```

That worked! I couldn't help thinking that this was rather verbose. Why do I have to do the
integer parsing? Why can't Cucumber do this for me using reflection like it used to do?

I'd love to be able to do this:

```java
typeRegistry.defineDataTableType(new DataTableType(
        "whatever",
        LocationSteps.Whereabout.class, 
        new DefaultTableEntryTransformer(LocationSteps.Whereabout.class)
));
```

Or even:

```java
typeRegistry.defineDataTableType(new DataTableType(
        "whatever",
        LocationSteps.Whereabout.class
));
```

We could make it simple and require an empty constructor.

Another thing that puzzled me was the name of the type, so I just used "whatever". From what I can tell, the table type name is only used in error messages. Could we remove it altogether and use the type name in error messages instead?

This would simplify both usage and implementation.

Next I tried out Cucumber Expressions:

```java
@Given("{word} is at {coordinate}")
public void person_is_at(String personName, Coordinate coordinate) {}
```

IntelliJ (trying to be smart) underlined my expression with a tooltip saying "Pattern expected". We should file a JetBrains bug about this.

This time I got the following error:

```
cucumber.runtime.CucumberException: io.cucumber.cucumberexpressions.UndefinedParameterTypeException: Undefined parameter type {coordinate}
```

This was thrown from the Cucumber Expressions library. I'd like to see a more helpful 
error message, similar to the one I saw for tables.

When I went ahead and defined the ParameterType, I realised I could either use a
`Function` or a `Transformer`. I was confused about which one to use!

So I went ahead and created a PR for that: https://github.com/cucumber/cucumber/pull/357

All in all, really great stuff, but I think we need to polish it a little bit before we release.

Summary checklist:

- [ ] Remove `class ` from `List<class shouty.LocationSteps$Whereabout>` error message
- [ ] Explain `cucumber.api.Configuration` in error message (unless it already exists)
- [ ] Hide TypeRegistry#parameterTypeRegistry()` and `TypeRegistry#dataTableTypeRegistry()`
- [ ] Remove `DataTableType`'s `name` property (?)
- [ ] Add `DefaultTableEntryTransformer`
- [ ] File IDEA bug about "Pattern expected" warning
- [ ] Undefined parameter type error should be similarly friendly as missing table type
- [ ] Merge and use https://github.com/cucumber/cucumber/pull/357 (?)

mpkorstanje:
> I took 3.0.0.SNAPSHOT for a spin. My overall experience was great! I'm really excited to get this out the door now :-)

Awesome! Glad to hear that.

> All in all, really great stuff, but I think we need to polish it a little bit before we release.

Definitely! 

> I think List<class shouty.LocationSteps$Whereabout> should be List<shouty.LocationSteps$Whereabout> (without class).

I'll see if I can do this without building a custom type system! :D

>  I guess it's smart enough to understand that the latter won't work for a List!

It's even smarter! [If you would have created a table with a single column the suggestion would have included the `TableCellTransformer`](
https://github.com/cucumber/cucumber/blob/master/datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistryTableConverter.java#L129-L135). :D

> However, I wasn't sure which one to use, so I just picked TableRowTransformer at first. That gave
me a method accepting a List<String> which I assumed was wrong, so I went with the other one
and ended up with this:
> ```typeRegistry.defineDataTableType(new DataTableType(
>                "whatever",
>                LocationSteps.Whereabout.class,
>                (TableEntryTransformer<LocationSteps.Whereabout>) entry -> new LocationSteps.Whereabout(
>                        entry.get("personName"),
>                        Integer.parseInt(entry.get("x")),
>                        Integer.parseInt(entry.get("y"))
>                )));
>```

You can clean that up by using `(Map<String, String> entry) -> ....` but the choice of what to map will stay. I've explained all possible transformations [here](https://github.com/cucumber/cucumber/tree/master/datatable#introduction).

The only way to really reduce the choice between the different transformers is to provide a default object mapper. This will transfer the complexity from understanding cucumber to understanding the object mapper but I expect it will also just work most of the time.

> That worked! I couldn't help thinking that this was rather verbose. Why do I have to do the integer parsing? Why can't Cucumber do this for me using reflection like it used to do?

For now you can use the TableEntryTransformer in combination with your favorite object mapper (using Jackson here) to avoid most of the boiler plate. 

```java
public class ParameterTypes implements Configuration {

    private final ObjectMapper objectMapper = new ObjectMapper();

    @Override
    public TypeRegistry createTypeRegistry() {
   
        parameterTypeRegistry.defineDataTableType(new DataTableType(
            "entry",
            Entry.class,
            (Map<String, String> row) -> objectMapper.convertValue(row, Entry.class)));
    }
}
```

There are quite a few object mapping frameworks to chose from (XStream, Jackson, Gson and many more). 

No matter which one you chose they all end up leaking meta information into the domain model. This can be annotations or method name conventions. Either that or you end up with a bunch of boiler plate `ClassNameConverter` classes that don't differ much from what you're doing here. While this is acceptable when the object mapping framework is also used in production it is not acceptable for a test system to do this. So any choice we make will be the wrong one. 

As such we should definitely look at making it pluggable and once it is pluggable we might even provide a default. However I also think that we should finish the bare bones version first before adding in the next feature.

> Another thing that puzzled me was the name of the type, so I just used "whatever". From what I can tell, the table type name is only used in error messages. Could we remove it altogether and use the type name in error messages instead?

I was hoping I'd be able to hammer out a cucumber expression like syntax for data tables so it works with lambdas and other weakly typed languages but I didn't get around to it. You'd do something like:

```java
Given("a table of persons:", "list of person", (List<Person> -> { .... })
Given("a table of persons:", "map of string to person", (Map<String, Person> -> { .... })
Given("a table of persons:", "map of id to person", (Map<Id, Person> -> { .... })
```

For now I reckon we can hide the name.

mlvandijk:
OK, so today I tried to replace `(\\d+)` in a stepdef with `{int}` and am getting the following error:
```
cucumber.runtime.CucumberException: Error creating bean with name 'cucumber.performance.StepDefs': Bean instantiation via constructor failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [cucumber.performance.StepDefs]: Constructor threw exception; nested exception is java.util.regex.PatternSyntaxException: Illegal repetition near index 15
^the seller has {int} strike(s)$
               ^

	at cucumber.runtime.java.spring.SpringFactory.getInstance(SpringFactory.java:212)
	at cucumber.runtime.java.JavaBackend.buildWorld(JavaBackend.java:127)
	at cucumber.runner.Runner.buildBackendWorlds(Runner.java:155)
	at cucumber.runner.Runner.runPickle(Runner.java:78)
	at cucumber.runtime.junit.PickleRunners$NoStepDescriptions.run(PickleRunners.java:140)
	at cucumber.runtime.junit.FeatureRunner.runChild(FeatureRunner.java:68)
	at cucumber.runtime.junit.FeatureRunner.runChild(FeatureRunner.java:23)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at cucumber.runtime.junit.FeatureRunner.run(FeatureRunner.java:73)
	at cucumber.api.junit.Cucumber.runChild(Cucumber.java:99)
	at cucumber.api.junit.Cucumber.runChild(Cucumber.java:56)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at cucumber.api.junit.Cucumber$1.evaluate(Cucumber.java:108)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'cucumber.performance.StepDefs': Bean instantiation via constructor failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [cucumber.performance.StepDefs]: Constructor threw exception; nested exception is java.util.regex.PatternSyntaxException: Illegal repetition near index 15
^the seller has {int} strike(s)$
               ^
	at org.springframework.beans.factory.support.ConstructorResolver.autowireConstructor(ConstructorResolver.java:279)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.autowireConstructor(AbstractAutowireCapableBeanFactory.java:1193)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1095)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:513)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:483)
	at org.springframework.beans.factory.support.AbstractBeanFactory$2.getObject(AbstractBeanFactory.java:345)
	at cucumber.runtime.java.spring.GlueCodeScope.get(GlueCodeScope.java:14)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:340)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:220)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.resolveNamedBean(DefaultListableBeanFactory.java:1018)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.getBean(DefaultListableBeanFactory.java:345)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.getBean(DefaultListableBeanFactory.java:340)
	at cucumber.runtime.java.spring.SpringFactory.getInstance(SpringFactory.java:210)
	... 28 more
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [cucumber.performance.StepDefs]: Constructor threw exception; nested exception is java.util.regex.PatternSyntaxException: Illegal repetition near index 15
^the seller has {int} strike(s)$
               ^
	at org.springframework.beans.BeanUtils.instantiateClass(BeanUtils.java:154)
	at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:122)
	at org.springframework.beans.factory.support.ConstructorResolver.autowireConstructor(ConstructorResolver.java:271)
	... 40 more
Caused by: java.util.regex.PatternSyntaxException: Illegal repetition near index 15
^the seller has {int} strike(s)$
               ^
	at java.util.regex.Pattern.error(Pattern.java:1955)
	at java.util.regex.Pattern.closure(Pattern.java:3157)
	at java.util.regex.Pattern.sequence(Pattern.java:2134)
	at java.util.regex.Pattern.expr(Pattern.java:1996)
	at java.util.regex.Pattern.compile(Pattern.java:1696)
	at java.util.regex.Pattern.<init>(Pattern.java:1351)
	at java.util.regex.Pattern.compile(Pattern.java:1028)
	at io.cucumber.cucumberexpressions.ExpressionFactory.createExpression(ExpressionFactory.java:28)
	at io.cucumber.stepexpression.StepExpressionFactory.createExpression(StepExpressionFactory.java:62)
	at io.cucumber.stepexpression.StepExpressionFactory.createExpression(StepExpressionFactory.java:52)
	at cucumber.runtime.java8.Java8StepDefinition.createExpression(Java8StepDefinition.java:78)
	at cucumber.runtime.java8.Java8StepDefinition.<init>(Java8StepDefinition.java:70)
	at cucumber.runtime.java8.Java8StepDefinition.create(Java8StepDefinition.java:31)
	at cucumber.api.java8.En.lambda$Given$42(En.java:593)
	at cucumber.runtime.java.JavaBackend.addStepDefinition(JavaBackend.java:165)
	at cucumber.api.java8.En.Given(En.java:592)
	at cucumber.performance.StepDefs.<init>(StepDefs.kt:35)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at org.springframework.beans.BeanUtils.instantiateClass(BeanUtils.java:142)
```

The stepdef is (in Kotlin, so using Kotlin `Int`): 
```
        Given("^the seller has {int} strike(s)$") { strikes: Int ->
            monitorPeriodMockSteps.expectMonitorPeriod(currentSeller.id, strikes)
        }
```

StepDef class uses:
```
@SpringBootTest(classes = [TestConfiguration::class]) // Testconfiguration provides webdriver
class StepDefs(
        val page: Page, // PageObject classes; edited
        val mockSteps: MockSteps, // MockStep classes edited
        val currentSeller: CurrentSeller // Domain object
): En {
```

Versions from pom.xml:
```
        <kotlin.version>1.2.30</kotlin.version>
        <cucumber.version>3.0.0-SNAPSHOT</cucumber.version> // built locally last monday
```

Dependencies:
```
            <dependency>
                <scope>test</scope>
            </dependency>
            <dependency>
                <artifactId>cucumber-java8</artifactId>
                <scope>test</scope>
            </dependency>
            <dependency>
                <artifactId>cucumber-spring</artifactId>
                <scope>test</scope>
            </dependency>
```

Our project is using an internal parent project, which includes `Spring Boot (1.5.10.RELEASE)`

Hoping the stacktrace means anything to you. Not sure I'd be able to provide an example project from this work project... Let me know if there's additional information I could provide.

aslakhellesoy:
@mlvandijk Cucumber Expressions don't use anchors.

The presence of anchors causes Cucumber to try to process this as a regular expression, and fails to do so because it is invalid.

We should probably anticipate this kind of user error and provide a better error message. For example, if we get a `PatternSyntaxException`, and the source contains `{` and `}` we can assume that the user has mixed regexp and cukexp syntax.

mlvandijk:
@aslakhellesoy By anchors you mean the `^` and `$` at beginning and end, right?
So the stepdefinition should be:
```
        Given("the seller has {int} strike(s)") { strikes: Int ->
            monitorPeriodMockSteps.expectMonitorPeriod(currentSeller.id, strikes)
        }
```

aslakhellesoy:
Yep that's right @mlvandijk

mlvandijk:
@aslakhellesoy I knew that, but let's make sure to document! :)

aslakhellesoy:
Better error message added to cucumber-expressions: https://github.com/cucumber/cucumber/commit/ee39f2fe91f747f1207380e7d43eb9163c4f6868

mlvandijk:
Using {byte} is ... interesting:
Having defined the following step definition:
```
    @Given("I have {byte} cucumbers in my belly/stomach")
    public void i_have_byte_cucumbers_in_my_belly(Integer int1) {
        // Write code here that turns the phrase above into concrete actions
        throw new PendingException();
    }
```
And converting between hex and decimal with:
https://www.rapidtables.com/convert/number/decimal-to-hex.html
https://www.rapidtables.com/convert/number/hex-to-decimal.html

**Example 1: The number 127 should be `7F`(a valid value)**
`Given I have 0x7F cucumbers in my belly`
throws
```
cucumber.runtime.CucumberException: Failed to invoke expressions.StepDefs.i_have_byte_cucumbers_in_my_belly(Integer) in file:/Users/maritvandijk/IdeaProjects/cucumber-expression-test/target/test-classes/, caused by java.lang.IllegalArgumentException: argument type mismatch
	at cucumber.runtime.Utils$1.call(Utils.java:29)
	at cucumber.runtime.Timeout.timeout(Timeout.java:16)
	at cucumber.runtime.Utils.invoke(Utils.java:20)
	at cucumber.runtime.java.JavaStepDefinition.execute(JavaStepDefinition.java:54)
	at cucumber.runtime.StepDefinitionMatch.runStep(StepDefinitionMatch.java:46)
	at cucumber.api.TestStep.executeStep(TestStep.java:113)
	at cucumber.api.TestStep.run(TestStep.java:94)
	at cucumber.api.TestCase.run(TestCase.java:58)
	at cucumber.runner.Runner.runPickle(Runner.java:80)
	at cucumber.runtime.Runtime.runFeature(Runtime.java:124)
	at cucumber.runtime.Runtime.run(Runtime.java:109)
	at cucumber.api.cli.Main.run(Main.java:36)
	at cucumber.api.cli.Main.main(Main.java:18)
Caused by: java.lang.IllegalArgumentException: argument type mismatch
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at cucumber.runtime.Utils$1.call(Utils.java:26)
	... 12 more
```
** Example 2: The number `F7` should be number 247 (invalid value)**
`Given I have 0xF7 cucumbers in my belly`
```
cucumber.runtime.CucumberException: Could not convert arguments for Step Definition 'expressions.StepDefs.i_have_byte_cucumbers_in_my_belly(Integer) in file:/Users/maritvandijk/IdeaProjects/cucumber-expression-test/target/test-classes/'
	at cucumber.runtime.StepDefinitionMatch.runStep(StepDefinitionMatch.java:41)
	at cucumber.api.TestStep.executeStep(TestStep.java:113)
	at cucumber.api.TestStep.run(TestStep.java:94)
	at cucumber.api.TestCase.run(TestCase.java:58)
	at cucumber.runner.Runner.runPickle(Runner.java:80)
	at cucumber.runtime.Runtime.runFeature(Runtime.java:124)
	at cucumber.runtime.Runtime.run(Runtime.java:109)
	at cucumber.api.cli.Main.run(Main.java:36)
	at cucumber.api.cli.Main.main(Main.java:18)
Caused by: io.cucumber.cucumberexpressions.CucumberExpressionException: ParameterType {byte} failed to transform [0xF7] to class java.lang.Byte
	at io.cucumber.cucumberexpressions.ParameterType.transform(ParameterType.java:98)
	at io.cucumber.cucumberexpressions.Argument.getValue(Argument.java:63)
	at io.cucumber.stepexpression.ExpressionArgument.getValue(ExpressionArgument.java:15)
	at cucumber.runtime.StepDefinitionMatch.runStep(StepDefinitionMatch.java:37)
	... 8 more
Caused by: java.lang.NumberFormatException: Value 247 out of range from input 0xF7
	at java.lang.Byte.decode(Byte.java:279)
	at io.cucumber.cucumberexpressions.ParameterTypeRegistry$3.apply(ParameterTypeRegistry.java:48)
	at io.cucumber.cucumberexpressions.ParameterTypeRegistry$3.apply(ParameterTypeRegistry.java:45)
	at io.cucumber.cucumberexpressions.SingleTransformer.transform(SingleTransformer.java:22)
	at io.cucumber.cucumberexpressions.ParameterType.transform(ParameterType.java:96)
	... 11 more
```
**Example 3: Value `0x80` or 128 (invalid value)**
`And I have 0x80 cucumbers in my stomach`
=> "Value 128 out of range from input 0x80" (like example 2)
`And I have 0xFF80 cucumbers in my stomach` (this should be the signed version?)
=> suggests the following snippet:
```
@Given("I have {int}xFF{int} cucumbers in my stomach")
public void i_have_xFF_cucumbers_in_my_stomach(Integer int1, Integer int2) {
    // Write code here that turns the phrase above into concrete actions
    throw new PendingException();
}
```

Maybe my test cases / hexadecimals are off, but I'd say this is not easy to use...

mpkorstanje:
I've had to look up how those worked again and the rules downright rather complicated because java treats them as 2s-complement numbers. 

As a normal person I generally expect `0xff` to give me a bit set of `11111111`.  In java you get  `011111111`. A bit annoying. So lets byte the bullet and drop it. The people who want to use it can sink their teeth in it.

aslakhellesoy:
I pushed https://github.com/cucumber/cucumber/commit/ecab2abda99344a690b9e3da3c15d529463d207e where `{byte}` use the same decimal pattern as `{short}`, `{int}` and `{long}`. That's less confusing.

I considered adding a `{hex}` type with pattern `0[xX][0-9a-fA-F]{1,16}` that would resolve to a `Long`. It could then be down-cast to any integer type when the stepdef gets called. But I resisted the temptation since I should add it to the 3 other implementations as well. If people really want to write hex in Gherkin they can write their own bloody stepdefs! 😈

mpkorstanje:
Hallelujah!

mpkorstanje:
>  Explain cucumber.api.Configuration in error message (unless it already exists)

Got the infra structure set up for that. A full blown explanation depending on the existence of the configuration adds too much complexity. As does combining all the messages in a completely sensible way. People can read a stack trace.

mpkorstanje:
Step expression should be completely hidden from the user. They'll only ever see the `cucumber.api.TypeRegistry`, they won't ever see `io.cucumber.stepexpression.TypeRegistry`.

The purpose of the package is to abstract the cucumber expressions and data table types and all the transformations involved from the java and java8 step definitions and the step definition match. 

This means: 
 * Creating a step expresssion from a cucumber expression + optional data table or doc string type
 * Combining the pickle parameters and pickle arguments into a list of method arguments (combining these makes things much simpler)
 * Transforming these arguments into java objects.

This used to be done in [`PickleStepDefinitionMatch.transformedArgs`](https://github.com/cucumber/cucumber-jvm/pull/1248/files#diff-11ef41317dfcc1e599ad71c8bc8345a4L60) and [`ParameterInfo .convert`](https://github.com/cucumber/cucumber-jvm/pull/1248/files#diff-60a46d646428c688433921a71f6c387fL133) neither of which should be concerned with transformations. Now it has been contained.

It also required `LocalizedXStreams` to be passed around everywhere. The `TypeRegistry` goes from the `Runtime` to the `Backend` to the `StepDefinition` and that is it.

mpkorstanje:
>  easier for us to version.

I've automated it a while ago. Break the api, break the build.

https://github.com/cucumber/cucumber-jvm/blob/master/pom.xml#L866

mpkorstanje:
Blog post drafted: https://github.com/cucumber/website/pull/290

mpkorstanje:
@aslakhellesoy  StepdefGenerator doesn't appear to be used by IDEA. Or anybody on Github. You sure it's not dead code?

aslakhellesoy:
Yeah just get rid of StepdefGenerator. It's old dead code.

mpkorstanje:
I'm going to split writing the docs regarding the configuration from this issue. The current docs are talking about a completely different configurations. This will take some work. And while that is being done I'd rather see another PR come in that is going to conflict with everything.

The IDEA people will need a release to work with so that also needs a merge.

PR #1254, opened 2017-10-13
fix:  changed stack trace list from 3 to 2, to get correct step defin…

tts-ll:
…ition file name in reports

## Summary

In "pretty" reports: 

Printing out code locations for step definitions was giving "
NativeConstructorAccessorImpl:-2"
for filename:line_number 
instead of the appropriate "step definition filename":line_number output.
This error only happens with Java8/Lambda step definition functions.

Changing 3 to 2, gets the appropriate stacktrace list item to report file and line number.


mpkorstanje:
Thanks for the bug report. This must have been broken while refactoring java 8. It appears there is no test for it.

PR #1258, opened 2017-10-14
Add junit-jupiter runner

mpkorstanje:
## Summary

See #1103.

The PR has been updated for cucumber 2.0.0 and along with that the Extension will provide a stream of DynamicContainers rather then Dynamic tests. This reflects the feature-scenario hierarchy better.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CO
 ### Added
  * [Core] Optimize MethodScanner ([#1238](https://github.com/cucumber/cucumber-jvm/pull/1236) Łukasz Suski) 
 ### Changed
+ * [Core]  Use gherkin 5.0.0 ([#1252](https://github.com/cucumber/cucumber-jvm/commit/5e305951026a1573ede77e05e86bbe8ed3bca55b) M.P. Korstanje) 
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CO
  * [Core] Correct the name of the Json Formatter embeddings node ([#1236](https://github.com/cucumber/cucumber-jvm/pull/1236) Haroon Sheikh)
  * [Spring] Exception is thrown complaining about multiple matching beans ([#1225](https://github.com/cucumber/cucumber-jvm/pull/1225), [#1226](https://github.com/cucumber/cucumber-jvm/pull/1226), M.P. Korstanje)
 
-## [2.0.1-SNAPSHOT](https://github.com/cucumber/cucumber-jvm/compare/v2.0.0...v2.0.1) (2017-09-17)
+## [2.0.1](https://github.com/cucumber/cucumber-jvm/compare/v2.0.0...v2.0.1) (2017-09-17)
 
 ### Added
  * [Core] cucumber.api.TableConverter interface ([#1223](https://github.com/cucumber/cucumber-jvm/pull/1223) M.P. Korstanje) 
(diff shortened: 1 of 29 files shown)


mpkorstanje:
@tenwit 

Looking at what [RunCukesTest](https://github.com/cucumber/cucumber-jvm/blob/2728abb52a1aff29e84646470042c60f8da20591/examples/java-calculator-jupiter/src/test/java/cucumber/examples/java/calculator/RunCukesTest.java) looks like I am not sure I am happy this this. There appears to be entirely too much boiler plate.

I think we can reduce it by adding a JUnitJupiterCucumber class that contains it so RunCukesTest doesn't have to.

```java
public abstract class JUnitJupiterCucumber {

    @TestFactory
    @ExtendWith(CucumberExtension.class)
    public Stream<DynamicContainer> runAllCucumberScenarios(Stream<DynamicContainer> scenarios) {
        return scenarios;
    }
}

@CucumberOptions(plugin = "json:target/cucumber-report.json")
public class RunCukesTest extends JUnitJupiterCucumber {
}
```

Thoughts?

reinhapa:
@mpkorstanje what would be the additional content within `RunCukesTest` besides the report definition itself? Possibly the way to go would be to create a special Cucumber `org.junit.platform.engine.TestEngine`...

tenwit:
This code was initially written with JUnit 5 M2, then upgraded a few times. JUnit has changed a lot in the mean time :) If I was to start from scratch, I'd do things quite differently.
I'd probably continue to build on Jupiter for the built-in tags and extensions support, but I think that ideally, Cucumber would have its own engine, working off feature files rather than fixtures. It's just that the cost of re-implementing extensions (e.g. Spring integration) and tags for a new non-Jupiter engine is too high.
Assuming that a CucumberExtension for Jupiter is the way to go, I would try to avoid creating a superclass for test fixtures. I would try to find a way to define a single class with the CucumberOptions annotation, and drive everything off that. With the DynamicNode stuff that's been in for the last couple of milestones, I think this should be possible.

mpkorstanje:
We're definitely not taking advantage of Jupiter right now. Except for the fact that the code now fits in two classes instead of 9. :)

What I'd like to be able to do with Jupiter is off load the DI to whatever other extensions are registered with Jupiter. Say I'm using Spring, Mockito and`cucumber-jupiter`. 

So when I run a feature cucumber asks jupiter to process all my step definitions as if they were test instances. This means that field annotated with `@Autowired` gets injected, everything annotated with `@Mock` gets a mock Injected into it.

Not sure if that is possible.

Art-B:
> This code was initially written with JUnit 5 M2, then upgraded a few times. JUnit has changed a lot in the mean time :) If I was to start from scratch, I'd do things quite differently.

Isn't it better to do things right before all the tooling and IDE integration come along? (I don't write such things, but I am spearheading adoption at work and such things are big sticking points).

reinhapa:
@tenwit I have not found any JUnit 5 extension that could generate dynamic tests which would be a way to avoid the test factory method... Unfortunately I did not found some thing like that..

mpkorstanje:
>  I think that ideally, Cucumber would have its own engine, working off feature files rather than fixtures. It's just that the cost of re-implementing extensions (e.g. Spring integration) and tags for a new non-Jupiter engine is too high.

I had a look at the jupiter test engine. It doesn't appear to be very difficult as the platform provides most of the abstract implementation and somewhat resembles the current infrastructure of the cucumber-junit.

I don't think the costs of reimplementing extensions such as Spring and Mockito is a valid concern.  The concept of extension as used by Jupiter doesn't line up with the way it is used in Cucumber. The 'test' in Jupiter is a class file . While in Cucumber it is a feature file.  Jupiter generates and/or post processes these class files. There is no equivalent with a feature file. And while there is a simile between Glue Classes and Test classes the main execution unit is a feature file so there would be no reasonable way to define which extensions should be used.

So in my opinion writing a proper test engine would be the proper solution.

mpkorstanje:
What I am currently puzzling over is how to tie glue code and feature files together. in the TestEngine.

With Jupiter the platform informs us about the file, directory or classpath that needs to be executed and it can be scanned for features. Though there seems to be no way to specify which glue goes with which feature. This means we can only choose to use all glue and hope to be correct.

bonigarcia:
Is there any plan to publish the `CucumberExtension` for JUnit 5 in the master branch?

PR #1259, opened 2017-10-15
Deprecate Spring context configuration by more than one class

brasmusson:
## Summary

Deprecate Spring context configuration by more than one class

## Details

Print a warning if more than one glue class that attempts to configure the Spring context is found (that is have a `@ContextConfiguration` or `@ContextHierarcy` annotation.

## Motivation and Context

We have decided to only allow one glue class to configure the Spring context, as this is a non-backward compatible change, it will be introduced in v3.0.0 (#1246). Until then a deprecation warning should be printed.

## How Has This Been Tested?

The automated test suite has been updated to verify this behavior.

## Types of changes

## Checklist:

### DIFF
--- a/spring/src/main/java/cucumber/runtime/java/spring/SpringFactory.java
+++ b/spring/src/main/java/cucumber/runtime/java/spring/SpringFactory.java
@@ -62,6 +62,7 @@ public class SpringFactory implements ObjectFactory {
 
     private final Collection<Class<?>> stepClasses = new HashSet<Class<?>>();
     private Class<?> stepClassWithSpringContext = null;
+    private boolean deprecationWarningIssued = false;
 
     public SpringFactory() {
     }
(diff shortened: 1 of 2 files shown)


brasmusson:
Removed all tabs.

tobamaestro:
Hi all. I'm new to Cucumber and I've seen most of the tutorials on Cucumber-Spring-Boot stack recommending that each step class should extend some base SpringBootIntegrationTest class, which configures context. How should this work now when this change is introduced? Is there any material explaining best new approach?

mpkorstanje:
Your step definitions no longer have to extend from that class. You can either pick any one of your step definition classes  and put the spring configuration on that one. Or you can use a dedicated class like 
[CucumberContextConfiguration](/cucumber/cucumber-jvm/blob/master/examples/spring-txn/src/test/java/cucumber/examples/spring/txn/CucumberContextConfiguration.java). As long as the class has at least one step def it will be discovered.

mpkorstanje:
>  Is there any material explaining best new approach?

The example project is your best guess at the moment.

https://github.com/cucumber/cucumber-jvm/tree/master/examples/spring-txn

tobamaestro:
Thanks. I've found it in the meantime and I removed my previous question, but I guess it's good you replied if anyone else wonders about the same thing again.

Issue #1268, opened 2017-10-25
Document supported JUnitAnnotations


mpkorstanje:
The only supported annotations are `@ClassRule`, `@BeforeClass` and `@AfterClass` in the runner class. They are invoked around the execution of the selected suite of feature files. This is not documented in [Cucumber.java](https://github.com/cucumber/cucumber-jvm/blob/master/junit/src/main/java/cucumber/api/junit/Cucumber.java). In fact it documents the exact opposite.

```java
 * Classes annotated with {@code @RunWith(Cucumber.class)} will run a Cucumber Feature.
 * The class should be empty without any fields or methods.
```

I would however discourage the use of these annotations. They are not portable between runners and should not be relied upon. If for example the commandline runner is used, the JUnit rules will not be invoked. This may lead to a different test execution.

jhorstmann:
> If for example the commandline runner is used, the JUnit rules will not be invoked

They also won't be used by some IDE integrations, which will be very confusing. For example IntelliJ cucumber plugin internally uses the commandline runner instead of JUnit.

PR #1272, opened 2017-10-26
Add supported JUnit annotations to javadoc

mlvandijk:
## Summary
Add supported JUnit annotations to javadoc

## Details
The only supported annotations are @ClassRule, @BeforeClass and @AfterClass in the runner class. This should be mentioned in Cucumber. This is not documented in [Cucumber.java](https://github.com/cucumber/cucumber-jvm/blob/master/junit/src/main/java/cucumber/api/junit/Cucumber.java).

## Motivation and Context
As suggested in #1268

## Types of changes

## Checklist:

### DIFF
--- a/junit/src/main/java/cucumber/api/junit/Cucumber.java
+++ b/junit/src/main/java/cucumber/api/junit/Cucumber.java
@@ -15,6 +15,9 @@ import cucumber.runtime.junit.FeatureRunner;
 import cucumber.runtime.junit.JUnitOptions;
 import cucumber.runtime.junit.JUnitReporter;
 import cucumber.runtime.model.CucumberFeature;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
 import org.junit.runner.Description;
 import org.junit.runner.notification.RunNotifier;
 import org.junit.runners.ParentRunner;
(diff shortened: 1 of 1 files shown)


junit/src/main/java/cucumber/api/junit/Cucumber.java
```diff
@@ -28,7 +31,8 @@
 /**
  * <p>
  * Classes annotated with {@code @RunWith(Cucumber.class)} will run a Cucumber Feature.
- * The class should be empty without any fields or methods.
+ * In general, the runner class should be empty without any fields or methods.
```
mpkorstanje: I think it might be better to replace this with a code sample.

junit/src/main/java/cucumber/api/junit/Cucumber.java
```diff
@@ -28,7 +31,8 @@
...
+ * However, it is possible to use the following JUnit annotations in the runner class: {@link ClassRule}, {@link BeforeClass} and {@link AfterClass}.
```
mpkorstanje: I think this should be rephrased to: "Cucumber supports the {@link ClassRule}, {@link BeforeClass} and {@link AfterClass} annotations." and moved to the end of the java doc.


In this instance I think it should be "should not"

mpkorstanje:
@mlvandijk I think you may have to escape some characters here. 

```
[ERROR] /home/travis/build/cucumber/cucumber-jvm/junit/src/main/java/cucumber/api/junit/Cucumber.java:37: error: unknown tag: RunWith

[ERROR] * @RunWith(Cucumber.class)

[ERROR] ^

[ERROR] /home/travis/build/cucumber/cucumber-jvm/junit/src/main/java/cucumber/api/junit/Cucumber.java:38: error: unknown tag: CucumberOptions

[ERROR] * @CucumberOptions(plugin = "pretty")

[ERROR] ^
```

Issue #147, opened 2012-01-17
Add verbose output


aslakhellesoy:
From CLI: `--verbose`

From JUnit: `@Opts(verbose=true)`

This would make it easier to diagnose problems like #146 and #1234.

mpkorstanje:
Most common problems seem to come from misconfigures paths. So logging which feature files and which glue files are found would be a good first step.

Logging what the DI container is doing would be a good second step.

mpkorstanje:
I guess logging which back ends was loaded also helps

mpkorstanje:
We can either log directly to standard out or we can use a logger framework. I would prefer the latter because it will work with existing frameworks and keep us from having to re-invent the wheel when it comes to log levels, parameter interpolation, ect, ect.

We are already using slf4j as a logging facade. Unfortunately it doesn't actually log. This is left to actual logger frameworks like log4j and logback. We can't use logback or  log4j because `cucumber-core` is library and not an application. Our choice of logging framework will conflict with theirs. 

A reasonable way to solve this would be to move `cucumber.api.cli.Main` and the option parsing to its own module and then set the log level programatically depending on the value of `--verbose`. This also requires that we untangle the option parsing from the actual runtime options and from the junit module.

dkowis:
Another way to handle it, if we want to forcibly include a logging output framework, is to shade log4j2 into the cucumber namespace. 

SLF4j complains each time an API is not included to standard out, so it does be annoying, but you have to be watching for it.

mpkorstanje:
Shading won't solve the right problem. When using slf4j there can only be one slf4j binding. 

https://www.slf4j.org/codes.html#multiple_bindings
> SLF4J API is designed to bind with one and only one underlying logging framework at a time. If more than one binding is present on the class path, SLF4J will emit a warning, listing the location of those bindings.
> ...
> Note The warning emitted by SLF4J is just that, a warning. Even when multiple bindings are present, SLF4J will pick one logging framework/implementation and bind with it. The way SLF4J picks a binding is determined by the JVM and for all practical purposes should be considered random. As of version 1.6.6, SLF4J will name the framework/implementation class it is actually bound to.
> ...
[...]

PR #1273, opened 2017-10-26
Print JUnit Options when unknown option is provided

mlvandijk:
## Summary
Print available JUnit Options when an unknown option is provided

## Details
Only a few JUnit options can be provided in the command line arguments or the CucumberOptions. For any other options a CucumberException is thrown stating that an "Unknown option" is used.

## Motivation and Context
Found while testing PR for #147 

## How Has This Been Tested?
Running cucumber-java-skeleton project with all known and several unknown options.

## Types of changes

## Checklist:


mpkorstanje:
This is not needed when cucumber is ran from the commandline. The commandline won't use junit. But I guess this makes sense when `mvn test -Dcucumber.options` is used.

PR #1276, opened 2017-11-10
- Update android compile and target sdk version from 21 to 26.

ahmed-adel-said:
- Update espresso core library and replace android testing support library to android runner.
- Update gradle version from 1 to 3 and also update distribution url of gradle from 2.2.1 to 4.1.

## Types of changes

## Checklist:

### DIFF
--- a/examples/android/android-studio/Cukeulator/app/build.gradle
+++ b/examples/android/android-studio/Cukeulator/app/build.gradle
@@ -1,13 +1,13 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 21
-    buildToolsVersion "21.1.2"
+    compileSdkVersion 26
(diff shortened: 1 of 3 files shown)


mpkorstanje:
I see that we're using sdk version 21 in travis.yml and I can also see that the android studio example example is not covered by travis.

All in all I lack the knowledge to judge this pull request properly.

lsuski:
Everything looks correct except that there should be `androidTestImplementation`  instead of `androidTestCompile` which is deprecated, I'm using the same versions in couple of projects and everything works

ahmed-adel-said:
@lsuski Agree 👍

mpkorstanje:
@ahmed-adel-said  coveralls is a bot. :)

I reckon it would be prudent to ensure Travis can execute the example before merging this PR.

PR #1287, opened 2017-11-30
[JUnit] Remove Cucumber.createRuntime method

mpkorstanje:
## Summary

Neither the runtime nor the backend or any of the classes involved in
their construction are part of the public API. As such they should not
be  exposed.

The recommended way to observe the cucumber process is to listen to
events by using a plugin. For example the JSONFormatter.

## Types of changes


cliviu:
What is the recommended way to register a plugin ?

PR #1288, opened 2017-12-01
Add feature tags in the JSON output

pgentile:
## Summary

Add feature tags in the JSON result of the JSON Formatter

## Details

New JSON field named "tags" in feature section

## Motivation and Context

Cucumber 1.x JSON reports contain feature tags.
We use feature tags to keep our features organized.

## How Has This Been Tested?

Updated JSON Formatter unit tests

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/runtime/formatter/JSONFormatter.java
+++ b/core/src/main/java/cucumber/runtime/formatter/JSONFormatter.java
@@ -170,6 +170,8 @@ final class JSONFormatter implements Formatter {
             featureMap.put("description", feature.getDescription() != null ? feature.getDescription() : "");
             featureMap.put("line", feature.getLocation().getLine());
             featureMap.put("id", TestSourcesModel.convertToId(feature.getName()));
+            featureMap.put("tags", feature.getTags());
+
         }
         return featureMap;
     }
--- a/core/src/test/java/cucumber/runtime/formatter/JSONFormatterTest.java
+++ b/core/src/test/java/cucumber/runtime/formatter/JSONFormatterTest.java
@@ -91,7 +91,8 @@ public class JSONFormatterTest {
                 "          }\n" +
                 "        ]\n" +
                 "      }\n" +
-                "    ]\n" +
+                "    ],\n" +
+                "    \"tags\": []\n" +
                 "  }\n" +
                 "]";
         assertPrettyJsonEquals(expected, formatterOutput);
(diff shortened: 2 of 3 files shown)


aslakhellesoy:
The JSON formatter will be deprecated and eventually removed. It will be replaced by the new event protocol.

This change seems innocuous enough, so happy to include it.

Just keep in mind it might not be around in a future release.

mpkorstanje:
Unfortunately this is a breaking change. Consumers of the json output expect a certain schema. Adding a node will cause strict parsers to fail. 

We can line this up for v3 though. If you have need of this immediately I would recommended duplication the json formatter.

@aslakhellesoy versioning the json formatter would also help.

pgentile:
Hi @mpkorstanje 

I have added the `tags` on the feature node because it was present on JSON reports generated by Cucumber 1.x.

I think that consumers of the JSON reports should ignore unknown fields, instead of throwing an error.

mpkorstanje:
In that case it is indeed a bug.

> I think that consumers of the JSON reports should ignore unknown fields, instead of throwing an error.

We should try to follow the [Robustness principle](https://en.wikipedia.org/wiki/Robustness_principle). So while it might be prudent for consumers to ignore unknown fields we should adhere to our schema and  not any.

aslakhellesoy:
There is no official schema (JSON Schema) for the JSON formatter. Java/Ruby/JavaScript Cucumbers all produce a *slightly* different JSON. I doubt adding this would cause failures anywhere, and I don’t think it warrants a major version bump.

mpkorstanje:
Seeing how the regression was introduced in 2.0.0 it is a moot point now.

aslakhellesoy:
Hi @pgentile,

Aslak Hellesøy
Creator of Cucumber

brasmusson:
The tags of the feature is included in the json output (in v2.0.0-v2.1.0), by the fact that each test-case/scenario in the json output includes all tags that applies to them, regardless of whether that come from the feature or scenario (for test cases from gherkin scenarios), or from the feature, scenario outline or examples table (for test cases from example rows in the gherkin source). In v1.2.x the feature tags was only included on the test cases from the example rows, not the test cases from the scenarios, but that was fixed in v2.0.0.

I might have very deliberately removed the tags from the features in the json output. I mean neither scenario outlines nor example tables (which can have tags in the gherkin source) are represented in the json output, and I think really interesting property is what tag are applicable to each of the test cases (which is the unit of execution).

PR #1289, opened 2017-12-06
cache matched steps definitions in RuntimeGlue

lsuski:
## Summary

This is a performance improvement in `RuntimeGlue.stepDefinitionMatches` method which is very slow on Android < 8.0 and with large test base

## Details

`stepDefinitionMatches` iterates through all steps definitions to find all definitions that match current `PickleStep` (by testing if step text matches regular expression of step definition) - lots of steps doesn't have arguments or have only table argument or are used with same arguments - this implies that matching `StepDefinition` for them can be safely cached to avoid repeating of iteration through all definitions and receiving the same results.
I've added simple cache where step text is a key and `StepDefinition` and `List<Argument>` are value. When a step with the same text is being processed then its definition is taken from cache.

E.g. in my project I have 5730 steps invocations and 813 steps definitions which normally causes 4 658 490 regular expression checks.  After adding cache 4218 invocations was processed by cache which saves lots of time.

## Motivation and Context

Cucumber should perform as quickly as possible and dryRun should complete very fast because it executes only Cucumber code. In fact on my project (653 executed unique scenarios) dryRun on Nexus 7 Android 6.0 takes 5 minutes and on Emulator Android 8.0 it takes 40 s.

After adding cache  dryRun on Nexus 7 takes 2 mins and on Emulator it takes 20 s

## How Has This Been Tested?

I've tested it on my project with 653 scenarios.
I've added unit test for `RuntimeGlue.stepDefinitionMatches`.

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/runtime/RuntimeGlue.java
+++ b/core/src/main/java/cucumber/runtime/RuntimeGlue.java
@@ -74,16 +74,14 @@ public class RuntimeGlue implements Glue {
         }
         if (matches.size() == 1) {
             StepDefinitionMatch match = matches.get(0);
-            matchedStepDefinitionsCache.put(stepText, new CacheEntry(match.getStepDefinition(),match.getArguments()));
+            matchedStepDefinitionsCache.put(stepText, new CacheEntry(match.getStepDefinition(), match.getArguments()));
             return match;
-        } else {
-            throw new AmbiguousStepDefinitionsException(step, matches);
         }
+        throw new AmbiguousStepDefinitionsException(step, matches);
     }
 
     private List<StepDefinitionMatch> stepDefinitionMatches(String featurePath, PickleStep step) {
         List<StepDefinitionMatch> result = new ArrayList<StepDefinitionMatch>();
-
         for (StepDefinition stepDefinition : stepDefinitionsByPattern.values()) {
             List<Argument> arguments = stepDefinition.matchedArguments(step);
             if (arguments != null) {
@@ -109,9 +107,9 @@ public class RuntimeGlue implements Glue {
 
     private void removeScenarioScopedHooks(List<HookDefinition> beforeHooks1) {
         Iterator<HookDefinition> hookIterator = beforeHooks1.iterator();
-        while(hookIterator.hasNext()) {
+        while (hookIterator.hasNext()) {
             HookDefinition hook = hookIterator.next();
-            if(hook.isScenarioScoped()) {
+            if (hook.isScenarioScoped()) {
                 hookIterator.remove();
             }
         }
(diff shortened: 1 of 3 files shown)


core/src/main/java/cucumber/runtime/RuntimeGlue.java
```diff
@@ -73,9 +74,26 @@ public StepDefinitionMatch stepDefinitionMatch(String featurePath, PickleStep st
...
             List<Argument> arguments = stepDefinition.matchedArguments(step);
             if (arguments != null) {
+                if (notFound) {
+                    matchedStepDefinitionsCache.put(stepText, new CacheEntry(stepDefinition, arguments));
+                    notFound = false;
+                }
+                else {
+                    //duplicate step - remove from cache
+                    matchedStepDefinitionsCache.remove(stepText);
+                }
```
mpkorstanje: So rather then invalidating the cache entry when dealing with a duplicate step it might be better to store a list of cache entries. When upon fetching the list from the cache, the list contains multiple entries we can throw the `AmbiguousStepDefinitionsException`.

lsuski:
I implemented it in this way at first time and it was simpler but then realized that in most cases (in properly written tests 100% of time)  it will be one element list so it is waste of memory. If this is ok I'll revert to that version

mpkorstanje:
Good point. I think it should be possible to throw the `AmbiguousStepDefinitionsException` between creating and storing the cache entries.

lsuski:
`AmbiguousStepDefinitionsException` should contain information about all ambiguous steps so I don't think we can throw it in the middle of the loop, or can we?

core/src/test/java/cucumber/runtime/RuntimeGlueTest.java
```diff
@@ -60,4 +78,63 @@ public void removes_glue_that_is_scenario_scoped() {
         assertEquals(0, glue.beforeHooks.size());
         assertEquals(0, glue.afterHooks.size());
     }
+
+    @Test
+    public void step_definition_match() {
```
mpkorstanje: This should be a name that describes what the test is testing. This may require splitting up this test case into separate smaller test cases.

core/src/main/java/cucumber/runtime/RuntimeGlue.java
```diff
@@ -60,19 +61,29 @@ public void addAfterHook(HookDefinition hookDefinition) {
...
         if (matches.isEmpty()) {
             return null;
         }
         if (matches.size() == 1) {
-            return matches.get(0);
+            StepDefinitionMatch match = matches.get(0);
+            matchedStepDefinitionsCache.put(stepText, new CacheEntry(match.getStepDefinition(),match.getArguments()));
+            return match;
         } else {
             throw new AmbiguousStepDefinitionsException(step, matches);
```
mpkorstanje: To stay consistent with the rest of this method it would be preferable to remove the else clause and simply throw the exception as the last statement in the method.

core/src/main/java/cucumber/runtime/RuntimeGlue.java
```diff
@@ -60,19 +61,29 @@ public void addAfterHook(HookDefinition hookDefinition) {
...
         }
     }
 
     private List<StepDefinitionMatch> stepDefinitionMatches(String featurePath, PickleStep step) {
         List<StepDefinitionMatch> result = new ArrayList<StepDefinitionMatch>();
+
```
mpkorstanje: Please remove.

mpkorstanje:
Thanks again for optimizing!

Issue #1297, opened 2017-12-11
Document use of DataTables with domain objects


mlvandijk:
## Summary
Had trouble using DataTable, due to small detail

## Expected Behavior
DataTables are easy to use

## Current Behavior
Variable names of the domain object have to match the column names of the data table in the feature file. 

## Possible Solution

Add detail to docs and javadoc

PR #1298, opened 2017-12-11
Add detail to DataTable hint in JavaSnippet and update test

mlvandijk:
## Summary
Add detail to DataTable hint in JavaSnippet to mention that variable names of domain object ("YourType") need to match column names in the feature file.

## Motivation and Context
First use of DataTable didn't quite work as expected (see #1297).
This hint may be helpful to others.

## How Has This Been Tested?
Unit test for JavaSnippet was updated accordingly.
'mvn clean install' built successfully.

## Types of changes

## Checklist:

### DIFF
--- a/java/src/main/java/cucumber/runtime/java/AbstractJavaSnippet.java
+++ b/java/src/main/java/cucumber/runtime/java/AbstractJavaSnippet.java
@@ -24,7 +24,9 @@ public abstract class AbstractJavaSnippet implements Snippet {
     public String tableHint() {
         return "    // For automatic transformation, change DataTable to one of\n" +
                 "    // List<YourType>, List<List<E>>, List<Map<K,V>> or Map<K,V>.\n" +
-                "    // E,K,V must be a scalar (String, Integer, Date, enum etc)\n";
+                "    // E,K,V must be a scalar (String, Integer, Date, enum etc)\n" +
+                "    // Make sure the variable names for YourType match the column names\n" +
+                "    // in your feature file.\n";
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


java/src/main/java/cucumber/runtime/java/AbstractJavaSnippet.java
```diff
@@ -24,7 +24,9 @@ public String arguments(List<Class<?>> argumentTypes) {
     public String tableHint() {
         return "    // For automatic transformation, change DataTable to one of\n" +
                 "    // List<YourType>, List<List<E>>, List<Map<K,V>> or Map<K,V>.\n" +
-                "    // E,K,V must be a scalar (String, Integer, Date, enum etc)\n";
+                "    // E,K,V must be a scalar (String, Integer, Date, enum etc)\n" +
+                "    // Make sure the variable names for YourType match the column names\n" +
+                "    // in your feature file.\n";
```
mpkorstanje: I think 'The field names of YourType must match the column names' is better.

Field names instead of variables. The latter exists only in a function. The former is part of a type.

And must instead of make sure because of the rfc.

Not sure about the subject and object. Swapping might be better (if only someone told me learning grammar would allow me to discuss the use of language :) ).

mlvandijk: Updated. Hope this is better.

java/src/main/java/cucumber/runtime/java/AbstractJavaSnippet.java
```diff
@@ -24,7 +24,9 @@ public String arguments(List<Class<?>> argumentTypes) {
     public String tableHint() {
         return "    // For automatic transformation, change DataTable to one of\n" +
                 "    // List<YourType>, List<List<E>>, List<Map<K,V>> or Map<K,V>.\n" +
-                "    // E,K,V must be a scalar (String, Integer, Date, enum etc)\n";
+                "    // E,K,V must be a scalar (String, Integer, Date, enum etc)\n" +
+                "    // Field names for YourType must match the column names in \n" +
```
mpkorstanje: I found some code that suggests the field name don't need to be an exact match.

`My Field`, `MyField`, `myField` will all match `myField`.

mpkorstanje: The PascalCaseStringConverter btw.

mpkorstanje: And also CamelCaseStringConverter.

So `my Field` will probably also work.

mlvandijk: Thanks, I'll have a look and rewrite a little to make this clear.

PR #1306, opened 2017-12-23
Add readme to cucumber-junit

mpkorstanje:
## Summary

Add readme to cucumber-junit

### DIFF
--- a/junit/README.md
+++ b/junit/README.md
@@ -1,0 +1,54 @@
+Cucumber JUnit 
+==============
+
+Use JUnit to execute cucumber scenarios.
+
+Add the `cucumber-junit` dependency to your pom.
+
+```xml
+<dependencies>
+  [...]
+    <dependency>
+        <groupId>io.cucumber</groupId>
+        <artifactId>cucumber-junit</artifactId>
+        <version>${cucumber.version}</version>
+        <scope>test</scope>
+    </dependency>
+  [...]
+</dependencies>
+```
+
+Create an empty class that uses the Cucumber JUnit runner.
+
+```java
+package com.example;
+
+import cucumber.api.CucumberOptions;
+import cucumber.api.junit.Cucumber;
+import org.junit.runner.RunWith;
+
+@RunWith(Cucumber.class)
+@CucumberOptions(plugin = "json:target/cucumber-report.json")
+public class RunCukesTest {
+}
+```
+
+This will execute all scenarios in same package as the runner, by default glue code is also assumed to be in the same 
+package. The `@CucumberOptions` can be used to provide
+[additional configuration](https://cucumber.io/docs/reference/jvm#list-configuration-options) to the runner. 
+
+
+## Using JUnit Rules ##
+
+Cucumber supports JUnits `@ClassRule`, `@BeforeClass` and `@AfterClass` annotations. These will executed before and 
+after all scenarios. Using these is not recommended as it limits the portability between different runners; they may not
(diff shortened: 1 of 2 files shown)


junit/README.md
```diff
@@ -0,0 +1,50 @@
...
+import cucumber.api.junit.Cucumber;
+import org.junit.runner.RunWith;
+
+@RunWith(Cucumber.class)
+@CucumberOptions(plugin = "json:target/cucumber-report.json")
+public class RunCukesTest {
+}
+```
+
+This will execute all scenarios in the package of the runner. By default glue code is assumed to be in the same
```
mlvandijk: [text edit suggestion] "This will execute all scenarios in the *same* package *as* the runner; by default glue code is assumed to be in the same package."

mlvandijk: [suggestion] Add an example of how to use the '@CucumberOptions'?

junit/README.md
```diff
@@ -0,0 +1,50 @@
...
+public class RunCukesTest {
+}
+```
+
+This will execute all scenarios in the package of the runner. By default glue code is assumed to be in the same 
+package. The `@CucumberOptions` can be used to provide additional configuration to the runner.
+
+## Using JUnit Rules ##
+
+Cucumber supports JUnits `@ClassRule`, `@BeforeClass` and `@AfterClass` annotations. These will be invoked around the
```
mlvandijk: This paragraph might not be clear Cucumber users who are not full-on software developers (yet). 

Particularly:

"Cucumber supports JUnits `@ClassRule`, `@BeforeClass` and `@AfterClass` annotations. These will be invoked around the 
+suite of features. " => what does it mean to run 'around the suite of features'? Is this like a BeforeAll/AfterAll?

"Using these is not recommended as it limits the portability between different runners. " => what is 'portability between different runners'? Why is limiting this a bad thing?

"Instead it is
+recommended to use Cucumbers `Before` and `After` hooks to setup scaffolding." => How to use 'Before' and 'After' hooks? What is 'setup scaffolding'?

Maybe this is not the place to answer all of these questions, but I think this could be clarified a little to be clearer to a wider audience.

junit/src/main/java/cucumber/api/junit/Cucumber.java
```diff
@@ -48,7 +48,8 @@
  * Additional hints can be given to Cucumber by annotating the class with {@link CucumberOptions}.
  * <p>
  * Cucumber also supports JUnits {@link ClassRule}, {@link BeforeClass} and {@link AfterClass} annotations.
- * These will be invoked around the suite of features" and moved to the end of the java doc.
+ * These will be invoked around the suite of features.  Using these is not recommended as it limits the portability
```
mlvandijk: See above. If any changes are made based on those comments, please update accordingly also here.

junit/README.md
mpkorstanje: Do we have some documentation about cucumber options to link here?

junit/README.md
mpkorstanje: Do we have some documentation about the Before and After hooks? I am assuming people are familiar with cucumber when they read this.

mlvandijk: Yes, we do. Currently here: https://cucumber.netlify.com/hooks/

junit/README.md
mlvandijk: Not atm. Fyi, it's just a suggestion. Not blocking at all.

Issue #1303, opened 2017-12-17
Reports does not work 


MoaadAmer:
https://stackoverflow.com/questions/47817891/cucumber-generates-blank-html-report

Cucumber version 1.2.5

brasmusson:
Which version of Cucumber-JVM are you using?

MoaadAmer:
@brasmusson  Cucumber version 1.2.5

haroon-sheikh:
@MoaadAmer Would you be able to share the failing example (maybe on gist.github.com)? I'm not able reproduce the issue on 1.2.5 or latest 2.3.1.

brasmusson:
@haroon-sheikh Do you have a `Before` hook that calls `write` on the `Scenario` argument of the hook? To me it looks like that is triggering the problem.

MoaadAmer:
@brasmusson yes i do have a before hook with argument of type Scenario

brasmusson:
Calling `write` on the `Scenario` argument of a hook will add a `formatter.write` call in the `report.js` file. The javascripts functions of https://github.com/cucumber/cucumber-html will display the content of the `formatter.write` call below the preceding step. For before hooks of the first scenario executed, there does not exist any preceding step, since no (gherkin) step has been executed yet, therefore the report will be blank from this point on (for `formatter.write` calls from before hooks of subsequent scenarios, the content will be displayed below the last step of the preceding scenario, which is incorrect but the report will at least not be blank).

To fix this the https://github.com/cucumber/cucumber-html project need to be updated. However, since the plan is to replace that project, see https://github.com/cucumber/cucumber-html/pull/44, I'm not sure anyone in the core team will take on that work.

The workaround is to not call `write` on the `Scenario` argument of before hooks, if the html-formatter is used.

MoaadAmer:
Okay Thanks !

brasmusson:
The issue exist in [cucumber-html](https://github.com/cucumber/cucumber-html) v0.2.6, which is the latest released version of cucumber-html.

MoaadAmer:
So that defect is also in cucumber 2.3.1 ?

brasmusson:
The issue will exist in all Cucumber-JVM versions until new version of cucumber-html is released with a fix.

PR #1323, opened 2018-02-14
[Core] Add Before and AfterStep hooks

coding-yogi:
This PR adds a new feature of AfterStep. Methods annotated with AfterStep will be executed after every step. For HTML Reports to work properly after this change , this [PR](https://github.com/cucumber/cucumber-html/pull/48) requires to be merged

## Motivation and Context

This change was requested by users of cucumber-jvm

## Types of changes

## Checklist:


core/src/main/java/cucumber/runner/Runner.java
```diff
@@ -128,6 +128,9 @@ private void addTestStepsForPickleSteps(List<TestStep> testSteps, PickleEvent pi
                 match = new FailedStepInstantiationMatch(pickleEvent.uri, step, t);
             }
             testSteps.add(new PickleTestStep(pickleEvent.uri, step, match));
+            if (!runtimeOptions.isDryRun()) {
+                addTestStepsForAfterStepHooks(testSteps, pickleEvent.pickle.getTags());
+            }
```
mpkorstanje: Could you provide a test that shows the after hook is executed after each step?

mpkorstanje: This also affects the  `JSONFormatter`. Currently it assumes all hooks are either before or after hooks. Using after step hooks will have undefined results.

mpkorstanje: @brasmusson where do you think the `TestStepStarted` events for step hooks should go in the json? I am thinking they should go into the same list as the the test steps themselves.

mpkorstanje: I can not. It does indeed all seem to work out.

mpkorstanje: What happens when a step fails? Are the after hooks of the following steps skipped?

coding-yogi: No, right now they aren't skipped. But makes sense to skip them if step is skipped

coding-yogi: Issue is hooks are marked Unskipable

mpkorstanje: The code emulates invoke-around semantics but doesn't fully implement the concept.

coding-yogi: what do you mean by that?

coding-yogi: what part of the implementation is missing?

mpkorstanje: Right now cucumber represents execution as a series of steps. For completeness I'll assume there is also a before step hook. It might look like this:

```
before hook
beforeStep hook
step1
afterStep hook
beforeStep hook
step2
afterStep hook
after hook
```

What it should try to achieve is:

```js
executeScenario(){
    executeBeforeHooks()
    executeSteps()
    executeAfterHooks()
}

executeSteps(){
    skipNextStep = false;

    for step in steps
      skip = skipNextStep
      beforeStepHook(skip)
      skipNextStep = step(skip)
      afterStepHook(skip)
      skip = skipNextStep
}
```

What it currently does in TestCase:
```
executeSteps(){
    skipNextStep = false;

    for step in steps
      skipNextStep = step(skipNextStep)
}
```

So not only should after step hooks be skipabble, they should also only be skipped when their step was skipped.

coding-yogi: I got your point. Reason for not implementing BeforeStep is I do not see any use case for the same. If you see previous step-hook requests, they all requested for AfterStep hook which I also feel is useful in many scenarios.
I do not have an issue implementing BeforeStep, my only point is it should not be implemented just because we are implementing AfterStep

Coming to skipable hook, currently all Hooks are of class *UnskipableStep*, if I workaround that, it defeats the purpose of UnskipableStep class. Other option is to have class called *SkipableHookStep* for AfterStep (and if required, BeforeStep) hooks.

The logic you mentioned for skipping has a catch, Any failed step will set skipNextStep to true, But failed step should not skip its AfterStep. AfterStep should only be skipped for skipped Steps, not failed Step. I can see how that can be done.

Any thoughts on having a new Class *SkipableHookStep* for AfterStep/BeforeStep?

coding-yogi: Another option is to rename class **UnskipableStep** to **HookStep** and then internally handle whether to execute hook step or not in **execute()** method

brasmusson: Cucumber-Ruby does add AfterStep hooks as ordinary steps, so the AfterStep hooks for failed steps are skipped. This may not be ideal, but on the other hand inconsistency between Cucumber implementation is not ideal either, so we should try to get a common vision on where all the Cucumber implementation are heading.

brasmusson: >  I can see afterstep being reported properly in json report

I do not see any test case that specify how after steps hooks are reported in the json report, so I think such a comment is premature.

In Cucumber-Ruby the AfterStep hooks are included in the json report in the "after" entry of each step, see https://github.com/cucumber/cucumber-ruby/blob/master/spec/cucumber/formatter/json_spec.rb#L671-L683

coding-yogi: I will update the class to HookStep instead of Unskipable step and handle the logic of skipping AfterStep for a failed/skipped step. This will be in sync with ruby.

Let me know if that should be ok

coding-yogi: I will have to do some changes for JSONFormatter as there is some incosistency when compared with Ruby Json report. Will add a test for same

coding-yogi: @mpkorstanje , @brasmusson , I have implemented logic of skipping AfterStep in case of failed or skipped Step and also tests for JSON formatter showing AfterSteps JSON arry node should belong to it's respective step node.

Would you mind reviewing the changes?

brasmusson: I'll look into them, but it will probably take half a week or so before I have the time and energy to do it.

coding-yogi: Sure, Thanks

core/src/main/java/cucumber/runner/HookStep.java
```diff
@@ -10,17 +10,26 @@
...
-    public UnskipableStep(HookType hookType, DefinitionMatch definitionMatch) {
+    public HookStep(HookType hookType, DefinitionMatch definitionMatch) {
         super(definitionMatch);
         this.hookType = hookType;
     }
 
     protected Result.Type executeStep(String language, Scenario scenario, boolean skipSteps) throws Throwable {
-        definitionMatch.runStep(language, scenario);
-        return Result.Type.PASSED;
+        if(hookType == HookType.After || hookType == HookType.Before) {
```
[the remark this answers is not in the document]
coding-yogi: Sure, I will fix these throughout PR

core/src/test/java/cucumber/runtime/RuntimeTest.java
```diff
@@ -1,10 +1,6 @@
 package cucumber.runtime;
 
-import cucumber.api.PendingException;
-import cucumber.api.Result;
-import cucumber.api.Scenario;
-import cucumber.api.StepDefinitionReporter;
-import cucumber.api.TestCase;
+import cucumber.api.*;
```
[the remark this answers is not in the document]
coding-yogi: Yes, will fix this one. I think it happened due to Intellij formatting it to .* in case all the classes are being used and imported from package

core/src/test/java/cucumber/runtime/formatter/JSONFormatterTest.java
```diff
@@ -557,6 +557,121 @@ public void should_format_scenario_with_hooks() throws Throwable {
...
+            "        \"type\": \"scenario\",\n" +
+            "        \"keyword\": \"Scenario\",\n" +
+            "        \"steps\": [\n" +
+            "          {\n" +
+            "            \"result\": {\n" +
+            "              \"duration\": 1000000,\n" +
+            "              \"status\": \"passed\"\n" +
+            "            },\n" +
+            "            \"line\": 4,\n" +
+            "            \"afterstep\": [\n" +
```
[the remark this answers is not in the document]
coding-yogi: yes, I saw that in ruby but there is a small catch here. In the code every where I have used string representation of HookType enum , for eg `mapToAddTo.put(hookType.toString(), new ArrayList<Map<String, Object>>());`
So in this case I will have to explicitly handle the AfterStep case to report it as After

core/src/test/java/cucumber/runtime/HookTest.java
```diff
@@ -50,4 +52,30 @@ public void after_hooks_execute_before_objects_are_disposed() throws Throwable {
...
+        PickleStep step1 = mock(PickleStep.class);
+        PickleStep step2 = mock(PickleStep.class);
+        PickleEvent pickleEvent = new PickleEvent("uri", new Pickle("name", ENGLISH, asList(step1, step2), Collections.<PickleTag>emptyList(), asList(mock(PickleLocation.class))));
+
+        runner.runPickle(pickleEvent);
+
+        InOrder inOrder = inOrder(step1, step2, hook, backend);
+        inOrder.verify(step1).getText();
+        inOrder.verify(step2).getText();
+        inOrder.verify(hook, times(2)).execute(Matchers.<Scenario>any());
```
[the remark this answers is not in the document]
coding-yogi: This test was commented and was supposed to be removed as this behaviour is verified in other test. I will check once again and remove this test

core/src/test/java/cucumber/runtime/formatter/HTMLFormatterTest.java
```diff
@@ -439,6 +439,67 @@ public void should_handle_after_hooks() throws Throwable {
...
+                "  \"keyword\": \"Given \",\n" +
+                "  \"name\": \"first step\"\n" +
+                "});\n", "" +
+                "formatter.match({\n" +
+                "  \"location\": \"path/step_definitions.java:3\"\n" +
+                "});\n", "" +
+                "formatter.result({\n" +
+                "  \"status\": \"passed\"\n" +
+                "});\n", "" +
+                "formatter.afterstep({\n" +
```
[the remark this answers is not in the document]
coding-yogi: I didn't exactly get what needs to be done for this, should I leave it as it is for now?

coding-yogi:
@brasmusson , I have made the requested changes except the HTML formatter one for which I didn't exactly get what needs to be done

core/src/test/java/cucumber/runtime/formatter/HTMLFormatterTest.java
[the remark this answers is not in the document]
brasmusson: I'll try to take a look at what can best be done with respect to the HTML formatter.

brasmusson:
Even thought there has not as wide participation as I would have liked in the discussion in the #committers channel, the conclusion seems to be that it is reasonable to run the after step hook for a (gherkin) step that fails (but not the after step hooks for subsequent steps). Therefore I took a stab at implementing this (f3c908c). The `TestCase` abstraction is still a (flat) sequence of `TestStep`, but the `skipStep` state now has three states (RUN_ALL, RUN_HOOKS, SKIP_ALL_SKIPABLE) instead of two (true, false).

core/src/main/java/cucumber/api/TestCase.java
```diff
@@ -50,16 +55,21 @@ public TestCase(List<TestStep> testSteps, PickleEvent pickleEvent, boolean dryRu
...
         ScenarioImpl scenarioResult = new ScenarioImpl(bus, pickleEvent);
+        boolean finishingGherkinStep = false;
         for (TestStep step : testSteps) {
+            if (skipNextStep == SkipStatus.RUN_HOOKS && finishingGherkinStep && step.startingGherkinStepType()) {
+                skipNextStep = SkipStatus.SKIP_ALL_SKIPABLE;
+            }
             Result stepResult = step.run(bus, pickleEvent.pickle.getLanguage(), scenarioResult, skipNextStep);
-            if (!stepResult.is(Result.Type.PASSED)) {
-                skipNextStep = true;
+            if (!stepResult.is(Result.Type.PASSED) && skipNextStep == SkipStatus.RUN_ALL) {
```
mpkorstanje: I find it a bit hard to qualify why but the concept underlying this solution feels wrong. TestStep is leaking information into TestCase as to how the test should proceed. It shouldn't have to.

I think this is caused by trying to represent the test execution as a list of steps rather then the tree that it is (it may help to imagine the existence of a before step hook).

coding-yogi: I have one suggestion, why can't we have skipStep as binary bit flags, or bit array or a map for eg ```{"skipBeforeStep": false, "skipGherkingStep": false, "skipAfterStep": false}```

when step fails set **skipGherkingStep** to **true** but **skipAfterStep** still remains **false** and executes the afterStep. After execution of afterStep, if **skipGherkingStep** is true , set **skipAfterStep** to **true** so that none of the next afterStep executes

mpkorstanje:
I've moved the after step hooks into the `PickleTestSteps`. This reduces the complexity of the skip logic in `TestCase` significantly. 

One thing we should still look at though are the `UnsupportedOperationException` in `HookStep` though. It smells like a refused bequest.

mpkorstanje:
I've extracted a `Step` and `HookStep` class from `TestStep`. Now it appears that `Step` doesn't contain any methods that are relevant to the event consumers of the `TestStepFinishedEvent`. They always have to cast to either `TestStep` or `HookStep`. 

This raises the question, shouldn't hook events be in their own events?

core/src/main/java/cucumber/api/Result.java
```diff
@@ -15,8 +25,8 @@
         PASSED,
         SKIPPED,
         PENDING,
-        UNDEFINED,
         AMBIGUOUS,
+        UNDEFINED,
```
brasmusson: `Ambiguous` and `Failed` both unconditionally make the exit code non-zero, but `Undefined` does not, so `Undefined` must be less severe than `Ambiguous`.

brasmusson:
> I've moved the after step hooks into the PickleTestSteps.

I'm not sure that is the right move. I have always envisioned that it is only the result of the test step from the PickleStep itself that are reported in the summary, even if an after step hook fails the summary should report "passed" of the test step from the PickleStep itself passed.

core/src/main/java/cucumber/api/Result.java
mpkorstanje: You're right.

mpkorstanje:
I am not sure I understand. The summary is created from events emitted through the bus not from the `StepResult` returned by the `PickleStep`. It should work as you've described.

The `StepResult` is only used to created the `Result` of the `TestCase` which is done by finding the most severe result and its error.

brasmusson:
> This raises the question, shouldn't hook events be in their own events?

The drive to unify the event model between the Cucumber-implementations (so that for instance common formatters processing the Json representation of the event stream), mean that we cannot do that in Cucumber-JVM without commitment that the other Cucumber-implementations also do it.

I think we in the community is very much into the view that everything that is executed during the execution of a "Test Case" is a "Test Step". A "Test Step" can be related to 0..1 "Pickle Step" (0 meaning that it is a "Test Step" created from a hook), and it can be related to 0..* code methods/functions (0 meaning that it is a "Test Step" created from a "Pickle Step" that could not be matched to any step definition - undefined, and * meaning that it is a "Test Step" created from a "Pickle Step" that could be matched to several step definitions - ambiguous). The "started" and "finished" event should therefore be of the same type for all "Test Step"s.

brasmusson:
>  It should work as you've described.

You are right.

I think we can use a hierarchical model for steps in test cases internally in the runner. However, I think that if it is possible to get the steps of a test case through classes in the api packages, then it should be a flat list of test steps, so that the abstraction of a test case as a flat sequence of test steps is maintained there.

mpkorstanje:
> However, I think that if it is possible to get the steps of a test case through classes in the api packages, then it should be a flat list of test steps, so that the abstraction of a test case as a flat sequence of test steps is maintained there.

That would be the depth first iteration order over the step tree.  Should be easy enough to change TestCase.getTestSTeps()` to do that

mpkorstanje:
> I think we in the community is very much into the view that everything that is executed during the execution of a "Test Case" is a "Test Step".

We should conform to that then. How should we name set of proper `TestSteps` and `HookSteps`? I've currently named them `Step`. I reckon `TestStep` would be a better name but that leaves the proper `TestSteps` without a name.

mpkorstanje:
What is the expected behavior of the `TestStepStarted` and  `TestStepFinished` events for hooks on a dry run? Currently the `AfterStepHook` emits and event while the other hooks don't. They're not added. Presumably because they're unskipable but ithat seems to be rather inconsistent with the idea that they're regular test steps.

coding-yogi:
Sorry, was off for a week, I will go through the changes as it looks like lot has changed from where I left :) , @brasmusson has some views on html formatter, I will also take a look

kathyrollo:
Hi Yogi,
Thank you for working on this much requested Cucumber feature. I would like to clarify if this also includes @BeforeStep hook?

coding-yogi:
@kathyrollo, yes it will include @beforestep as well

mpkorstanje:
@mlvandijk can you explain which documentation needs to be updated where?

mlvandijk:
@mpkorstanje Yes, ofcourse! The [Cucumber reference](https://github.com/cucumber/docs.cucumber.io/blob/master/content/cucumber.md) contains a section on [Hooks.]

This page is "polyglot" meaning it contains "blocks" on Java/JavaScript/Ruby. The new Hooks in cucumber-jvm should be added to the relevant section. And I'd add something like "As of version xx.." so there's no confusion ;)

If possible, please provide a PR. If not, create an issue with the information to add, and I'll add it.

mlvandijk:
More specifically, [Step Hooks](https://cucumber.netlify.com/cucumber/#step-hooks)

coding-yogi:
@mpkorstanje , @brasmusson, have some questions regarding displaying of step hooks in HTML report

- Should step hooks follow the same approach of Before/After hooks i.e. they will be displayed only if ```status != "Passed"```
- Should they appear as part of the Steps List? Something like below? Currently its only list of steps
```
   <ol class="steps">
        <li class="beforestep">
        <li class="step">
        <li class="afterstep">
        <li class="beforestep">
        <li class="step">
        <li class="afterstep">
```

core/src/main/java/cucumber/runtime/formatter/JSONFormatter.java
```diff
@@ -294,13 +300,17 @@ private void addHookStepToTestCaseMap(Map<String, Object> currentStepOrHookMap,
...
 
+
         Map<String, Object> mapToAddTo;
         if (hookType == HookType.After || hookType == HookType.Before) {
             mapToAddTo = currentTestCaseMap;
         }
         else {
-            //get latest added step
-            mapToAddTo = currentStepsList.get(currentStepsList.size() - 1);
+            if (hookType == HookType.AfterStep)
```
mpkorstanje: I think it'd be better to explicitly check the hook type.  `if (hookType == HookType.AfterStep || hookType == HookType.BeforeStep) ` and flatten the nested if statement into `else if`.

mpkorstanje: Or a `switch` statement.

coding-yogi:
I will fix the nested if..else thing. Any suggestions of the HTML Reporter part? I am in process of fixing it, would be go to know how it should look like in case of step hooks.

I feel they should be displayed in 2 scenarios
1. If they fail
2. If they generate any embeddings

core/src/main/java/cucumber/runner/PickleTestStep.java
```diff
@@ -19,23 +19,32 @@
...
+        for (HookStep before : beforeStepHookSteps) {
+            Result result = before.run(bus, language, scenario, skipSteps);
+            skipNextStep |= !result.is(Result.Type.PASSED);
+            results.add(result);
+        }
+
+        results.add(super.run(bus, language, scenario, skipNextStep));
 
         for (HookStep after : afterStepHookSteps) {
             results.add(after.run(bus, language, scenario, skipSteps));
```
coding-yogi: This method call should also accept param ```skipNextStep``` instead of ```skipSteps```. Currently if ```skipNextStep``` evaluates to ```true``` after a failed ```BeforeStep```, the ```step``` is skipped but ```AfterStep``` still executes

coding-yogi:
This is how the HTML report will look like with the current changes I have made. 

- Step hooks appear as part of Steps 
- If Step hook embeds text or image it is visible in report , else it's just a node element with status
- BeforeStep and AfterStep are skipped for all skipped steps

**Note** : 2nd step is failed on purpose, just for testing

screen shot 2018-03-13 at 3 39 37 pm

brasmusson:
>> I think we in the community is very much into the view that everything that is executed during the execution of a "Test Case" is a "Test Step".

> We should conform to that then. How should we name set of proper TestSteps and HookSteps? I've currently named them Step. I reckon TestStep would be a better name but that leaves the proper TestSteps without a name.

I think we should go with `TestStep` instead of `Step`, and then use `PickleStepTestStep` and `HookTestStep` for the proper test steps created from `PickleSteps` and `Hooks` respectively. Might seen like unnecessary long names, but it makes it clear that it is about "test steps for pickle steps" and " test steps for hooks" respectively.

brasmusson:
> What is the expected behavior of the TestStepStarted and TestStepFinished events for hooks on a dry run? 

In Cucumber-Ruby test steps for hook are only added to the test case when not in dry run. IRCC I implemented the same for before and after hooks in Cucumber-JVM. So I would say that the expected behavior is that hook are neither executed nor are there any events issued for them in dry run.

brasmusson:
> I've changed how dry run interacts with before and after hooks for now. The before and after hooks are always added to the test case. This keeps all the decisions related to dryrun isolated to the testcase.

Cucumber-Ruby takes the opposite approach. The test case is there totally ignorant of dry run. Dry run handled totally when creating the test case, in that hooks are not added and that the step matches used always report themselves as skipped.

brasmusson:
> Should step hooks follow the same approach of Before/After hooks i.e. they will be displayed only if status != "Passed" . But in case of Before/After hooks the only other status is Failed, but Step hooks may have status as Failed or Skipped. What should happen if Step hook is skipped?

If before and after hooks in practice only is displayed when failed, then the skip hooks behave the same, that is only be displayed when failed.

brasmusson:
> Should they appear as part of the Steps List? Something like below? Currently its only list of steps

If before and after hook does not appear on the Steps List, then before step and after step hooks should not appear on it either.

mpkorstanje:
> I think we should go with TestStep instead of Step, and then use PickleStepTestStep and HookTestStep for the proper test steps created from PickleSteps and Hooks respectively. Might seen like unnecessary long names, but it makes it clear that it is about "test steps for pickle steps" and " test steps for hooks" respectively.

Agreed.

mpkorstanje:
> In Cucumber-Ruby test steps for hook are only added to the test case when not in dry run. IRCC I implemented the same for before and after hooks in Cucumber-JVM. So I would say that the expected behavior is that hook are neither executed nor are there any events issued for them in dry run.

> Cucumber-Ruby takes the opposite approach. The test case is there totally ignorant of dry run. Dry run handled totally when creating the test case, in that hooks are not added and that the step matches used always report themselves as skipped.

I don't understand why ruby does this. For example if you were to run cucumber with `--dry-run --tags @someTag` I would expect to see events from hooks tagged with `someTag` and none from those marked with `not someTag`. Now I won't get any at all which I believe defeats the point of doing a dry run.

mpkorstanje:
> This is how the HTML report will look like with the current changes I have made.

Thanks for the image! Looks good.

core/src/main/java/cucumber/runner/PickleTestStep.java
```diff
@@ -50,7 +50,7 @@ Result run(EventBus bus, String language, Scenario scenario, boolean skipSteps)
         results.add(super.run(bus, language, scenario, skipNextStep));
 
         for (HookStep after : afterStepHookSteps) {
-            results.add(after.run(bus, language, scenario, skipSteps));
+            results.add(after.run(bus, language, scenario, skipNextStep));
```
mpkorstanje: Why did you make this change? I would always expect an after step hook to run so it can clean up anything setup by a before step.

mpkorstanje: And that should have been captured in a test I guess.

coding-yogi: @mpkorstanje, if beforeStep fails, step is skipped but afterStep runs, is that a correct behaviour? for eg, I was taking a snapshot in after step and it appeared in the report even if the actual step was skipped from executing

mpkorstanje: Yes. There may be multiple before steps. If a step other then the failing one allocates some resource then it should be deallocated in an after step. Skipping the after step execution will leave that resource allocated and may cause a leak.

coding-yogi:
> Thanks for the image! Looks good.  

HTML Formatter changes are  dependent on https://github.com/cucumber/cucumber-html/pull/48 
Kindly let me know if I should push them

brasmusson:
> @brasmusson has some views on html formatter

My views on the [Cucumber-HTML project] are:
* The goal is to replaced it (see https://github.com/cucumber/cucumber-html/pull/44), so why invest time and effort in it.
* Its interface is obsolete (the functions in `formatter.js`; `this.step`, `this.match`, `this.result`, `this.after` etc. are based on the formatter interface of Gherkin v2), so if it wasn't to be replaced, it should be rewritten.
* It is buggy (for instance https://github.com/cucumber/cucumber-jvm/issues/1303).

I see two options, either
* do not do anything in the Cucumber-HTML-project for this PR and instead use the functions `this.before` and `this.after` also for the step hooks, and say that this is (not good but) good enough.
* do changes in the Cucumber-HTML-project for this PR, and fix also things like https://github.com/cucumber/cucumber-jvm/issues/1303 before releasing it.

core/src/main/java/cucumber/runner/Step.java
```diff
@@ -0,0 +1,89 @@
...
+import cucumber.api.TestStep;
+import cucumber.api.event.TestStepFinished;
+import cucumber.api.event.TestStepStarted;
+import cucumber.runtime.AmbiguousStepDefinitionsException;
+import cucumber.runtime.DefinitionMatch;
+import cucumber.runtime.UndefinedStepDefinitionException;
+
+import java.util.Arrays;
+
+abstract class Step implements TestStep {
```
brasmusson: I think `Step`should be reserved for gherkin steps (`gherkin.ast.Step`) and use `cucumber.runner.TestStep` (the same pattern as `cucumber.api.HookTestStep` and `cucumber.runner.HookTestStep`)

mpkorstanje:
While the goal might be to replace  cucumber-html nobody has invested any time into it. It also appears that both `cucumber-js` and `cucumber-ruby` are using their own formatters.  So I can see a third option:

* Drop support for the html output format until somebody fixes it.

core/src/main/java/cucumber/api/PickleTestStep.java
```diff
@@ -0,0 +1,65 @@
+package cucumber.api;
+
+import java.util.List;
+
+/**
+ * A test step matches a line in a Gherkin scenario or background.
+ */
+public interface PickleTestStep extends TestStep {
```
brasmusson: I'm not sure about `PickleTestStep`. Is it a test step of a pickle? No pickles does not contain test steps but pickle steps. Is it a test step created from a pickle? No it is a test case that is created from a pickle. I'm considering `PickleStepTestStep`, maybe a bit cumbersome but clear on that it is a test steps created from a pickle step.

core/src/main/java/cucumber/runner/Step.java
mpkorstanje: Done.

core/src/main/java/cucumber/api/PickleTestStep.java
mpkorstanje: I've renamed it to `PickleStepTestStep`. I'm generally not enamored by the `ClassNameParentClassName` pattern for the exact reasons you mentioned but anything else leads to clashes with the Gherkin namespace.

coding-yogi:
> do changes in the Cucumber-HTML-project for this PR, and fix also things like #1303 before releasing it.  

@brasmusson , I think I know what the issue is, if you see method [handleTestStepStarted](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/cucumber/runtime/formatter/HTMLFormatter.java#L163), it does not register any JSFunction for hooks, so any **write** or **embedding** happening is written first to report.js via [handleEmbed](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/cucumber/runtime/formatter/HTMLFormatter.java#L182) and then when Hook step ends , it writes hook step with status when [handleTestStepFinished](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/cucumber/runtime/formatter/HTMLFormatter.java#L178) is called.

I faced the same issue for BeforeStep and AfterStep hooks, my embeddings from StepHooks would appear inside of preceding step. So I thought of making a change where ```handleTestStepStarted``` would just register what type of StepHook has started executing and result of hook step would appear below that once the hook step ends. This is how it works for normal steps as well. So in ```report.js``` it would appear something like this

```
formatter.afterstep();
formatter.embedding("image/png", "embedded0.png");
formatter.stepHookResult({
  "status": "passed"
});
```

instead of 
```
formatter.embedding("image/png", "embedded0.png");
formatter.afterstep({
  "status": "passed"
});
```

Now in this case, embedding appears inside hook step as current step is now hook step when the HTML is generated from report.js. I think the similar logic would work for ```Before``` and ```After``` hooks as well, which would display embeddings correctly. I can try to fix https://github.com/cucumber/cucumber-jvm/issues/1303

brasmusson:
When using the pretty formatter an exception in the step definitions of last step of the scenario and an exception in an after hook are displayed exactly the same (just below the last step of the scenarios). Examples for the java-calculator example, first exception in the step definition of the last step of the scenario:
```
    Then my change should be 4     # ShoppingStepdefs.my_change_should_be_(int)
      java.lang.AssertionError: expected:<4> but was:<3>
      <snip>
```
then an exception in an after hook.
```
    Then my change should be 4     # ShoppingStepdefs.my_change_should_be_(int)
      java.lang.AssertionError: expected:<1> but was:<0>
      <snip>
```
The html formatter handles the exceptions in after hook in a similar fashion, the exception trace is populated in the last step of the scenario (is displayed below it). Based on this I question the decision to put before step hooks and after step hooks in the step list. After step hooks can be handled like after hooks and display what is needed to be displayed below the preceding step of the scenario. The problem with before hooks (and before step hooks of the first step of the scenario) is that there is no preceding step, therefore the error described in #1303. In this case what is needed to be displayed should be displayed below the scenario (between the scenario and the first step of the scenario) - which is where the pretty formatter displays exceptions and output (`scenario.write()`) from the before hooks.

coding-yogi:
@brasmusson , although I am not much aware of pretty formatter but for HTML formatter with the changes I am suggesting these issues will be resolved including https://github.com/cucumber/cucumber-jvm/issues/1303. please take a look at below images, they are generated with the changes I have made. I have put some embeddings in every hook to display how it would look like.

screen shot 2018-03-18 at 11 33 22 pm

The only issue is when there is no embedding, none of the step hooks will be displayed (which is as expected) but an empty Before and After hook will be displayed as seen in below image. I am thinking about how to get rid of it if there is no embedding

screen shot 2018-03-18 at 11 40 22 pm

coding-yogi:
I have fixed the HTML reports including issue https://github.com/cucumber/cucumber-jvm/issues/1303, let me know if I should include those changes in this PR

brasmusson:
@coding-yogi I have looked into the behavior of the HTML formatter of Cucumber-Ruby. Since the intention is to replace the HTML formatters for all Cucumber implementation with one binary that can process the event stream and produce a HTML report, we should keep the different HTML formatters as aligned as reasonable possible until then.

When using `afterStep` hooks and the HTML formatter in Cucumber-Ruby and all the step hooks pass, then nothing is added to the HTML DOM regarding the `afterStep` hooks. Therefore I do not think that the js functions for step hooks here should unconditional add anything to the HTML DOM either, that is not doing something like:
```
  this.afterstep = function(afterStep) {
    currentStep = $('.afterstep', $templates).clone();
    currentStep.appendTo(currentSteps);
```

The HTML formatter of Cucumber-Ruby will display output (in Cucumber-JVM `Scenario.write`) from:
* the step definition (this is not possible in Cucumber-JVM)
* the after step hook
* the after hook (in case of the last step)

in exactly the same way (they will all appear below the text of the Gherkin step), therefore I do not think that there is any need to do anything new to handle output and embeddings with the introduction of step hooks (like making a change where `handleTestStepStarted` would just register what type of StepHook has started), the current behavior is sufficient.

In the end I think that the changes needed in Cucumber-HTML is https://github.com/cucumber/cucumber-html/commit/d757b902c617ae83e9c0849ec9cd285729f5dba2 (now when https://github.com/cucumber/cucumber-html/issues/49 has been fixed).

mpkorstanje:
> Since the intention is to replace the HTML formatters for all Cucumber implementation with one binary that can process the event stream and produce a HTML report, we should keep the different HTML formatters as aligned as reasonable possible until then.

Is that ideal still alive? From what I can see all the other implementations moved away from it.

aslakhellesoy:
I think the single binary is a very exciting development, and it makes perfect sense to me to put formatters in there too.

mpkorstanje:
I must be misunderstanding something somewhere. I am I right in assuming the `go` project is a second attempt at unifying the different formatters? In that case nevermind.

aslakhellesoy:
@mpkorstanje: It’s much more:

https://github.com/cucumber/cucumber/issues/221

https://github.com/cucumber/cucumber-pickle-runner

coding-yogi:
@brasmusson , thanks for the clarification, so I won't push any further changes for HTML formatters

naelabdeljawad:
Hi,
Did you release new version which includes the new hooks (AfterStep/BeforeStep) for java?

aslakhellesoy:
@naelabdeljawad if you're ever in doubt about what's been released, check out the [CHANGELOG](https://github.com/cucumber/cucumber-jvm/blob/master/CHANGELOG.md)

aslakhellesoy:
An open pull request (such as this one) means the code has not been merged, and therefore, not released.

mpkorstanje:
Documentation still needs to be updated:

https://docs.cucumber.io/cucumber/#beforestep
https://github.com/cucumber/docs.cucumber.io/blob/master/content/cucumber.md

aslakhellesoy:
Hi @coding-yogi,

Aslak Hellesøy
Creator of Cucumber

aslakhellesoy:
Hi @coding-yogi,

Aslak Hellesøy
Creator of Cucumber

PR #1335, opened 2018-03-15
[Core] Remove deprecated legacy tags

aslakhellesoy:
## Summary

The old `--tags "@foo" --tags "@bar,@zap"` tags are no longer supported.

## Details

* Removed the old parser
* Only allow specifying a single expression (with `--tags` as well as in hooks)

## Motivation and Context

The old style was deprecated in a previous version, and it's time to remove it in 3.0.0

## How Has This Been Tested?

With tests

## Types of changes

## Checklist:


brasmusson:
I do not think we can limit the number of `--tags` options on the command line to one, because we support `@CucumberOptions` annotation on several levels in the runner inheritance which each can have a `tags` option (this is translated to several `--tags` options on the command line that `RuntimeOptions` is seeing). This is similar to the case in Cucumber-Ruby where there can be both be `--tags` options in the profile file (on several levels) and on the command line. There the potential multiple `--tags` are all processed with "and" semantics, I think we should do the same here in Cucumber-JVM

mpkorstanje:
Mmh. I'd like to see that the commandline, `@CucumberOptions`, the environment and any future profiles we might implement would all modify some configuration object directly rather then translating their options into command line arguments first.

brasmusson:
The behavior of:
```
@CucumberOptions(tags = "not @manual");
class RunnerBase {}

@CucumberOptions(tags = "not @wip");
class SpecializedRunner exterds RunnerBase {}
```
when using `SpecializedRunner` should not change with this PR (and after a quick look, I think it does)

mpkorstanje:
It does. I believe it also changes how Env interacts with the command line and cucumber options. 

I meant to suggest a structural improvement that would allow us to retain the cascading effects from cucumber options and other sources while removing it from the command line.

brasmusson:
The idea of modifying an configuration object seems as an improvement. It is not the support of multiple `--tags` options that I want, but the to preserve some current behavior, so I restated my first comment to focus on behavior instead of implementation.

mpkorstanje:
Then we should start by adding some tests that verify the current behaviour! :D

aslakhellesoy:
I’ve started a little library in the monorepo called cucumber-config. I’ll refactor cucumber to use that :-)

mpkorstanje:
Looks interesting! Reviewed some API stuff on there.

mpkorstanje:
I think we can suffer supporting legacy tags for another major release.

mpkorstanje:
I've rewritten this to still allow multiple tags from the configuration but only a single tag expression from the hooks.

PR #1338, opened 2018-03-31
[Core, TestNG] Support the TestNG SkipException

brasmusson:
## Summary

Let a `org.testng.SkipException` thrown from a test step make the result of that test step skipped.

## Details

Let a `org.testng.SkipException` thrown from a test step make the result of that test step skipped.

Also use the `SkipException` to signal to the TestNG runner that the test should be considered as skipped, for instance Pending and Undefined test cases in non-strict-mode.

## Motivation and Context

The JUnit exception (`org.junit.AssumptionViolatedException`) has been supported by Cucumber-JVM for some time. Cucumber-JVM should also support the corresponding TestNG exception ([`org.testng.SkipException`](https://jitpack.io/com/github/cbeust/testng/master/javadoc/org/testng/SkipException.html)).

## How Has This Been Tested?

In addition to the updated in the automated test suite, the changes has been manually tested using the `java-calcualator-testng` example in the repository.

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/api/TestStep.java
+++ b/core/src/main/java/cucumber/api/TestStep.java
@@ -15,7 +15,8 @@ import java.util.List;
 public abstract class TestStep {
     private static final String[] ASSUMPTION_VIOLATED_EXCEPTIONS = {
             "org.junit.AssumptionViolatedException",
-            "org.junit.internal.AssumptionViolatedException"
+            "org.junit.internal.AssumptionViolatedException",
+            "org.testng.SkipException"
     };
 
     static {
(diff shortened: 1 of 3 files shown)


mpkorstanje:
Apparently throwing a `SkipException` doesn't automatically mean a test should be skipped.

[SkipException#isSkip](https://jitpack.io/com/github/cbeust/testng/master/javadoc/org/testng/SkipException.html#isSkip--)
> Flag if the current exception marks a skipped method (true) or a failure (false). By default Subclasses should override this method in order to provide smarter behavior.

The [TimeBombSkipException](https://jitpack.io/com/github/cbeust/testng/master/javadoc/org/testng/TimeBombSkipException.html) is an example of this.  

I don't see a problem with ignoring this in Cucumber but I think it should be documented that we only only recognize the `SkipException` and none of its subclasses.

mpkorstanje:
And to make the documentation part even more confusing:

* Throwing a `SkipException` results in both Cucumber and TestNg marking the test as skipped.
* Throwing a subclass of `SkipException` results in Cucumber marking the test as failed and TestNG marking the test as skipped.

PR #1339, opened 2018-03-31
[TestNG] Always map Pickles/TestCases to TestNG tests.

brasmusson:
## Summary

Always map Pickles/TestCases to TestNG tests.

## Details

Remove the support for mapping the whole Cucumber test suite or each Cucumber feature to TestNG tests.

## Motivation and Context

The first TestNG support implemented in Cucumber-JVM mapped the whole Cucumber test suite to one TestNG test. It did enable running Cucumber from a TestNG runner, but did provide practically no notification of the result in the TestNG domain.

Later the TestNG support was expanded to allow the mapping of each Cucumber feature (file) to a TestNG test. It provided a little better notification of the result in the TestNG domain, but it was still limited.

Now the natural mapping, one Pickle/TestCase to one TestNG test, is supported, and a major release is on the way, the legacy support for mapping the whole Cucumber test suite or each Cucumber feature to TestNG tests should be removed.

## How Has This Been Tested?

The automated test suite has been updated to cover the remaining behavior.

## Types of changes

## Checklist:


testng/src/main/java/cucumber/api/testng/PickleEventWrapperImpl.java
```diff
@@ -0,0 +1,27 @@
...
+
+import gherkin.events.PickleEvent;
+
+/**
+ * The only purpose of this class is to provide custom {@linkplain #toString()},
+ * making TestNG reports look more descriptive.
+ *
+ * @see AbstractTestNGCucumberTests#runScenario(cucumber.api.testng.PickleEventWrapper, cucumber.api.testng.CucumberFeatureWrapper)
+ */
+public class PickleEventWrapperImpl implements PickleEventWrapper {
```
mpkorstanje: This should probably be package private.

mpkorstanje:
* I've tightened up the the modifiers to avoid exposing the internal api. 
* I've removed unused any unused methods.
* I've removed the javadoc from the `*Impl` classes. The interfaces should describe everything. 
* I've added a ReadMe that describes how TestNG can be used. Including the composite version I wasn't aware of.

mpkorstanje:
Everything else looks good to me.

mpkorstanje:
I'm moving the read me to a separate PR so I can also include  #1338.

PR #1344, opened 2018-04-12
[Core] Add feature list to TestRunStarted event

stuartbirrell:
## Summary

Extended the TestRunStarted event to take in the list of CucumberFeatures that is scheduled to be executed.

## Details

I have extended the TestRunStarted event to accept additional property of List<CucumberFeature>. Which is the list of features that will be executed by the run.

## Motivation and Context

In terms of creating customised plugins to monitor and report on execution. It would be a great addition to the TestRunStarted event to share the list of features that are scheduled to be executed in the run.

It would facilitate run progression tracking(Since the TestCaseStarted & TestCaseFinished events both declare what test case has been executed.)

## How Has This Been Tested?
Executing the existing suite of tests to ensure no regressions.

I have also stepped through execution using a custom plugin that I have created which analyses the events.

Additional statement/Question:
I had a look around the source - there does not appear to be specific tests that check any events? Is this an oversight?

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/api/event/TestRunStarted.java
+++ b/core/src/main/java/cucumber/api/event/TestRunStarted.java
@@ -1,8 +1,14 @@
 package cucumber.api.event;
 
-public final class TestRunStarted extends TimeStampedEvent {
+import cucumber.runtime.model.CucumberFeature;
 
-    public TestRunStarted(Long timeStamp) {
+import java.util.List;
+
+public final class TestRunStarted extends TimeStampedEvent {
+    public final List<CucumberFeature> features;
+
+    public TestRunStarted(Long timeStamp, List<CucumberFeature> features) {
         super(timeStamp);
+        this.features = features;
     }
 }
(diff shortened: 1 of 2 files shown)


aslakhellesoy:
Please fill in the template (summary, details, motivation etc) and add tests.

Otherwise I'm afraid we'll have to reject this.

mpkorstanje:
`CucumberFeature` is part of the internal api (it's not in the `cucumber.api` package) so it shouldn't be exposed. All of the information in `CucumberFeature` is already exposed through the `TestSourcesRead` event.

The `TestSourcesRead` does lack a version of the feature in the form of a `gherkin.ast.GherkinDocument` but you can compile it yourself from `TestSourceRead.source`  or we can have a discussion about including it or not.

I don't quite see a reason too though.

brasmusson:
I do not think that the `TestRunStarted` event should contain any information about the features to be executed, because the `TestRunStarted` event should be issued before the file system has been searched for the feature files of the feature path arguments. 

Cucumber-JVM does not do that because we still live with the design from Cucumber-JVM v1.x.x that a parse error in a feature file stops the execution immediately. 

A better design would be to (after the `TestRunStarted` event) issued `TestSourceRead` events and `ParseError` events (and possibly execute all the feature files not raising a parser error).

I also think that we should avoid creating phases, there is no reason why (with parallelism) test cases from some features have started to be executed before all feature files have been read and parsed, rather than an event listing all the features, an event notifying that all the feature files have been read is preferable (then the `TestSourceRead` events combined have listed all features to be executed). Until that event has been received, a monitor plugin will experience a parallel build up of features to be executed and test cases executed.

PR #1354, opened 2018-04-24
Android-Studio Examples, Prepare build.gradle for sharding capability

spalguyev:
## Summary
Prepare build.gradle for sharding capability and parallel execution

## Details
1) Allow multiple devices to be connected. Fixes gradle exception, when multiple devices are connected.
2) Add additional separate test result outputs, per device.

## Motivation and Context
There was an exception thrown by gradle, when multiple devices were online. Since most android projects are tested using multiple devices, in both parallel and sharded manner, it only made sense to fix the issue, and allow examples to run in parallel on multiple devices.

It would be great to add an ability in a future PR, where all device test results would be merged into a single cucumber result set. Currently adb pull from the last device seems to be overwriting the adb pull from a previous device.

## How Has This Been Tested?
1) Ran examples/android/android-studio/Cukeulator on a single emulator
2) Ran examples/android/android-studio/Cukeulator on multiple emulators

## Screenshots
Additions to folder structure
screen shot 2018-04-23 at 7 58 13 pm

## Types of changes

## Checklist:

### DIFF
--- a/examples/android/android-studio/Cukeulator/app/build.gradle
+++ b/examples/android/android-studio/Cukeulator/app/build.gradle
@@ -1,3 +1,7 @@
+import com.android.ddmlib.AndroidDebugBridge
+import com.android.ddmlib.IDevice
+import com.android.ddmlib.CollectingOutputReceiver
+
 apply plugin: 'com.android.application'
 
 // ==================================================================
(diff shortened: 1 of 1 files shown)


mpkorstanje:
This is really nice but this example should be a minimal example. This makes it easier to understand what is going on. Sharding capability and parallel execution aren't needed for that. And from a more practical perspective, would be fairly hard to test.

Issue #630, opened 2013-11-12
Support concurrent executions of scenarios


whiskeysierra:
Currently it's only possible to execute feature classes in parallel. In case you have independent scenarios (and you really should) it would be nice to run them concurrently. The current codebase does not support this. First of all the reporting is inherently sequential and there is some state in the Runtime class (`skipNextStep` and `scenarioResult`). I would be happily contributing a patch, but this would require a major change of some of the internal structure.

danielwegener:
:+1: I am carrying around this issue with me for quite a while now. Especially for UI tests it is absolutly nessessary to run tests in parallel. I've seen several attempts to make this possible from slicing feature files into single scenario files (within a build job) with massive jvm-forking to runners with ugly Reflection-Hacks that replace thread-unsafe code in Runners and Runtimes - none of them very satisfying.  I tried to build a JUnit Runner around it but there are some internal design flaws that absolutely prevent one from efficiently executing cucumber scenarios in parallel. The core problem is IMO that the `Runtime` and `Backend` both couple expensive operations, like class- and glue-loading as well as the state of a single executing scenario (or world). Splitting this up a bit and seeing Runtimes as a factory of ExecutionContext where a single Scenarios could be run with, would make this much easier. Additionally, reporters (at least JSON and XML) are not thread safe and would need an interface overhaul or need to be proxied, such that reportings are recorded during parallel test execution and replayed  in serial when a scenario execution is done.

I'd really like to spent some time on this but it would, in one way or another introduce changes in the core apis. Thats why I'd kindly ask if such changes would be welcome at all :)

dkowis:
I broke the hell out of core back in the day making the Jruby stuff work. I don't believe anyone would be opposed to seeing code. You'd just have to be willing to accept feedback and possibly change the code even more.

aslakhellesoy:
Hear, hear!

It would be a big, but welcome change.

In the next months I intend to refactor the internals to use [Gherkin3](https://github.com/cucumber/gherkin3) and in order to avoid merge hell I'd like to see parallelism added after this.

klausbayrhammer:
Two thumbs up @danielwegener. I really appreciate the proposed changes. If there's anything I can support you with, just let me know. I'll close #664 so we track the parallel scenario executions in one issue - and this one looks more promising!

danielwegener:
Good to hear that I am not alone with this requirement :) I find some time to investigate the code a bit and tried to push pieces around.
[...]

PR #1357, opened 2018-04-25
Parallel support

boaty82:
I’ve previously used another implementation to deliver parallel execution of cucumber tests, however it basically wrapped up the creation and execution of `Runtime`’s, followed by the need to merge the multiple JSON/HTML etc reports that followed.

I reviewed the current implementation and also took note of comments in [Support concurrent executions of scenarios](https://github.com/cucumber/cucumber-jvm/issues/630) also I perused all tickets related to running in [parallel cukes](https://github.com/cucumber/cucumber-jvm/issues?utf8=%E2%9C%93&q=is%3Aissue+parallel) for other insights

### Added `--threads` argument to runtime options 
Allows users to specify the (max) number of threads to be used to run the tests

The actual number of threads used will be the smaller of the given `--threads` value and the number of features to be executed

Tests are placed into a `Queue` and consumed by each thread when needed, resulting in a less deterministic order of Features being run

### Ability to still run tests synchronously
There may be situations where certain Features/Scenarios need to be run in a synchronized/serial manner.  To this end features/scenarios will be scanned for usages of a tag beginning with `@synchronized` (case insensitive) and will run these Feature files separately, prior to any Feature files not found to have this tag.  Buckets of synchronized Feature files can also be created by using any suffix e.g. `@synchronized-1` or `@synchronized-bob`

### New formatter introduced `TimelineFormatter` 
Which produces reports using [vsjis.org timeline](http://visjs.org/examples/timeline/groups/groups.html) to highlight which feature was run on which Thread and when.  
**Note:** the resources (js, html etc) are currently within `cucumber-core` if the PR is accepted then I’d imagine these would have to be moved out into its own project, similarly to cucumber-html.
timeline
[sync_and_parallel_timeline.zip](https://github.com/cucumber/cucumber-jvm/files/1947348/sync_and_parallel_timeline.zip)

### `Glue` / `RuntimeGlue`
This was the main problem in that it serves 2 purposes.  

1. Glue Loader/Builder
2. Glue Context

Before tests are executed the glue is loaded (`Backend.loadGlue`) and then during the test run the step defs etc held in cache are modified, and hence the glue instance cannot be shared.

Also it appears that `JavaBackend` held onto the glue reference to be used during `buildWorld` and `disposeWorld` calls - to support Java8 Lambdas

Ideally this class/interface would be broken down into more granular APIs to reduce this issue, but this is quite a big change.

To reduce the number of breaking changes I went with the premise that the Glue initially loaded is a template and it is cloned per thread during execution, which resulted in very few changes to be made.  

But there were 2 changes - `Backend.buildWorld` and `Backend.disposeWorld` - whereby the glue is now passed in as an argument, so it is the glue “clone” applicable to the running thread.  This means any other implementations of `Backend` outside of this repository need updating.

Unfortunately, I could not think of another way around this, but happy to discuss further.

## Others Issues encountered / rectified
- `Stats`
Was not thread safe

- `UndefinedStepsTracker`
Was not thread safe

- `EventBus` thread safety
`registerHandlerFor` is not thread safe, originally I updated it to be but couldn’t find any sign of sharing instances - so I reverted the change

- `UnreportedStepExecutor` implementation
Extracted this out of `Runner` and into its own class
Glue is injected, but as there was no real usage of this functionality I couldn’t tell if the Glue “template” was OK to be injected or not - I assume it is but have left a TODO comment for review/discussion

- `LocalizedXStreams` not thread safe
Internal Map to hold cache of `LocalizedXStream` was not threadsafe

- `Formatter`s output being incorrect as not thread safe
I have updated all Formatter implementations located in `cucumber-jvm`, specifically
`AndroidInstrumentationReporter`, `HTMLFormatter`, `JSONFormatter`, `JUnitFormatter`, `PrettyFormatter`, `RerunFormatter`, `TestNGFormatter`, `UsageFormatter`
to be thread safe
Also added a note to the javadocs. 
**Note** any impl's in other repositories still need reviewing/updating

- `TestSourcesModel` updated this to be thread safe in the event any instances are shared (none in cucumber-jvm repo)

- Update `ObjectFactory` implementations to be thread safe 
Specifically
`DefaultJavaObjectFactory`, `GuiceFactory`, `NeedleFactory`, `OpenEJBObjectFactory`, `OsgiObjectFactoryBase`, `PicoFactory`, `WeldFactory` all were minor changes so I felt new tests were not needed
`SpringFactory` appeared to be threadsafe already so I added some tests to `SpringFactoryTest` and it proved to be the case
**Note** any impl's in other repositories still need reviewing/updating


boaty82:
Oh and please note that tests in `Examples: Pax Exam: Calculator Test` were failing as soon as I cloned the repository
```[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.74 s <<< FAILURE! - in cucumber.examples.java.paxexam.test.CalculatorTest
[ERROR] cucumber(cucumber.examples.java.paxexam.test.CalculatorTest)  Time elapsed: 0.56 s  <<< ERROR!
shaded.org.eclipse.aether.resolution.VersionRangeResolutionException: No highest version found for io.cucumber:pax-exam-calculator-api:jar:(0.0,]
```
I just haven't had a chance to look into this

boaty82:
Hmm seems that some of my tests are not working remotely.  

Will happily fix this, but for now would like review/feedback on the changes in the PR

aslakhellesoy:
Wow that's some significant work!

I haven't had time to look into the details yet, but I wanted to make you aware that we're pursuing a [shared binary](https://github.com/cucumber/cucumber-pickle-runner) which grew out of https://github.com/cucumber/cucumber/issues/221

We'll need to make significant refactorings to support that, and because of the size of this PR, they can't happen in parallel. I can see three options:

* Refactor to use pickle runner after this PR gets merged (assuming we don't reject it)
* Refactor to use pickle runner first, then consider this PR (which would probably have to be rewritten)

What are the tradeoffs for each approach?

boaty82:
Yeah was a LOT of work!  Also spent quite a lot of time going down various routes too, to only find a problem e.g. JavaBackend modifying glue during buildWorld/disposeWorld.

But anyway, the tradeoffs.  Someone is going to have some fun merging and re-testing.  Happy to discuss this at the point someone is happy with the PR (in premise) and then the additional changes/merging can be discussed and put into action at that point.

How does that sound?

boaty82:
Also, having reviewed the Timeline output, for some reason the dates/times being rendered say June when infact the test was run today.  I need to look into that, may be some difference between Java and Javascript or something

boaty82:
it would appear that I need to rethink/adjust my parallel tests in CI - as I may request 3 threads but they may not be available

```
[ERROR] shouldHandleTestsBeingRunConcurrently(cucumber.runtime.formatter.TimelineFormatterTest)  Time elapsed: 0.043 s  <<< FAILURE!
java.lang.AssertionError: Number of groups was not as expected expected:<3> but was:<1>
	at cucumber.runtime.formatter.TimelineFormatterTest.assertTimelineDataIsAsExpected(TimelineFormatterTest.java:244)
	at cucumber.runtime.formatter.TimelineFormatterTest.shouldHandleTestsBeingRunConcurrently(TimelineFormatterTest.java:198)
```

again, will await feedback before investing more time in this

mpkorstanje:
I had a real brief look at this. I am really impressed by the amount of work that has been done. 

However the architecture isn't good.

The complexity of running features in parallel should not be leaking into all these different parts. Ideally only a very small part should be aware of the multi-threaded nature. This keeps the tests simple and gives us a reasonable grantee that if a component passes its unit tests it will behave the same when used with multiple threads.

Another problem with this architecture is that everything attached to the event bus has to be made thread-safe. This places quite burden on people who wish to implement a plugin. The same goes for the object factory.

I think a way to resolve this would be by creating multiple instances of Runner, Glue, Backend and ObjectFactory and EventBus, one for each thread rather then sharing the same instance between threads. The main thread can then pick up the messages and pass them onto the different plugins.

boaty82:
@mpkorstanje thanks for the feedback, very much appreciate it and your time!

Whilst I understand and agree with most of your comments, I do have issue with some of them
```
However the architecture isn't good.

The complexity of running features in parallel should not be leaking into all these different parts. 
Ideally only a very small part should be aware of the multi-threaded nature
```

 I made changes based on comments in other PRs e.g. keep changes to a minimum, don't call expensive methods multiple times - hence I worked within the *existing architecture* rather than changing a lot of it.

There was very little change in the Formatters other than mainly refactoring member variables into wrapper classes.  I personally believe there is a problem with the Formatter API, well more to the point, in the actual implementations.  They are receiving the events *and* handling the processing of what they should look like in the final output.  Ideally the result of the tests would be held until all are complete and then they would be written out in the main thread, avoiding the need for each impl to be thread safe - whether or not this would increase the memory footprint is questionable as you would have 1 class holding the results, rather than each formatter, and then this could, at the end, pass the results off to each of the plugins to be processed as an output.  **But** this would not work with the all the implementations e.g. ProgressFormatter

`Another problem with this architecture is that everything attached to the event bus has to be made thread-safe. This places quite a burden on people who wish to implement a plugin.`
I quite agree, and didn't like effectively having APIs that require any impl's to be threadsafe to support the parallel running feature.  But again I was trying to avoid changing too much of the existing architecture

`The same goes for the object factory.`
I did look into creating an ObjectFactory, so I could create on per thread but it wasn't necessary after fixing the thread issues in the various impl's and cloning the glue.

`I think a way to resolve this would be by creating multiple instances of Runner, Glue, Backend and ObjectFactory and EventBus` 
- Glue I did implement creating multiple instances, though a constructor could've probably achieved the same as the clone method
- Backends and Glue can be passed in as arguments to Runtime, so creating new instances isn't straight forward - either cloning must be implemented per impl, or reflection (or a clone framework) would be needed and could be slow
- Loading glue appears to be an expensive operation and thus calling it multiple times should really be avoided
- Multiple instances of EventBus, this could have wide reaching changes.  Plus each formatter, Stats, UndefinedStepsTracker and potentially other classes (in other repos') are registered to the EventBus to receive notifications.

I'm more than happy to have another go, with the provision of changing some of the architecture, along with taking into account your feedback - plus any response you have to this comment

mpkorstanje:
I should have started with saying the current architecture of cucumber-jvm isn't very good either. Lots of stuff is in the wrong place.

I don't disagree in principle with most points. I do think they are all less important then getting the the architecture right. Especially performance optimizations can come afterwards. So I don't have a problem changing more then required if it improves the design.

So lets talk design:

The goal should be to 
  * minimize the the amount of state shared between threads.
 * ensure any object shared between threads is immutable and/or threadsafe.

## Naive ## 

To run a scenario a pickle goes in and events come out. To achieve our goals and execute pickles in parallel we can create a `FutureTask<List<Event>>` that creates new `Backend` and `EventBus`. The `EventBus` used in this task will simply store all messages. Once the task is complete the `Runtime` get the events and replays them on it's own bus.

Performance wise this will be pretty terrible and events will be late but we'll only have to ensure a handful of messages are immutable.

## Fixing Events  ##

Instead of returning all the events as the result of the completion of`FutureTask` we stream them as they happen. This can be done by replacing the tasks event bus by one that writes to a `ConcurrentQueue`. This bus will be shared with all tasks. After all tasks have been submitted to the executor the Runtime will start to poll the event queue and put the messages on its own bus. 

So far only the bus that is shared between tasks needs to be thread-safe.  And maybe split event bus interface into one specific for event consumers and event generators. 

## Reusing Backend and Glue ##

Work out a way to reuse the backend between tasks on the same thread. Don't know how yet. Probably comes down to implementing our own executor. I don't think reusing the backend is that important though. It is cheap to create.

I think the best way to reuse the Glue would be to separate the mutable scenario scoped glue from the other glue. Each backend can be initialized with an immutable glue object, and wrap it in a mutable glue object to hold the scenario scoped glue and simply discard it after the run and create a new glue object to hold the scenario scoped glue.

Doing it this way we only have to worry about creating the immutable glue once before any of the runners start and making sure it is immutable.

mpkorstanje:
I've just played around with Surefire a bit to see how their JUnit integration handles parallelism. Using the configuration below Cucumber scenarios will be run in parallel.

```
  <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-surefire-plugin</artifactId>
                <version>2.21.0</version>
                <configuration>
                    <threadCount>2</threadCount>
                    <parallel>both</parallel>
                </configuration>
            </plugin>
```

It does however break in quite a few ways. Mostly because the `PickleRunners` are reusing the same runner/backend/object factory/event.  

```
Feature: Setting the check in the Given lambda expression org.picocontainer.PicoCompositionException: Duplicate Keys not allowed. Duplicate for 'class com.example.SettingBCheckInGivenTest'
        at org.picocontainer.DefaultPicoContainer.addAdapterInternal(DefaultPicoContainer.java:438)
        at org.picocontainer.DefaultPicoContainer.addComponent(DefaultPicoContainer.java:548)
        at org.picocontainer.DefaultPicoContainer.addComponent(DefaultPicoContainer.java:501)
        at org.picocontainer.DefaultPicoContainer.addComponent(DefaultPicoContainer.java:488)
        at cucumber.runtime.java.picocontainer.PicoFactory.start(PicoFactory.java:22)
        at cucumber.runtime.java.JavaBackend.buildWorld(JavaBackend.java:116)
        at cucumber.runner.Runner.buildBackendWorlds(Runner.java:155)
        at cucumber.runner.Runner.runPickle(Runner.java:78)
        at cucumber.runtime.junit.PickleRunners$NoStepDescriptions.run(PickleRunners.java:140)
        at cucumber.runtime.junit.FeatureRunner.runChild(FeatureRunner.java:68)
        at cucumber.runtime.junit.FeatureRunner.runChild(FeatureRunner.java:23)
        at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
        at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
        at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
        at cucumber.runtime.junit.FeatureRunner.run(FeatureRunner.java:73)
        at cucumber.api.junit.Cucumber.runChild(Cucumber.java:118)
        at cucumber.api.junit.Cucumber.runChild(Cucumber.java:56)
        at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
        at org.apache.maven.surefire.junitcore.pc.Scheduler$1.run(Scheduler.java:410)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

mpkorstanje:
I had some more time to play with it and I now better understand why you're reusing the backends and cloning the glue. I think we can remove both from the Runtime constructor. Or if people still need them replace them with `Providers`.

mpkorstanje:
I hacked together a way to get JUnit to run in parallel. Just to get a feel for stuff it needs: https://github.com/cucumber/cucumber-jvm/tree/junit-parallel

I used your idea about event buffering and flushing only after the run. It is really ugly at the moment but it works.

However I don't see a way to get rid of the runner.  The buffer-flush system needs a hierarchy of of event buses that flush upwards. Not just for the formatters but also because JUnit's event listener is also listening to the bus. But it should only listen to the bus of one specific runner. So it has to access the runner to get its bus.

> My initial thoughts regarding Events were that I thought of retaining certain events (by introducing a new interface), per thread, until an "ending" event was received e.g.

I like this. 

> Good to know you are OK with removing public constructors, where necessary, usually I would be concerned about doing so without initially deprecating for a version or so - but the kind changes needed necessitates a breaking change

In general, the occasional exception not withstanding, anything outside of `cucumber.api` can be changed with relative ease. And otherwise, if it is an improvement and you're willing to write a short migration guide, major version numbers are free of charge.

> I'd made some changes already before reading your feedback and created a BackEndFactory which has allowed me to defer the creation of backends until necessary and also allowed me to pull out the logic in cucumber.runtime.Runtime#loadBackends thus reducing the complexity of Runtime

It might be a good idea to start over from scratch.

 * I think it would better to make it possible to run pickles in parallel through JUnit or TestNG before we make this possible through the command line and start adding extra features like `@synchronized` or command line options like `--parallel`. 
* I also think it will be okay to leave out the performance improvements to Glue for now.  Don't throw the code away though.

By using Surefire/JUnit/TestNG we'll have a system that can be run in parallel without having to work out how this is done or deal with any of the resulting problems. We just have to make it possible to run pickles in parallel in the backend. This reduces the amount of code and features we'll have to discuss.. Any extra features and performance improves should also be fairly easy to add in following PRs.

boaty82:
> I hacked together a way to get JUnit to run in parallel. Just to get a feel for stuff it needs: https://github.com/cucumber/cucumber-jvm/tree/junit-parallel

OK, I will take a look

> I used your idea about event buffering and flushing only after the run. 

:smiley: good to know my thoughts were on the right lines

> Not just for the formatters but also because JUnit's event listener is also listening to the bus. But it should only listen to the bus of one specific runner and it does need those events in real time

I was thinking of creating a decoratpr for the original `EventBus` per running thread which can store the event(s), until they should be dispatched - they could simply then dispatch/flush them onto the original `EventBus` - just need to be careful and ensure synchronization is done correctly to avoid the issue with the `EventBus` receiving events from multiple sources and thus the events being out of sequence.

>It might be a good idea to start over from scratch.
I think it would better to make it possible to run pickles in parallel through JUnit or TestNG before we make this possible through the command line and start adding extra features like @synchronized or command line options like --parallel.

OK sounds like a reasonable plan of attack.  I still believe these features should be added, as it's always good to test a system with many threads (as a replacements for users/customers) to ensure performance/thread/state safety

>I also think it will be okay to leave out the performance improvements to Glue for now. Don't throw the code away though.

Don't worry I won't I'll just stash it for now

mpkorstanje:
> I was thinking of creating a decorator for the original EventBus per running thread which can store the event(s), until they should be dispatched - they could simply then dispatch/flush them onto the original EventBus - just need to be careful and ensure synchronization is done correctly to avoid the issue with the EventBus receiving events from multiple sources and thus the events being out of sequence.

Sounds about right. 

We can synchronize on `send` and add a synchronized `sendAll` method.

mpkorstanje:
> OK sounds like a reasonable plan of attack. I still believe these features should be added, as it's always good to test a system with many threads (as a replacements for users/customers) to ensure performance/thread/state safety

Cucumber isn't a performance or load testing tool though. Scenarios can't be repeated and the rate at which tests are executed is unpredictable and uncontrollable. Using it as such would give you meaningless results.

The only good use case for parallel execution is to speed up tests that spend a significant amount of time waiting for things to happen (webdriver, io, ect). Calling glue is inherently slow so cucumber won't  ever be able to output any significant volume and the message bus and attached listeners are going bottleneck any serious amount of parallel work.

mpkorstanje:
Ah in that case I think your needs should be met. Surefire has a large suite of configuration options for running in parallel. You can configure the number of forks per build, the number of threads per fork per cpu, ect.  

It'd still be nice to have `--threads` eventually though but most people don't use command line runner.

boaty82:
Not quite.  I still have a use case for `--threads` argument, in which I need tests from within a running application, via Cucumbers Runner, as the application (not mine, but a commercial platform) is quite monolithic and takes ~5-6 mins to startup - plus we need to execute multiple test sets so don't really want to wait for startup each time

mpkorstanje:
That is a reasonable use case. Let's discuss it in more depth when we get to it. We may learn more things along the way.

mpkorstanje:
Would you be able to work off  #1367? 

I've pushed some refactoring to this branch that will make it easier to implement parallel support aswell as some other refactoring that are happening. I don't want to merge it just yet because we've just released 3.0.0. Best to give people some time to find bugs before everything is changed again.

boaty82:
If I can, I will.  I’ll pull it down and start working on it in the next few days

boaty82:
@mpkorstanje do you want me to fork from the `refactor-runtime-constructors` branch?

Which tasks do you specifically want me to work on?  Or are you expecting me to just add support for `--threads` ?

mpkorstanje:
@boaty82 I've re-factored the runtime to the point that parallel execution should be fairly easy to implement. JUnit and TestNG may already work but I've yet to confirm this. 

I believe you should be able to implement parallel support for the Runtime and make the plugins able to handle scenarios that are reported out of order. I am thinking it'd be best to leave `--parallel` out of `@CucumberOptions`. This should be left to JUnit and TestNG.

mpkorstanje:
And yes, you please work from refactor-runtime-constructors branch.

mpkorstanje:
> Ability to still run tests synchronously
> There may be situations where certain Features/Scenarios need to be run in a synchronized/serial manner. To this end features/scenarios will be scanned for usages of a tag beginning with @synchronized (case insensitive) and will run these Feature files separately, prior to any Feature files not found to have this tag. Buckets of synchronized Feature files can also be created by using any suffix e.g. @synchronized-1 or @synchronized-bob

I don't think we should support this. People are told to write their scenarios in such a way that they are independent from each other.  So we shouldn't make an effort facilitate this sort of behavior. It also won't be possible to support this from JUnit or TestNG.

However it is very easy to implement this behavior your own test scripts.

```
cucumber --tags @synchronized
cucumber --tags "not @synchronized"  --threads 10
```

Or when running multiple sets

```
cucumber --tags "@synchronized and @synchronized-1"
cucumber --tags "@synchronized and @synchronized-2"
cucumber --tags "not @synchronized" --threads 10
```

I happen to know that for JUnit and TestNG maven provides similar functionality. You can create multiple executions with different configurations and select different tests in each configuration. I don't know about gradle.

boaty82:
`I don't think we should support this. People are told to write their scenarios in such a way that they are independent from each other`
Hi, that's fine I won't add it.  We already are doing what you describe (multiple executions) but just thought it would a nice supported feature. 

BTW features being independent is not the problem (as they are), but the production functionality being executed is not independent e.g. timed jobs

mpkorstanje:
I believe the scheduler has a queu for runnables to be executed.

boaty82:
`I believe you should be able to implement parallel support for the Runtime and make the plugins able to handle scenarios that are reported out of order`

But I thought we agreed that issues like this shouldn't leak into many classes.  I'm going to  decorate `EventBus` to queue certain types of events from the thread until another certain event e.g. `TestCaseFinished` was received, upon which it could then flush all the events to ensure they are received in the expected order

I can see you done something like this in `BatchEventBus` but I'm a little confused by the
```
        @Override
        public void send(Event event) {
            super.send(event);
            queue.add(event);
            if(event instanceof TestCaseFinished){
                parent.sendAll(queue);
                queue.clear();
            }
        }
```
I'm not 100% sure why it should be calling `super.send(event)` for all events as well as adding them to the queue to then be sent again later once `TestCaseFinished` is received

mpkorstanje:
> I thought we agreed that issues like this shouldn't leak into many classes.

i don't believe it does. It only affects runtime, the runner supplier and the event bus. The main loop in Runtime could look like:

```java
        ExecutorService executor = Executors.newFixedThreadPool(5);
        
        for (CucumberFeature cucumberFeature : features) {
            for (final PickleEvent pickleEvent : compiler.compileFeature(cucumberFeature)) {
                    executor.submit(new Runnable() {
                        @Override
                        public void run() {
                        }
                    });
                }
            }
        }
```

mpkorstanje:
> I'm not 100% sure why it should be calling super.send(event) for all events as well as adding them to the queue to then be sent again later once TestCaseFinished is received

We have two types of listeners to deal with. 
* JUnit which listens to individual pickles in isolation. These events shouldn't be queued up. Queuing them up will cause JUnit to think these tests completed instantly.
* Formatters/Plugins. These need to get their events as an synchronized batch per pickle because the events lack the context to associated them to a pickle when multiple pickles are executed concurrently.

 Junit ---listens to--->  BatchEventBus ---sends batches of events to--> EventBus  <---listens too--- HTML formatter

mpkorstanje:
That said. `BatchEventBus` is a terrible name. If you know a better one please let me know!

boaty82:
PR replaced by [Parallel cukes #1389](https://github.com/cucumber/cucumber-jvm/pull/1389)

Issue #1366, opened 2018-05-20
Refactor RuntimeOptions.cucumberFeatures out of RuntimeOptions


mpkorstanje:
Currently `RuntimeOptions` has too many responsibilities. In addition to managing the runtime options:
 * It parses command line options
 * It manages and instantiates plugins
 * It loads and instantiates cucumberFeatures

Ultimately `RuntimeOptions` should become an immutable data class that implements several different interfaces to be used by different consumers.

Solving this can be done in several steps. This is one such step:

`RuntimeOptions.cucumberFeatures` and related functions should be in their own `CucumberFeatures` class. The responsibility of emitting the `TestRunStarted` should be moved to the callers of `cucumberFeatures`.

So ultimately the dependency graph will be : 

```
RuntimeOptions <--- CucumberFeatures <--- Runtime (use constructor dependency injection)
EventBus       <---|
ResourceLoader <---|
```

aslakhellesoy:
Sounds like a good plan. Do you think we could fit in [config](https://github.com/cucumber/cucumber/tree/master/config) to populate `RuntimeOptions` with values @mpkorstanje?

mpkorstanje:
I'm making theses issues really small so any one can pick them up. Using config to populate run time options would be a separate task. There are a few more tasks that involve trimming down RuntimeOptions. After this is done implementing config would be easier.

mpkorstanje:
Closed by 56c34773f9a5b3599b3f842fa49b13545db38204

Issue #1352, opened 2018-04-23
 Refactor Runtime.filters and matchesFilters out of Runtime


mpkorstanje:
To make the first baby steps towards implementing a Cucumber Runner for JUnit 5 we'll need to make it possible to filter a list of features without creating a `Runtime` instance.

Currently `Runtime` assembles `Runtime.filters` internally from `RuntimeOptions` but only accesses it through `Runtime.matchesFilters`. This should make it possible to extract `Runtime.filters` and `Runtime.matchesFilters` in to Filters class.

It'd be also useful to create a `Filters.getFilters()` class that  handles the creation of a list of filter. This is currently done in Runtime through.

```java
        this.filters = new ArrayList<PicklePredicate>();
        List<String> tagFilters = runtimeOptions.getTagFilters();
        if (!tagFilters.isEmpty()) {
            this.filters.add(new TagPredicate(tagFilters));
        }
        List<Pattern> nameFilters = runtimeOptions.getNameFilters();
        if (!nameFilters.isEmpty()) {
            this.filters.add(new NamePredicate(nameFilters));
        }
        Map<String, List<Long>> lineFilters = runtimeOptions.getLineFilters(resourceLoader);
        if (!lineFilters.isEmpty()) {
            this.filters.add(new LinePredicate(lineFilters));
        }
```

Desired dependency graph: 

```
RuntimeOptions <--- Filters <--- Runtime (use constructor dependency injection)
```

mpkorstanje:
This partially been done by 56c34773f9a5b3599b3f842fa49b13545db38204

However `addLineFilters` is duplicated a few times and suffers from a primitive obsession.

Issue #1149, opened 2017-06-13
Add JUnit 5 Support


reinhapa:
## Summary
Provide a JUnit 5 extension for cucumber.

## Expected Behavior
When writing a test with JUnit 5 in our environment, we need to replicate a base environment in a way that we use different Extensions in order to to run tests. So we like  to have a Scenario something like this:

~~~
@ExtendWith(CucumberExtension.class)
@ExtendWith(MockitoExtension.class)
@ExtendWith(BaseBusinessEnvironmentExtension.class)
class MyBusinessTest {
   @BeforeEach
    void init(@Mock Person person) {
        when(person.getName()).thenReturn("Dilbert");
    }

    @Test
    void simpleTestWithInjectedMock(@Mock Person person) {
        assertEquals("Dilbert", person.getName());
    }
}
~~~

## Current Behavior
At the moment under JUnit 4 we needed to tweak the Cucumber runner in order to call default JUnit 4 annotations.

## Possible Solution
We like to help create an CucumberExtension that handles the Cucumber specific annotations for JUnit 5  as an JUnit 5 extension, so that it can be combined with other extensions as for example mocking frameworks.

mpkorstanje:
Awesome! I don't have any experience with JUnit5. Perhaps you can use #1103 as a starting point, see if it builds, meets your needs and review the code.

reinhapa:
I will take a look into it

reinhapa:
@mpkorstanje I looked into it briefly and it would be great to open a feature branch for this extension. I will try to contact @tenwit to see if I can help furter extend/test it...

mpkorstanje:
Excellent! If you need help with anything let me know. I'll do what I can.

reinhapa:
@mpkorstanje I just trying to merge the the current master and the existing pull request into my local repo right know. I will come back to you as soon I got it working.

tenwit:
I'm here :) How can I help? My solution works using extensions; I see there's an alternative solution using the engine at https://github.com/signed/junit-cucumber-engine. I hadn't considered using an engine when I did my work. I think it's more idiomatic to use an engine.

reinhapa:
[...]

PR #1367, opened 2018-05-22
[Core] Refactor Runtime

mpkorstanje:
## Summary

Extracting:
 * Backend creation, 
 * Glue creation,  
 * Runner creation
 * Feature compilation 

from the runtime allows  
 * Tests on runners to skip the creation of Runtime
 * Makes extracting Filter provider(#1352) and  Feature provider(#1366) easier.
 * Will allow other runtimes such as JUnit and TestNg  to skip the creation of the runtime
 * Will facilitate parallel (#1357), junit 5 (#1149) and  pickle runner support.
 * Clarifies the execution loop of Cucumber


boaty82:
Pulled the branch to work on some changes for [PR 1357](https://github.com/cucumber/cucumber-jvm/pull/1357) and before I've made any changes I'm running the tests (whilst perusing the changes already made for me) and I get the following issue

```
[ERROR] Failed to execute goal org.revapi:revapi-maven-plugin:0.10.2:check (check) on project cucumber-java: The following API problems caused the build to fail:
[ERROR] java.annotation.noLongerPresent: @interface cucumber.api.java.ht.SipozeKe: Old API referenced the annotation type but the new API version no longer does. (breaks semantic versioning)
[ERROR] java.annotation.noLongerPresent: @interface cucumber.api.java.ht.SipozeKe: Old API referenced the annotation type but the new API version no longer does. (breaks semantic versioning)
[ERROR]
[ERROR] If you're using the semver-ignore extension, update your module's version to one compatible with the current changes (e.g. mvn package revapi:update-versions). If you want to explicitly ignore this change and provide a justification for it, add the following JSON snippet to your Revapi configuration under "revapi.ignore" path:
[ERROR] {
[ERROR] "code": "java.annotation.noLongerPresent",
[ERROR] "old": "@interface cucumber.api.java.ht.SipozeKe",
[ERROR] "package": "cucumber.api.java.ht",
[ERROR] "classQualifiedName": "cucumber.api.java.ht.SipozeKe",
[ERROR] "classSimpleName": "SipozeKe",
[ERROR] "elementKind": "@interface",
[ERROR] "justification": <<<<< ADD YOUR EXPLANATION FOR THE NECESSITY OF THIS CHANGE >>>>>
[ERROR] },
...
```

mpkorstanje:
Hey, feel free to comment out the revapi plugin. If you comment out the one
in the plugins section, not the plugin management section it should be
easy.

Not sure what is causing this. Looks like the languages aren't generated.

boaty82:
I’ve done that. Just thought there might be a more “proper” fix

mpkorstanje:
No. Don't have a solution yet. There is something going wrong when you run the build. Maybe something executed in the wrong order that causes the languages not to be generated.

mpkorstanje:
I've moved the semantic version check to a different profile. This slows down the feedback loop a bit, but this is okay because we'll figure out which version it has to go in when merging. Not when building the feature.

mpkorstanje:
Also feel free to ignore any problems with `pax-exam`.

boaty82:
I've found a nice solution.

I've created a new event `TimerEvent`, which `StepDurationTimeService` now registers for. 
 Updated `TestStepStarted` to implement it

I then had to update `EventBus` to loop through the handler map and check if the incoming event is an instance of the class in the key, rather than getting the key.

mpkorstanje:
Hey Boaty,

I'd rather not introduce a new event. I don't think it is needed.

The current tests for the formatters are setup such that cucumber generates
a stream of events. For tests we mock the time service so we can tick the
clock manually and get predictable timestamps.

The generation of these events is currently closely tied to the rest of the
implementation and the tests. This was okay in the past. There was only one
ordering of events. Now we have multiple orderings to deal with.

Generating these in a controlled fashion requires first that we make the
concept of an event stream explicit.

An event stream is a sequence of events with some ordering:

[TestStarted, TestSourceRead, {[ScenarioStarted, Step1, Step2,
ScenarioFinished]*}, TestFinished]

[This is in order], {these can come in any order}, these* can repeat.

So by recording a regular event stream and shuffling the groups of scenario
events we can simulate the result of a concurrent execution.

This might also make the implementation of formatters simpler. Rather then
trying to make all the formatters smart, we can record the events and
replay them in the right order (sort the groups of scenario events by uri).

-MP

PS: Can't look at the code. On a boat.

boaty82:
Hi.

Haven't needed to add a new event interface in the end.  Managed to just create a new EventBus implementation now, purely for tests, so `TestStepStarted` events are sent for only certain `EventListener`s - in my case `StepDurationTimeService`

Have done the changes to support `--threads`, just adding the thread timeline report now.  Do the HTML/JS/etc files need to go in a new jar & repo link the HTML plugin?  I won't have permission to do this

PR #1368, opened 2018-05-22
Upgrade cucumber expressions to 5.0.18

mpkorstanje:
## Summary

Upgrade cucumber expressions to 5.0.18. This resolves https://github.com/cucumber/cucumber/issues/391for cucumber-jvm.


aslakhellesoy:
LGTM - I think small fixes this should be done on master to reduce the process overhead.

https://trunkbaseddevelopment.com/

PR #1377, opened 2018-05-30
Upgrade gherkin and cucumber expressions

aslakhellesoy:
Upgrade to latest versions


mpkorstanje:
I don't expect either change to break much in practice. However that is not what semantic versioning is about. We do have  `--parallel` in the works though. That will definitely break the API (events are no longer ordered by feature), so we might as well merge this along with that.

PR #1379, opened 2018-06-01
Add pending tag support to failing scenarios to prevent build fail

Heziode:
## Summary

This feature all users to use `@pending` tag to not fail a build.

## Details

Imagine you have a bug in your project. The bug are reported but, you can't fix it imediatly due to roadmap, or anything else.

With this feature, you can add a `@pending` tag before your scenario (or feature, or hook) to continue to pass the build, because `pending` status does not have an effect on exit status, and the failing test continues to appear in your report (to not forget it).

Example:

```gherkin
Feature: Go on moon
	
	@pending
	Scenario: Launch a rocket
	When I click on button launch
	Then The rocket take off
```

Java side:

```java
@When("I click on button launch")
public void I_click_on_button_launch() {
    throw new Exception("Your are not yet ready");
}
```

This example will not fail the test suit, instead, this scenario have a pending result.

## Motivation and Context

This feature is very useful when you work in team on project with built pipeline (automated build, etc.) and you (and other developers) don't have the time to fix the bug quickly.

## How Has This Been Tested?

The tests check if we run a scenario with a `@pending` tag, it returns the `PENDING` status if the a step throws an error, and `PASSED` otherwise.

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/runner/TestStep.java
+++ b/core/src/main/java/cucumber/runner/TestStep.java
@@ -42,7 +42,7 @@ abstract class TestStep implements cucumber.api.TestStep {
             status = executeStep(language, scenario, skipSteps);
         } catch (Throwable t) {
             error = t;
-            status = mapThrowableToStatus(t);
+            status = mapThrowableToStatus(t, scenario.getSourceTagNames().contains("@pending"));
         }
         Long stopTime = bus.getTime();
         Result result = mapStatusToResult(status, error, stopTime - startTime);
@@ -60,8 +60,8 @@ abstract class TestStep implements cucumber.api.TestStep {
         }
     }
 
-    private Result.Type mapThrowableToStatus(Throwable t) {
-        if (t.getClass().isAnnotationPresent(Pending.class)) {
+    private Result.Type mapThrowableToStatus(Throwable t, boolean isPending) {
+        if (t.getClass().isAnnotationPresent(Pending.class) || isPending) {
             return Result.Type.PENDING;
         }
         if (Arrays.binarySearch(ASSUMPTION_VIOLATED_EXCEPTIONS, t.getClass().getName()) >= 0) {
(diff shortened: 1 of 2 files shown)


Heziode:
> - [ ] My change requires a change to the documentation.
> - [ ] I have updated the documentation accordingly.

What kind of documentation do you mean ? JavaDoc or [Cucumber doc](https://docs.cucumber.io/) ?

aslakhellesoy:
I understand the problem you're trying to solve, but I'm not sure I agree with the implementation.

First of all, we don't want to impose any semantics around predefined tags. This makes too many assumptions about the language in use, and also what the tags mean.

There are two alternative solutions:

## Explicit Work In Progress (WIP)

Cucumber-Ruby has a `--wip` command line option that causes Cucumber to exit with non-zero exit status if any scenarios *pass*.

Assuming your scenarios are tagged with `@pending`, you can have two runs: One that uses `--tags "not @pending"` and another one with `--wip --tags "@pending"`.

We could (and should, I think) implement `--wip`.

## Custom Strictness

The `--strict` option tells Cucumber to treat undefined and pending steps as errors. The default is `--no-strict`, so I am a little puzzled why your scenarios are failing. By default, pending steps should not be treated as errors.

Heziode:
Java does not have the `--wip` ?
I have do this modification for that.

aslakhellesoy:
I'm suggesting you create a new PR with support for `--wip`. That would be a great addition to Cucumber-JVM!

But did you try to turn off `--strict` mode?

Heziode:
> I'm suggesting you create a new PR with support for --wip. That would be a great addition to Cucumber-JVM!

Yes, I will do that.

> But did you try to turn off --strict mode?
The scenario that fails are already implemented it throw an exception due to a bad implementation, they are not undifined or pending.

PR #1381, opened 2018-06-02
[Core] Add --wip option

Heziode:
## Summary

This feature is the java implementation of `--wip` of Cucumber-Ruby. But, instead of not support `--wip` with `--strict` (in Ruby version), this version do it

## How Has This Been Tested?

I have do tests in:
- Result
- RuntimeOptions
- Runtime

Test in ExitStatus are not relevant here because it is tested in Runtime, isn't it ?

## Types of changes

## Checklist:


core/src/main/java/cucumber/runtime/ExitStatus.java
```diff
@@ -24,7 +25,8 @@ public void receive(TestCaseFinished event) {
         }
     };
 
-    ExitStatus() {
+    ExitStatus(boolean wip) {
```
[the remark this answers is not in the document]
Heziode: It is done

core/src/main/java/cucumber/api/Result.java
```diff
@@ -78,11 +78,15 @@ public boolean is(Result.Type status) {
...
 
     public boolean isOk(boolean isStrict) {
-        return hasAlwaysOkStatus() || !isStrict && hasOkWhenNotStrictStatus();
+        return hasAlwaysOkStatus(false) || !isStrict && hasOkWhenNotStrictStatus();
     }
 
-    private boolean hasAlwaysOkStatus() {
-        return is(Result.Type.PASSED) || is(Result.Type.SKIPPED);
+    public boolean isOkWip(boolean isStrict) {
+        return hasAlwaysOkStatus(true) || !isStrict && hasOkWhenNotStrictStatus();
```
mpkorstanje: I don't think this method is needed. 

When using  `--wip` the build should fail  when there are any passing results (when result.isOk(true/false) return true).

core/src/main/java/cucumber/runtime/ExitStatus.java
```diff
@@ -24,16 +25,18 @@ public void receive(TestCaseFinished event) {
...
     public void setEventPublisher(EventPublisher publisher) {
         publisher.registerHandlerFor(TestCaseFinished.class, testCaseFinishedHandler);
     }
 
-    public byte exitStatus(boolean isStrict) {
-        return results.isEmpty() || max(results, SEVERITY).isOk(isStrict) ? 0x0 : ERRORS;
+    public byte exitStatus() {
+        return results.isEmpty() ||
+            (runtimeOptions.isWip() ? max(results, SEVERITY).isOkWip(runtimeOptions.isStrict()) :
+                max(results, SEVERITY).isOk(runtimeOptions.isStrict())) ? 0x0 : ERRORS;
```
mpkorstanje: It's okay to use if statements here and use [a guard clause](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) to handle the empty case.

I also think this will fail when there is one test case that passes and one test case that failed.

mpkorstanje: Can you add a test case to prove it works?

Heziode: Yep, I can add test to check this, but… where I write it ?

mpkorstanje: In runtime test. But you put more then one event on the bus.

core/src/main/java/cucumber/api/Result.java
Heziode: @mpkorstanje If I only use isOk, and I do what you tell, I obtain an unexpected behaviour with strict mode.

With wip and strict, if I have an undefined step, the result is 0 instead of 1.

core/src/main/java/cucumber/runtime/ExitStatus.java
Heziode: It is (normally) fixed with 99e535e

core/src/main/java/cucumber/api/Result.java
mpkorstanje: I thought it wasn't possible to use wip and strict in together. But if it is, that makes sense. An undefined result in strict mode makes the test fail. So when running with wip and strict  it should pass. That is what I get from reading [work_in_progress.feature](https://github.com/cucumber/cucumber-ruby/blob/d73e337a5d347313984f811e8c9b139e74c06551/features/docs/work_in_progress.feature):

```
  In order to ensure that feature scenarios do not pass until they are expected to
  Developers should be able to run cucumber in a mode that
            - will fail if any scenario passes completely
            - will not fail otherwise
```

Heziode: Ok, do you want I disable the use of strict with wip ?

mpkorstanje: Why do you think the behaviour is unexpected?

aslakhellesoy: I mentioned somewhere else that I couldn’t remember why we disallowe strict and wip at the same time.

If we can’t think of a good reason for it (I can’t) we should allow it.

mpkorstanje: I am thinking the same thing. But I am trying to work out why Heziode thinks the result of combining wip and strict with an undefined result is unexpected.

Heziode: Here is a result I want expect:

- `--wip`: Test fail only if one or more scenario have status `PASSED`
- `--wip` + `--strict`: Test fail if one or more scenario have status `PASSED`, `UNDEFINED`, `SKIPPED`

What do you think about this ?

mpkorstanje:
Well I made a table. I used the [work_in_progress.feature](https://github.com/cucumber/cucumber-ruby/blob/d73e337a5d347313984f811e8c9b139e74c06551/features/docs/work_in_progress.feature) for the `--wip, --no-strict` column, the `--no-wip` columns are how cucumber currently works. The `--wip --strict` column is just the two obvious ones.

```feature
           |       --wip            |       --no-wip         |
           | --strict | --no-strict | --strict | --no-strict | 
PASSED     |   fail   |  fail       |   pass   |  pass       |
SKIPPED    |    ?     |  ?          |   pass   |  pass       |
PENDING    |    ?     |  pass       |   fail   |  pass       |
UNDEFINED  |    ?     |  pass       |   fail   |  pass       |
AMBIGUOUS  |    ?     |  pass       |   fail   |  fail       |
FAILED     |  pass    |  pass       |   fail   |  fail       |
```

The question mark at `--wip  --no-strict` is okay because as far as I know you can't skip steps in ruby without making a prior step fail.

You can however skip steps in java using `Assume`. So we have two options to fill out that one question mark:

A:
```feature
           |       --wip            |       --no-wip         |
           | --strict | --no-strict | --strict | --no-strict | 
PASSED     |   fail   |  fail       |   pass   |  pass       |
SKIPPED    |    ?     |  fail       |   pass   |  pass       |
PENDING    |    ?     |  pass       |   fail   |  pass       |
UNDEFINED  |    ?     |  pass       |   fail   |  pass       |
AMBIGUOUS  |    ?     |  pass       |   fail   |  fail       |
FAILED     |  pass    |  pass       |   fail   |  fail       |
```
Interesting here is that  `--wip  --no-strict` is the inverse of  `--no-wip  --strict`. Which would explain why wip and strict modes can't be used together. The wip mode is already strict (and inversed).

B:
```feature
           |       --wip            |       --no-wip         |
           | --strict | --no-strict | --strict | --no-strict | 
PASSED     |   fail   |  fail       |   pass   |  pass       |
SKIPPED    |    ?     |  pass       |   pass   |  pass       |
PENDING    |    ?     |  pass       |   fail   |  pass       |
UNDEFINED  |    ?     |  pass       |   fail   |  pass       |
AMBIGUOUS  |    ?     |  pass       |   fail   |  fail       |
FAILED     |  pass    |  pass       |   fail   |  fail       |
```
You can also make an argument for this. When work is in progress you wouldn't expect a the build to fail because you made a scenario pass by skipping it entirely.

And again the wip mode more stricter then strict (and inverted).

Either way it seems that `--no-strict` in combination with `--wip` makes no sense.

@aslakhellesoy what kind of work flow was wip supposed to support?

Heziode:
@mpkorstanje The option **B2** is what I expect

mpkorstanje:
Let's go for option B2 then. Basically just ignore strict when wip is active.

```feature
           |       --wip            |       --no-wip         |
           | --strict | --no-strict | --strict | --no-strict | 
PASSED     |   fail   |  fail       |   pass   |  pass       |
SKIPPED    |   pass   |  pass       |   pass   |  pass       |
PENDING    |   pass   |  pass       |   fail   |  pass       |
UNDEFINED  |   pass   |  pass       |   fail   |  pass       |
AMBIGUOUS  |   pass   |  pass       |   fail   |  fail       |
FAILED     |   pass   |  pass       |   fail   |  fail       |
```

A very concise way to check if the least severe result is a passed would be:  `min(results, SEVERITY).is(Type.PASSED);` so the wip methods in result aren't needed.

Heziode:
It is done.

mpkorstanje:
Looks good. I'll try to merge this and the runtime constructors branch later this week.

PR #1398, opened 2018-06-16
[Core] Use fully classified name for PendingException

### LINKED ISSUE #1383: Use fully-qualified class names in snippets
Snippets don't qualify `PendingException`, and this often causes people to import the wrong class, or simply be confused about what class to use.

We should just make it fully qualified.

mlvandijk:
## Summary
Use fully classified name for PendingException.

## Motivation and Context
See #1383

## How Has This Been Tested?
Updated tests and ran `mvn clean install`

## Types of changes

## Checklist:

- [x] I have updated the tests accordingly.


mpkorstanje:
This doesn't fix all of #1383 but it's a good PR (atomic, improvement) nonetheless.

PR #1389, opened 2018-06-09
[Core] Parallel cukes

boaty82:
## Notice
This PR replaces [Parallel support #1357](https://github.com/cucumber/cucumber-jvm/pull/1357) and relies on [Refactor Runtime #1367](https://github.com/cucumber/cucumber-jvm/pull/1367)

## Outline

I’ve previously used another implementation to deliver parallel execution of cucumber tests, however it basically wrapped up the creation and execution of `Runtime`’s, followed by the need to merge the multiple JSON/HTML etc reports that followed.

I reviewed the current implementation and also took note of comments in [Support concurrent executions of scenarios](https://github.com/cucumber/cucumber-jvm/issues/630) also I perused all tickets related to running in [parallel cukes](https://github.com/cucumber/cucumber-jvm/issues?utf8=%E2%9C%93&q=is%3Aissue+parallel) for other insights

### Added `--threads` argument to runtime options 
Allows users to specify the (max) number of threads to be used to run the tests

The actual number of threads used will be the smaller of the given `--threads` value and the number of features to be executed

Tests are placed into a `Queue` and consumed by each thread when needed, resulting in a more performant usage of the threads & less deterministic order of Features being run

### New formatter introduced `TimelineFormatter` 
Which produces reports using [vsjis.org timeline](http://visjs.org/examples/timeline/groups/groups.html) to highlight which feature was run on which Thread and when.  
**Note:** the resources (js, html etc) are currently within `cucumber-core` if the PR is accepted then I’d imagine these would have to be moved out into its own project, similarly to cucumber-html.
timelineformatter
[example.zip](https://github.com/cucumber/cucumber-jvm/files/2132738/example.zip)

## Types of changes
- [ ]  Bug fix (non-breaking change which fixes an issue).
- [x]  New feature (non-breaking change which adds functionality).
- [x]  Breaking change (fix or feature that would cause existing functionality to not work as expected).
 - `Runtime` constructor now private `Runtime.Builder` introduce to hide away boiler plate setup/instantiation code

## Checklist:
- [x]  I've added tests for my code.
- [x]  My change requires a change to the documentation.
- [ ]  I have updated the documentation accordingly.

PR #1402, opened 2018-06-21
WIP: Use Gherkin 6 / Remove rich formatters.

aslakhellesoy:
Removed JSON/HTML/Pretty/JUnit XML formatters and related cruft.

See:
* https://github.com/cucumber/cucumber/pull/416
* https://github.com/cucumber/cucumber/issues/415

This is a big change, and it may conflict with #1389.


mpkorstanje:
Had a quick look. I am seeing allot of changes in imports that only seem to change the order.

aslakhellesoy:
Is that going to be a problem? I use IDEA optimise imports a lot.

mpkorstanje:
Merge conflicts I imagine. We could run optimize imports on both branches though and hope it works out though.

aslakhellesoy:
This branch is stale, but let's keep it around. It's useful to look at some of the changes when we give this another try.

PR #1404, opened 2018-06-27
[OSGi] Remove OSGi

aslakhellesoy:
## Summary

Remove `osgi` module and `pax-exam` examples. Don't package jars as OSGi bundles.

Ref cucumber/cucumber#412

## Motivation and Context

All the OSGi stuff complicates the codebase and build system significantly.
People who want to maintain this are invited to create new projects outside of the `cucumber-jvm` repo, in a similar vein to `cucumber-scala` etc. See https://cucumber.io/blog/2018/05/19/cucumber-jvm-languages-support for more details.


aslakhellesoy:
Yeah https://github.com/cucumber/cucumber/issues/412 has a checklist. Not sure if I forgot something...

mpkorstanje:
I don't think so. 

We'll draw in the dependencies over time as we upgrade them. So I'll merge this now.

aslakhellesoy:
Just a note for anyone who is sad OSGi is gone. If you want OSGi support to come back, all you need to do is to dig it out of git, set up a standalone GitHub project and ask us to move it to the `cucumber` organisation. You would be responsible for maintaining it and making releases.

/cc @rjking99 @HendrikSP

HendrikSP:
How should this work? You removed all bundle descriptions. One who would have to build each jar file in cucumber/cucumber#412 as OSGi bundles again. But of course that is not feasible. So you have to rename the maven coordinates to something else which would be a complete fork of this project. I still don't get why you had to remove the OSGi meta-information. removing the pax exam stuff should have been enough

aslakhellesoy:
Most java libraries in the wild are not packaged as OSGi bundles. What do people do when they want to use those libraries in an OSGi application?

HendrikSP:
"Most"? I don have any numbers. But I would say the most important projects have OSGi Metadata in their jars. Please don't count Spring here, I won't work well with OSGi anyway. ;)

aslakhellesoy:
There are nearly [20k](https://github.com/search?q=OSGI+bundle&type=Issues) open issues on GitHub about adding OSGi Metadata to various libraries.

Surely there must be several ways you can add OSGi Metadata to 3rd party jars as part of your own build process?

d-led:
If necessary, I'd go with this: [Gradle OSGi Plugin](https://docs.gradle.org/current/userguide/osgi_plugin.html). It could be used in creative ways: re-package a JAR, or simply use Gradle during build & use the plugin

HendrikSP:
Yes, there are, but they are more based on guessing. The bundler does cost you nothing and does all the required steps you need. The only thing that you can break in an OSGi way is bad class loader usage. If it breaks in an OSGi container, most of times is bad implementation and/or design anyway.

HendrikSP:
@mpkorstanje What is the problem for this project that it fails in an OSGi container or the shade plugin? If it is the shade plugin, I guess the problem is due to the package renaming, in this case bnd can be configured easily to export only the new packages. If it's the OSGi container, any OSGi user may provide a PR, I guess that is not much work, or do I miss something?

mpkorstanje:
The java part of this project has 2-3 people working on it in their spare time.

I'd rather spend that time working on actual problems/improvements rather then dealing with OSGi, its PRs and extra releases these create.

I'd rather see that OSGi users solve the problem of dealing with non-OSGi dependencies.

PR #1407, opened 2018-06-28
[Core] Deprecate Formatter interface

### LINKED ISSUE #1401: [Core] Deprecate the Formatter interface
They formatter interface is a marker interface that used to inform Cucumber that the class implementing it was a plug-in that should be provided with events. This functionality has since fallen apart into `Plugin` and `EventListener`. 

All usages of `Formatter` can be replaced with `Plugin` and `EventListener`. The class itself can be deprecated.

mlvandijk:
## Summary

Deprecate Formatter marker interface

## Details

Deprecate Formatter marker interface and replace all usages of Formatter with Plugin and EventListener.

## Motivation and Context

Fixes #1401 

## How Has This Been Tested?
`mvn clean install`

## Types of changes

Not sure

## Checklist:

- [x] I have updated the tests accordingly.


core/src/main/java/cucumber/api/formatter/ColorAware.java
```diff
@@ -1,9 +1,11 @@
 package cucumber.api.formatter;
 
+import cucumber.api.Plugin;
+
 /**
- * Interface for Formatters that use ANSI escape codes to print coloured output.
+ * Interface for Plugins that use ANSI escape codes to print coloured output.
  */
-public interface ColorAware extends Formatter {
+public interface ColorAware extends Plugin {
```
mpkorstanje: Nice!

core/src/main/java/cucumber/api/formatter/Formatter.java
```diff
@@ -4,11 +4,8 @@
 import cucumber.api.event.EventListener;
 
 /**
- * This is the interface you should implement if you want your own custom
- * formatter.
- *
- * @see EventListener
- * @see Plugin
+ * @deprecated as of version 4.0.0; use {@link Plugin } or {@link EventListener } instead.
```
mpkorstanje: Is this "or" or "and"?

core/src/main/java/cucumber/api/formatter/StrictAware.java
```diff
@@ -1,9 +1,11 @@
 package cucumber.api.formatter;
 
+import cucumber.api.Plugin;
+
 /**
- * Interface for Formatters that need to know if the Runtime is strict.
+ * Interface for Plugins that need to know if the Runtime is strict.
  */
-public interface StrictAware extends Formatter {
+public interface StrictAware extends Plugin {
```
mpkorstanje: Do you think it would be a good idea for all marker interfaces to extend from Plugin?

core/src/main/java/cucumber/api/formatter/Formatter.java
mlvandijk: Should be both, although sometimes ColorAware and/or StrictAware instead of Plugin.

core/src/main/java/cucumber/api/formatter/StrictAware.java
mlvandijk: Not sure what you mean; did I miss anything?

core/src/main/java/cucumber/runtime/RuntimeOptionsFactory.java
```diff
@@ -86,7 +86,7 @@ private void addPlugins(CucumberOptions options, List<String> args) {
         for (String plugin : plugins) {
             args.add("--plugin");
             args.add(plugin);
-            if (PluginFactory.isFormatterName(plugin)) {
+            if (PluginFactory.isPluginName(plugin)) {
                 pluginSpecified = true;
```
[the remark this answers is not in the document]
mlvandijk: Will try :)

mlvandijk:
Some updates, based on review. Will need to look at first comment later.
A question of my own:
- I've updated the Javadoc for ColorAware/StrictAware to mention plugins instead of formatters. Should names be updated elsewhere/verywhere? (i.e. package "formatter", several "SomethingFormatter" classes)?

core/src/main/java/cucumber/api/formatter/StrictAware.java
mpkorstanje: The EventListner I believe

mpkorstanje:
Ideally yes, but between parallel cukes and Aslak upgrading gherkin this will be painfull.

core/src/main/java/cucumber/api/formatter/StrictAware.java
mlvandijk: EventListener is not a marker interface; it has a method: `void setEventPublisher(EventPublisher publisher);`

mpkorstanje: Good one. 

Would it be  a good idea to make EventListener extend from Plugin?

mlvandijk: Yes. (Updated)

mlvandijk:
Ok, so let's do the renaming later (after those other PRs are merged?).

core/src/main/java/cucumber/runtime/RuntimeOptionsFactory.java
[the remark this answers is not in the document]
mlvandijk: Removing the if statement, changes the result of `RuntimeOptionsFactoryTest.create_null_formatter_when_no_formatter_plugin_is_defined()` so that "cucumber.runtime.formatter.AnyStepDefinitionReporter" plugin is used, rather than "cucumber.runtime.formatter.NullFormatter".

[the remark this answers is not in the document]
mpkorstanje: Okay. So now the null formatter is gone (replaced by no formatter), this test should be changed to check if no formatter is there instead.

PR #1442, opened 2018-08-28
[JUnit,TestNG] Update documentation

mpkorstanje:
## Summary
Document how to enable parallel execution for JUnit and TestNG

### DIFF
--- a/testng/README.md
+++ b/testng/README.md
@@ -35,7 +35,6 @@ This will execute all scenarios in same package as the runner, by default glue c
 package. The `@CucumberOptions` can be used to provide
 [additional configuration](https://cucumber.io/docs/reference/jvm#list-configuration-options) to the runner. 
 
-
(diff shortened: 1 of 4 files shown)


testng/README.md
```diff
@@ -48,4 +47,19 @@ Cucumber provides limited support for [SkipException](https://jitpack.io/com/git
...
+
+## Parallel execution ##
+
+Cucumber TestNG supports parallel execution of scenarios. Override the `scenarios` method to enable parallel execution.
+
+```java
+public class RunCukesTest extends AbstractTestNGCucumberTests {
+
+    @Override
+    @DataProvider(parallel = true)
```
mpkorstanje: @kathyrollo you've been using TestNG more then I have. Is this the only way to make tests  with a a data provider run in parallel? Do you happen to know if we can control this from maven instead?

PR #1451, opened 2018-09-07
Bugfix/v4.x jpms

nhojpatrick:
## Summary

Define Automatic-Module-Names for all jar files generated so that they can be used within the Java Platform Module System, instead of being put into the UNNAMED bucket.

## Details

Change only the build so each jar created has a 'Automatic-Module-Name' added into the META-INFO/MANIFEST.MF file.

Also add maven wrapper as util tool so project decides what version of maven to use for building instead of each develop having to choose or download a specific version.

## Motivation and Context

Trying to write some cucumber frameworks (state between steps, object to json conversion, json manipulation, calling rest endpoints sending/receiving json), as i rewrite this type of thing at each client i work for so want to write it once and feed back to cucumber as extra so when i start my next contract i can reuse instead of write from scratch.

I'm trying to write a pure java 9/10/11 build and getting issues with the framework I've written as they are UNNAMED modules, or modules based upon the jar file name which isn't a good idea.

## How Has This Been Tested?

No code changed, only jar file manifest altered.

## Types of changes

## Checklist:

Not changed documentation about switching to building using Maven Wrapper.


continuous-deployment/pom.xml
```diff
@@ -40,4 +40,7 @@
         </plugins>
     </build>
 
+    <properties>
+        <project.Automatic-Module-Name>io.cucumber.continuous.deployment</project.Automatic-Module-Name>
```
mpkorstanje: This module is excluded from release. It doesn't need the module name.

kotlin-java8/pom.xml
```diff
@@ -13,6 +13,7 @@
 
     <properties>
         <kotlin.version>1.2.41</kotlin.version>
+        <project.Automatic-Module-Name>io.cucumber.kotlin.java8</project.Automatic-Module-Name>
```
mpkorstanje: This module is excluded from release. It doesn't need the module name.

pom.xml
```diff
@@ -84,6 +84,7 @@
         <typetools.version>0.5.0</typetools.version>
         <cucumber-expressions.version>6.0.1</cucumber-expressions.version>
         <datatable.version>1.1.3</datatable.version>
+        <project.Automatic-Module-Name>io.cucumber.jvm</project.Automatic-Module-Name>
```
mpkorstanje: The pom is not released as a jar. So I don't think it needs a module name.

mpkorstanje:
> Also add maven wrapper as util tool so project decides what version of maven to use for building instead of each develop having to choose or download a specific version.

While useful I would rather not include compiled files through a PR. They are non trivial to verify and inspect.

pom.xml
nhojpatrick: if the default changes to <project.Automatic-Module-Name></project.Automatic-Module-Name> then your other comments can simply have the property removed, from memory the property needs to be defined as maven-jar-plugin can't allow conditional options so for these jars add this extra configuration

continuous-deployment/pom.xml
nhojpatrick: changed

kotlin-java8/pom.xml
nhojpatrick: removed

pom.xml
nhojpatrick: made default empty property so that koltin-java8 and continious-deployment also pick up that property

mpkorstanje:
I am doing some catching up on the module system:

http://branchandbound.net/blog/java/2017/12/automatic-module-name/
> ### Sanity-Check Your Library
>
>Is this really all there is to do as a first step toward modularization? Ideally, yes. But if you want your library to be used as automatic module on Java 9 and later, there's a few other potential issues you need to verify:
>
>  * Make sure your library doesn't use internal types from the JDK (run jdeps --jdk-internals mylibrary.jar to find offending code). JDeps (as bundled with Java 9 and later) will offer publicly supported alternatives for any use of encapsulated JDK APIs. When your library runs from the classpath on Java 9 and later, you can still get away with this. Not so if your library lives on the module path as automatic module.
>  * Your library can't have classes in the default (unnamed) package. This is a bad idea regardless, but when your library is used as automatic module, this rule is enforced by the module system.
> * Your library can't split packages (two or more JARs defining the types in the same package), nor can it redefine JDK packages (javax.annotation is a notorious example, being defined in the JDK's java.xml.ws.annotation module but also in external libraries).
> * When your library's JAR has a META-INF/services directory to specify service providers, then the specified providers must exist in this JAR (as described in the ModuleFinder JavaDoc)

> Addressing these concerns is a matter of good hygiene. If you encounter one of these issues and you can't address those, don't add the Automatic-Module-Name entry yet. For now, your library is better off on the classpath. It will only raise false expectations if you do add the entry.

We currently have split packages so we can't do this yet. 

In fact we're using a split package in cucumber-spring to work around a specific problem with the spring text context not being thread-safe.

mpkorstanje:
Merged by 0987a74687290d81d75d3b1cbb5c2a4c0d754cca

mpkorstanje:
I've merged this branch with the develop v5 branch. Couldn't do it via github because it had conflicts and I wanted to squash the commits in it. The commit has been properly attributed so that should be fine.

PR #1452, opened 2018-09-08
Highlight arguments in html report

ekalin:
## Summary

This highlights the arguments of the steps in the HTML report, displaying them in bold.

## Details

The changes here add details about arguments to the .js file with the report data. A pull request against cucumber-html (https://github.com/cucumber/cucumber-html/pull/51) uses that information to display them in bold.

## How Has This Been Tested?

I've added a unit test. Unfortunately, I had to deviate a little from the style of the other tests. It seems the test harness used in other tests does not allow expressions with arguments and capturing them. So I tested the relevant function individually (which required making it package-protected). I've modelled the new test based on a similar one in `PrettyFormatterTest.java`.

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/runtime/formatter/HTMLFormatter.java
+++ b/core/src/main/java/cucumber/runtime/formatter/HTMLFormatter.java
@@ -381,7 +381,7 @@ final class HTMLFormatter implements EventListener {
         return false;
     }
 
-    private Map<String, Object> createTestStep(PickleStepTestStep testStep) {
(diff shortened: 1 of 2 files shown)


mpkorstanje:
Hey thanks for contributing. However we're looking to drop the cucumber report plugin in the future. You may also find you'll be able to generate better reports using:

* https://github.com/allure-examples/allure-cucumber-jvm-example
* https://github.com/damianszczepanik/cucumber-reporting

And I reckon there are even more third party tools out there.

PR #1454, opened 2018-09-09
Shellwords increase test count

nhojpatrick:
## Summary

Increase tests and switch to Parameterized test so easier to add new tests quicker

## Details

4 to 14 test, mainly whitespace checks added
depending upon regex order it either includes double quote or single quote so logic removes both
check `'"string"'` is extracted to `"string"` and `"'string'"` to `'string'`

## Motivation and Context

Continual improvement and test coverage

## How Has This Been Tested?

Build still passes

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/runtime/Shellwords.java
+++ b/core/src/main/java/cucumber/runtime/Shellwords.java
@@ -6,7 +6,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public class Shellwords {
-    private static final Pattern SHELLWORDS_PATTERN = Pattern.compile("[^\\s']+|'([^']*)'");
+    private static final Pattern SHELLWORDS_PATTERN = Pattern.compile("[^\\s'\"]+|[']([^']*)[']|[\"]([^\"]*)[\"]");
 
     private Shellwords() {
     }
@@ -18,7 +18,13 @@ public class Shellwords {
             if (shellwordsMatcher.group(1) != null) {
                 matchList.add(shellwordsMatcher.group(1));
             } else {
-                matchList.add(shellwordsMatcher.group());
+                String shellword = shellwordsMatcher.group();
+                if (shellword.startsWith("\"")
+                        && shellword.endsWith("\"")
+                        && shellword.length() > 2) {
+                    shellword = shellword.substring(1, shellword.length() - 1);
+                }
+                matchList.add(shellword);
             }
         }
         return matchList;
--- a/core/src/main/java/cucumber/runtime/Shellwords.java
+++ b/core/src/main/java/cucumber/runtime/Shellwords.java
@@ -7,30 +7,47 @@ import java.util.regex.Pattern;
 
 public class Shellwords {
 
-    private static final Pattern SHELLWORDS_PATTERN = Pattern.compile("[^\\s'\"]+|[']([^']*)[']|[\"]([^\"]*)[\"]");
+    private static final Pattern SHELLWORDS_PATTERN;
+
+    static {
+        final String DEFAULT = "([^\\s'\"]+)";
+        final String SINGLE_QUOTES = "[\\^\\s]?[']([^']*)['$]?";
+        final String DOUBLE_QUOTES = "[\\^\\s]?[\"]([^\"]*)[\"$]?";
+        final String SHELLWORDS_FORMAT = String.format("%s|%s|%s",
+            SINGLE_QUOTES,
+            DOUBLE_QUOTES,
+            DEFAULT
+        );
+        SHELLWORDS_PATTERN = Pattern.compile(SHELLWORDS_FORMAT);
+    }
 
     private Shellwords() {
     }
 
-    public static List<String> parse(String cmdline) {
-        List<String> matchList = new ArrayList<>();
-        Matcher shellwordsMatcher = SHELLWORDS_PATTERN.matcher(cmdline);
+    /**
+     * <p>Parse command line input and extract all shellwords into a {@code List<String>}.</p>
+     *
+     * <p>Shellwords are whitespace seperated, apart from when a shellword starts with a Double Quotes {@code "}, or Single Quotes {@code '}.
+     * Where a Quote starts a shellword, the shellword is only closed by another quote or the end of the line.</p>
+     *
+     * @param cmdline the command line arguments passed in
+     * @return a {@code List<String>} of the individual words supplied.
+     */
+    public static List<String> parse(final String cmdline) {
+        final List<String> matchList = new ArrayList<>();
+        final Matcher shellwordsMatcher = SHELLWORDS_PATTERN.matcher(cmdline);
         while (shellwordsMatcher.find()) {
-            if (shellwordsMatcher.group(1) != null) {
-                matchList.add(shellwordsMatcher.group(1));
-            } else {
-                String token = shellwordsMatcher.group();
-                boolean singleQuoted = token.startsWith("'")
-                    && token.endsWith("'");
-                boolean doubleQuoted = token.startsWith("\"")
-                    && token.endsWith("\"");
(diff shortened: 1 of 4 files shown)


nhojpatrick:
> So far you've been testing valid strings. What happens when invalid input is provided?

Added some weird usecases and tests are working but not sure if expected is correct. i.e.

1) Before Prefix'Suffix After - Should that be "Before", "Prefix", "Suffix", "After" or "Before", "Prefix'Suffix", "After" or "Before", "Prefix", "Suffix After"

2) Before Prefix"Suffix After - Should that be "Before", "Prefix", "Suffix", "After" or "Before", "Prefix"Suffix", "After" or "Before", "Prefix", "Suffix After"

3) "Before Prefix"Suffix After - Should that be "Before Prefix", "Suffix After"

4) "Before Prefix "Suffix After - Should that be "Before Prefix ", "Suffix After"

What should happen if a non whitespace character touches a quote?

core/src/main/java/cucumber/runtime/Shellwords.java
```diff
@@ -6,27 +6,43 @@
 import java.util.regex.Pattern;
 
 public class Shellwords {
-    private static final Pattern SHELLWORDS_PATTERN = Pattern.compile("[^\\s'\"]+|[']([^']*)[']|[\"]([^\"]*)[\"]");
+
+    private static final Pattern SHELLWORDS_PATTERN;
+
+    static {
+        final String SHELLWORDS_FORMAT = "\\s*(?>([^\\s\\\\\\'\\\"]+)|'([^\\']*)'|\"((?:[^\\\"\\\\]|\\\\.)*)\"|(\\\\.?)|(\\S))(\\s|\\z)?";
```
mpkorstanje: In one of your previous commits you split this up into several variables. This made the purpose of each capture group clear. That was a good idea.

mpkorstanje: You can also name capturing groups using `(?<name>X)` where X is the regex.

nhojpatrick: I'm not 100% sure which each of these capture groups does...
By debugging I've worked out the 2nd highest capture group is the capture group that should be used, so using names would make that harder instead of easier. I'm also assuming I can removed the last capture group but when I do the highest capture group isn't always the one you want...

core/src/main/java/cucumber/runtime/Shellwords.java
```diff
@@ -6,27 +6,43 @@
...
-                matchList.add(shellwordsMatcher.group(1));
-            } else {
-                String shellword = shellwordsMatcher.group();
-                if (shellword.startsWith("\"")
-                        && shellword.endsWith("\"")
-                        && shellword.length() > 2) {
-                    shellword = shellword.substring(1, shellword.length() - 1);
+            // default token to group
+            String token = shellwordsMatcher.group();
+            // find highest group count that is not null and use that
```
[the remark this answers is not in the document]
nhojpatrick: already written tests with leading and trailing whitespace
also already written tests with unbalanced quotes and nested double quotes within single quotes and nested single quotes within double quotes

core/src/test/java/cucumber/runtime/ShellwordsTest.java
```diff
@@ -1,30 +1,189 @@
...
+            // TODO is this correct with trailing empty string
+            new Object[]{
+                "--uniform \"Victor Whiskey --xray \"Yankee Zulu\"",
+                asList("--uniform", "Victor Whiskey --xray ", "Yankee", "Zulu", "\"")
+            },
+
+            // opening double_quote only with following single_quotes
+            new Object[]{
+                "--alfa \"Bravo Charlie --delta 'Echo Foxtrot'",
+                asList("--alfa", "\"", "Bravo", "Charlie", "--delta", "Echo Foxtrot")
```
mpkorstanje: I would expect this to be `asList("--alfa", "\"Bravo", "Charlie", "--delta", "Echo Foxtrot")`

core/src/test/java/cucumber/runtime/ShellwordsTest.java
```diff
@@ -1,30 +1,189 @@
...
+            },
+
+            // triple single_quotes
+            // TODO is this correct with trailing empty string
+            new Object[]{
+                "--golf 'Hotel India --juliett 'Kilo Lima'",
+                asList("--golf", "Hotel India --juliett ", "Kilo", "Lima", "'")
```
mpkorstanje: Trailing whitespace after julliet is correct.

core/src/test/java/cucumber/runtime/ShellwordsTest.java
```diff
@@ -1,30 +1,189 @@
...
+            },
+
+            // opening single_quotes only with following double_quotes
+            new Object[]{
+                "--mike 'November Oscar --papa \"Quebec Romeo\"",
+                asList("--mike", "'", "November", "Oscar", "--papa", "Quebec Romeo")
```
mpkorstanje: This has an unbalanced quote and should throw an error.

core/src/test/java/cucumber/runtime/ShellwordsTest.java
```diff
@@ -1,30 +1,189 @@
...
+            },
+
+            // double_quotes within word
+            new Object[]{
+                "sierra\"tango",
+                asList("sierra", "\"", "tango")
```
mpkorstanje: This should be a single string.

core/src/main/java/cucumber/runtime/Shellwords.java
mpkorstanje: The algorithm used in Ruby isn't trivial. I'll try to explain it line by line:

https://github.com/sj26/ruby-1.9.3-p0/blob/master/lib/shellwords.rb#L32-L34
```rb
words = []
field = ''
```
Our goal is to out put an array of words (strings separated by whitespace, unless quoted). To do this we are going to search through the string. We'll put the intermediate results that we find into field. When we encounter a separator we'll know that we've found a complete word.

https://github.com/sj26/ruby-1.9.3-p0/blob/master/lib/shellwords.rb#L35-L36
```rb
    line.scan(/\G\s*(?>([^\s\\\'\"]+)|'([^\']*)'|"((?:[^\"\\]|\\.)*)"|(\\.?)|(\S))(\s|\z)?/m) do
|word, sq, dq, esc, garbage, sep|
```
This searches the line for any bits that match the regular expression. The results is an array that contains the captured values of each capture group. The elements of this array are unpacked into the variables word, sq, dq, esc, garbage and sep.

https://github.com/sj26/ruby-1.9.3-p0/blob/master/lib/shellwords.rb#L37
```rb
 raise ArgumentError, "Unmatched double quote: #{line.inspect}" if garbage
```
If the garbage capture group capture anything it throws an exception.

https://github.com/sj26/ruby-1.9.3-p0/blob/master/lib/shellwords.rb#L38-L41
```rb
      field << (word || sq || (dq || esc).gsub(/\\(?=.)/, ''))
      if sep
        words << field
        field = ''
```
Then the fields word, single quote, double quote and escaped are appended to the current field we are building. The latter two have any escape character removed. Note that the separator character is not appended.

If there is a separator or line ending  we are done building the current field and put it in the list of fields.

mpkorstanje: The underlying principle is called [Lexical Analysis](https://en.wikipedia.org/wiki/Lexical_analysis). And normally you'd write a state machine that follows the lexical grammar. In this case the Ruby implementation is being clever by expressing that state machine as a regular expression.

That said. We shouldn't be in the business of being clever. It creates unreadable code.

Looking through the Java API I found the [StreamTokenizer](https://docs.oracle.com/javase/8/docs/api/java/io/StreamTokenizer.html). You may find it useful.

Issue #1450, opened 2018-09-02
Use ServiceLoader to find Backend and Object Factory implementations


mpkorstanje:
## Summary

We currently use class path scanning to find backend and factory implementations. This is slow and makes setting up mock for tests a pain (they are also discovered). A ServiceLoader avoids this problem. This wasn't possible earlier versions of Cucumber because we compiled these against java 6.

https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html

aslakhellesoy:
Yes, this would be a great improvement!

mpkorstanje:
Great! And yes, please use the v5 branch! You'll notice it has significant structural changes. 

This will help move us towards automatic module names, expose less of cucumbers implementation details, ect.

mpkorstanje:
Resolved by a03dd01.

PR #1463, opened 2018-09-13
Use ServiceLoader for ObjectFactory

nhojpatrick:
## Summary

Is this the type of changes you where expecting for #1450?

## Details

Build fails at spring tests, but before I send more time is this what you where thinking?

## How Has This Been Tested?

Not fully tested yet, tests are failing. Still work in process...

## Types of changes

## Checklist:


java/src/main/java/io/cucumber/java/ObjectFactoryLoader.java
```diff
@@ -2,45 +2,55 @@
...
-import io.cucumber.core.reflection.Reflections;
-import io.cucumber.core.reflection.TooManyInstancesException;
 
-import static java.util.Arrays.asList;
+import java.util.Iterator;
+import java.util.ServiceLoader;
 
 public class ObjectFactoryLoader {
+
+    private static final ServiceLoader<ObjectFactory> LOADER = ServiceLoader.load(ObjectFactory.class);
```
mpkorstanje: The service loader is not safe for use by multiple concurrent threads. Runners are executed in parallel and each runner will create it's own backends with object factory. You can either make all fields and methods non-static or move this field into the method.

java/src/main/java/io/cucumber/java/ObjectFactoryLoader.java
```diff
@@ -2,45 +2,55 @@
...
+
     private ObjectFactoryLoader() {
     }
 
     /**
      * Loads an instance of {@link ObjectFactory}. The class name can be explicit, or it can be null.
-     * When it's null, the implementation is searched for in the <pre>cucumber.runtime</pre> packahe.
+     * When it's null, the implementation is searched for in the <pre>cucumber.runtime</pre> package.
      *
+     * @deprecated as of version 4.0.0; use {@code loadObjectFactory() } instead.
```
mpkorstanje: While currently not package private, the ObjectFactoryLoader will be after merging. There is no need to keep a deprecated method around.

mpkorstanje:
> Build fails at spring tests, but before I send more time is this what you where thinking?

Looks exactly as expected! Thanks!

Btw. You'll notice that by adding whitespace changes in files that you otherwise didn't touch you'll now have merge conflicts. While I really appreciate the sentiment of cleaning things up, try to make these changes only in combination with other changes.

nhojpatrick:
yep i auto format style often without thing, sometimes i only do it so it appears in git as a file i need to go back to check later.

undone changes

mpkorstanje:
> Merge remote-tracking branch 'upstream/master' into develop-v5_1450_s…

You may want to revert that commit and instead use upstream/develop-v5

mpkorstanje:
I cherry picked your changes into a03dd01. Cheers!

PR #1475, opened 2018-10-04
[Java8] Apply identity transform when target type is unknown 

mpkorstanje:
## Summary

Apply identity transform when target type is unknown 

## Motivation and Context 

When using groovy in combination with `cucumber-java8` there is no type information available at all because:

```groovy
public class Stepdefs implements En {
	Stepdefs() {
		Given(/^this docstring:$/, { String s ->
			println s
		} as A1)
	}
}
```

is compiled into:

```java
public class Stepdefs implements En, GroovyObject {
    public Stepdefs() {
        CallSite[] var1 = $getCallSiteArray();
        super();
        MetaClass var2 = this.$getStaticMetaClass();
        this.metaClass = var2;
        var1[0].callCurrent(this, "^this docstring:$", ScriptBytecodeAdapter.createPojoWrapper((A1)ScriptBytecodeAdapter.asType(new Stepdefs._closure1(this, this), A1.class), A1.class));
    }

    public final class _closure1 extends Closure implements GeneratedClosure {
        public _closure1(Object _outerInstance, Object _thisObject) {
            CallSite[] var3 = $getCallSiteArray();
            super(_outerInstance, _thisObject);
        }

        public Object doCall(String s) {
            CallSite[] var2 = $getCallSiteArray();
            return var2[0].callCurrent(this, s);
        }

        public Object call(String s) {
            CallSite[] var2 = $getCallSiteArray();
            return !__$stMC && !BytecodeInterface8.disabledStandardMetaClass() ? this.doCall(s) : var2[1].callCurrent(this, s);
        }
    }
}

```
The `ScriptBytecodeAdapter.asType` forces a conversion of Closure to `A1`. Because `cucumber-java8` applies constant pool inspection to determine `T1` of the generic types of `A1<T1>` this yields an unknown result. There are no generic types to begin with.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,14 @@ Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CO
  * [Core] Reduce plugin memory usage ([#1469](https://github.com/cucumber/cucumber-jvm/pull/1469) M.P. Korstanje) 
 
 ### Changed
- * [Core] Use the docstring content type from pickle in the json formatter ([#1265](https://github.com/cucumber/cucumber-jvm/pull/1265) Robert Wittams, M.P. Korstanje) 
+ * [Core] Use the docstring content type from pickle in the json formatter ([#1265](https://github.com/cucumber/cucumber-jvm/pull/1265) Daryl Piffre, M.P. Korstanje) 
 
 ### Deprecated
    
 ### Removed
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


azryelryvel:
the issue that led to this change can be reproduced by running a "mvn test" on this repo : https://github.com/azryelryvel/cucumber-groovy-jvm

mpkorstanje:
Cheers! There will be a 4.0.1 release in a few hours from now.

PR #1476, opened 2018-10-06
Allow different sources for configuration

martinschneider:
Added CucumberOptionsProvider to dynamically create and use CucumberOptions

Usage example:
```
@RunWith(Cucumber.class)
@CucumberOptionsProvider(PropertiesFileCucumberOptionsProvider.class)
public class JUnitRunnerClass {}
```

An implementation of `AbstractCucumberOptionsProvider` can be extended to load or create a configuration from any source.
Options passed via `@CucumberOptions` will be loaded after the ones from the options-provider.

## Summary

A limitation of using `@CucumberOptions` is that its values are constant (i.e. determined at compile time). This PR introduces an option to contribute these values dynamically during runtime.

## Details

An additional annotation `CucumberOptionsProvider` specifies a class which will be called in `RuntimeOptionsFactory` to fetch the `CucumberOptions` to be used.

An example using the JUnit Runner could look like this:
```
@RunWith(Cucumber.class)
@CucumberOptionsProvider(PropertiesFileCucumberOptionsProvider.class)
public class JUnitRunnerClass {}
```

## Motivation and Context

Every test framework I have written myself or co-worked on which uses Cucumber + JUnit at some point faces the issue described above. I think this PR is a simple enough and useful way to remove this limitation.

It has also been discussed on forums multiple times without a fully satisfying solution. The only option giving the same flexibility as implemented here would be to use a custom JUnit runner instead of `cucumber.api.junit.Cucumber`.

* https://stackoverflow.com/questions/22774099/pass-cucumber-options-dynamically-in-cucumber-junit
* https://stackoverflow.com/questions/45095557/java-cucumber-take-cucumberoptions-from-external-source-like-a-property-file/52663213#52663213

## How Has This Been Tested?

I have added unit tests in `cucumber.runtime.RuntimeOptionsFactoryTest`.
I have created an additional test class `cucumber.runtime.PropertiesFileCucumberOptionsProviderTest`
I have tested the change within my own test framework: https://github.com/martinschneider/yasew/pull/17

## Types of changes

## Checklist:


syamsasi99:
+1

abhivaikar:
Yes. Much needed. +1

pom.xml
```diff
@@ -83,6 +83,7 @@
         <typetools.version>0.5.0</typetools.version>
         <cucumber-expressions.version>6.1.0</cucumber-expressions.version>
         <datatable.version>1.1.3</datatable.version>
+        <geantyref.version>1.3.4</geantyref.version>
```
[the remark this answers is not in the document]
mpkorstanje: Though given my main objection I doubt GeAnTyRef will be necessary.

core/pom.xml
```diff
@@ -8,6 +8,7 @@
     </parent>
 
     <artifactId>cucumber-core</artifactId>
+    <version>4.0.2-SNAPSHOT</version>
```
mpkorstanje: Please rebase on master instead. This will cause merge conflicts down the line. You can ignore the failing semantic versioning check at the moment.

martinschneider:
Thanks for your comments @mpkorstanje. I will look into it.

martinschneider:
Thanks again for your feedback @mpkorstanje. I had a dive into this and updated the PR.

This is far from finished. However, could you have a quick look and tell me if this is the general direction you were looking for.

There is a new annotation called `CucumberConfiguration` which takes two values `type` and `path`.

`type` can be `YAML`, `JSON`, `PROPERTIES` etc. and `ANNOTATION` which is the fallback to the current behaviour with `@CucumberOptions`. I've played around with `YAML` but it would be easy to implement the others (and I'd be happy to do so).

`path` is the path to the config file (file or classpath).

The YAML syntax is rather simple:
```tags:
  - `@tag1`
  - `@tag2`
glue:
  - io.github.glue1
  - io.github.glue2
plugins:
dry: false
monochrome: false
strict: true
snippets:
junit:
features: src/test/resources/features
```

When a parameter expects a list, single values are also accepted, so e.g. this will work too:
```glue: io.github.glue1```

The flow is as follows:

1. `RuntimeOptionsFactory` evaluates the annotation and delegates parsing of the config file to an implementation of `ConfigurationParser` (so far I've done `OptionsConfigurationParser` and `YamlConfigurationParser`)
2. `ConfigurationParser` returns `Map<String, ?>`
3. `RuntimeOptionsFactory` transforms this map into `RuntimeOptions`

It works ok so far but tbh I haven't tested much more than the standard cases. It's still a long way and even more testing to go.

Just asking for some guidance before I continue with this.

File: core/src/main/java/cucumber/runtime/RuntimeOptionsFactory.java [hunk 3]
```diff
@@ -1,170 +1,244 @@
...
+  public RuntimeOptionsFactory(Class<?> clazz) {
+    this.clazz = clazz;
+  }
+
+  public RuntimeOptions create() {
+    CucumberConfiguration configuration = clazz.getAnnotation(CucumberConfiguration.class);
+
+    // TODO: evaluate annotations on super classes
+    if (configuration == null || configuration.type().equals(Type.ANNOTATION)) {
+      return new RuntimeOptions(buildArgsFromOptions());
```
mpkorstanje: I am not sure if it is wise to let the `RuntimeOptionsFactory` handle both the `@CucumberConfiguration` and `@CucumberOptions` annotations. They are different sources of configuration that can exist together and should be applied in some order.

core/src/main/java/cucumber/api/CucumberConfiguration.java
```diff
@@ -0,0 +1,18 @@
...
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE})
+public @interface CucumberConfiguration {
+  Type type() default Type.YAML;
+
+  String path();
+
+  public enum Type {
+    YAML, JSON, PROPERTIES, ANNOTATION;
```
mpkorstanje: As @CucumberConfiguration` and `@CucumberOptions` are different sources of configuration and  as yaml, json and properties can be derived from the file name I don't think this type is needed.

pom.xml
```diff
@@ -480,6 +482,16 @@
...
             </dependency>
+            <dependency>
+                <groupId>com.google.code.gson</groupId>
+                <artifactId>gson</artifactId>
+                <version>${gson.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.yaml</groupId>
+                <artifactId>snakeyaml</artifactId>
+                <version>${snakeyaml.version}</version>
```
mpkorstanje: One of the downsides of building a testing framework is that you can't use all the popular libraries. There are ways to work around this. We'll come to that later.

mpkorstanje:
So I would imagine something like this to create the options in JUnit/TestNG

```java
RuntimeOptions runtimeOptions = RuntimeOptions.builder()
   .addRuntimeOptionsProvider(new CucumberConfigurationOptionsProvider(class))
   .addRuntimeOptionsProvider(new CucumberOptionsFactory(env))
   .addRuntimeOptionsProvider(new EnvironmentOptionsProvider(env))
   .build();
```

And something like this for the CLI:

```java
RuntimeOptions runtimeOptions = RuntimeOptions.builder()
   .addRuntimeOptionsProvider(new CommandlineOptionsProvider(class))
   .addRuntimeOptionsProvider(new CucumberOptionsFactory(env))
   .addRuntimeOptionsProvider(new EnvironmentOptionsProvider(env))
   .build();
```

martinschneider:
@aslakhellesoy Sure, I will create a PR using cucumber-config

core/src/main/java/cucumber/api/CucumberConfiguration.java
martinschneider: Makes sense. I will change it accordingly.

pom.xml
martinschneider: Understood. However, if I base the PR on `cucumber-config` instead as suggested these libraries will come as transitive dependencies. What's the reason for not having them (footprint, Java version limitations)? What's the alternative you suggest?

mpkorstanje: Cucumber is a test library. Adding dependencies means that those dependencies are also present on the class path when tests are executed. Now depending on the build system used, the order in the dependency declaration it may also occur that our transitive dependencies replace the transitive dependencies of the project we're testing. This is highly undesirable.

We can resolve it by adding the dependency with scope  `provided` and checking if the class we're interested in is available on the class path before creating an instance of it.

aslakhellesoy: We’ll shade those deps, created a ticket in the monorepo

mpkorstanje:
I'm closing this issue for now. While useful the way this was originally implemented is not sustainable. I've made several attempts to refactor `RuntimeOptions` to make this possible but the whole thing is to tangled up to do this with any degree of certainty.

The only way out that I see is to implement new runners (JUnit5, Maven, Gradle) that do not use `RuntimeOptions` to parse their options. Then in due time we can deprecate JUnit and TestNG. To this end `RuntimeOptions` now  implements `FeatureOptions`, `FilterOptions`, `PluginOptions` and `RunnerOptions`.

mlvandijk:
(deleted duplicate comments)

Issue #1485, opened 2018-10-18
IllegalStateException thrown when parallel running with Guice + @ScenarioScoped annotation


james-bjss:
## Summary

When using the new parallel feature of cucumber 4 with cucumber-guice and the JUnit runner an ```IllegalStateException``` is thrown. The error was observed when Step classes were annotated with ```@ScenarioScoped```

## Expected Behavior

Scopes should be separate across threads.

## Current Behavior

The following Exception is thrown:
```Feature: Applicant does not provide a day in Date Of Birth fieldjava.lang.IllegalStateException: A scoping block is already in progress
        at cucumber.runtime.java.guice.impl.SequentialScenarioScope.checkState(SequentialScenarioScope.java:64)
        at cucumber.runtime.java.guice.impl.SequentialScenarioScope.enterScope(SequentialScenarioScope.java:52)
        at cucumber.runtime.java.guice.impl.GuiceFactory.start(GuiceFactory.java:34)
        at cucumber.runtime.java.JavaBackend.buildWorld(JavaBackend.java:114)
        at cucumber.runner.Runner.buildBackendWorlds(Runner.java:120)
        at cucumber.runner.Runner.runPickle(Runner.java:38)
        at cucumber.runtime.junit.PickleRunners$NoStepDescriptions.run(PickleRunners.java:146)
        at cucumber.runtime.junit.FeatureRunner.runChild(FeatureRunner.java:68)
        at cucumber.runtime.junit.FeatureRunner.runChild(FeatureRunner.java:23)
        at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
        at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
        at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
        at cucumber.runtime.junit.FeatureRunner.run(FeatureRunner.java:73)
        at cucumber.api.junit.Cucumber.runChild(Cucumber.java:124)
        at cucumber.api.junit.Cucumber.runChild(Cucumber.java:65)
        at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
[...]

PR #1486, opened 2018-10-18
[Guice] Replace CucumberModules.SCENARIO with threadsafe factory method

james-bjss:
## Summary

Updating SequentialScenarioScope to use a threadlocal. This is to prevent issues when running cucumber in parallel mode.

Fixes: https://github.com/cucumber/cucumber-jvm/issues/1485

## Details
Have updated SequentialScenarioScope.class to use ThreadLocal to prevent it from thinking a scoping block is already in progress when running test from surefire with the parallel settings enabled.

## Motivation and Context
This was preventing us from running our test suites with the new parallel running functionality of Cucumber4

## How Has This Been Tested?
Have tested this by using the new scope within our project.
I have also added tests which can be used to reproduce the exception and also prove the change.

Not sure if this is the best solution or has other implications, hoping somebody here can review and let me know if there is a better way.

## Types of changes

## Checklist:

### DIFF
--- a/guice/src/main/java/cucumber/api/guice/CucumberScopes.java
+++ b/guice/src/main/java/cucumber/api/guice/CucumberScopes.java
@@ -8,5 +8,13 @@ import cucumber.runtime.java.guice.impl.SequentialScenarioScope;
  * in implementations of <code>com.google.inject.Module</code>.
  */
 public class CucumberScopes {
+    /**
+     * @deprecated please use {@link #createScenario()} instead
+     */
     public static final ScenarioScope SCENARIO = new SequentialScenarioScope();
+
+    public static ScenarioScope createScenario() {
+        return new SequentialScenarioScope();
+    }
+
 }
--- a/guice/src/main/java/cucumber/api/guice/CucumberModules.java
+++ b/guice/src/main/java/cucumber/api/guice/CucumberModules.java
@@ -1,6 +1,7 @@
 package cucumber.api.guice;
 
 import com.google.inject.Module;
+import cucumber.runtime.java.guice.ScenarioScope;
 import cucumber.runtime.java.guice.impl.ScenarioModule;
 
 /**
@@ -9,5 +10,17 @@ import cucumber.runtime.java.guice.impl.ScenarioModule;
  * <code>cucumber.runtime.java.guice.ScenarioScope</code>.
  */
 public class CucumberModules {
+    /**
+     * @deprecated please use {@link #createScenarioModule()} instead
+     */
+    @Deprecated
     public static final Module SCENARIO = new ScenarioModule(CucumberScopes.SCENARIO);
+
+    public static Module createScenarioModule() {
+        return new ScenarioModule(CucumberScopes.createScenario());
+    }
+
+    public static Module createScenarioModule(ScenarioScope scenarioScope) {
+        return new ScenarioModule(scenarioScope);
+    }
 }
--- a/guice/src/main/java/cucumber/api/guice/package.html
+++ b/guice/src/main/java/cucumber/api/guice/package.html
@@ -121,7 +121,7 @@
     Guice injector and it's Guice modules. The injector must provide a binding for
     <code>cucumber.runtime.java.guice.ScenarioScope</code>. It should also provide a binding for the
     <code>cucumber.runtime.java.guice.ScenarioScoped</code> annotation if your classes are using the annotation. The
-    easiest way to do this it to use <code>CucumberModules.SCENARIO</code>. For example:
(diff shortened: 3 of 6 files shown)


james-bjss:
Hi, thanks for the quick reply. Sure, happy to have a go. There is a chance I might not be able to work on this until the weekend though but I will try my best :)

mpkorstanje:
Cheers! No rush.

james-bjss:
It seems that updating the existing tests to use ``CucumberModules.createScenario()``` causes failures in some of the tests.

```[ERROR] shouldGiveNewInstanceOfAnnotatedSingletonClassForEachScenarioWhenOverridingModuleBindingIsScenarioScope(cucumber.runtime.java.guice.impl.GuiceFactoryTest)  Time elapsed: 0.019 s  <<< ERROR!
com.google.inject.ProvisionException: 
Unable to provision, see the following errors:

1) Error in custom provider, com.google.inject.OutOfScopeException: Cannot access Key[type=cucumber.runtime.java.guice.impl.GuiceFactoryTest$AnnotatedSingletonClass, annotation=[none]] outside of a scoping block
  at cucumber.runtime.java.guice.impl.GuiceFactoryTest$3.configure(GuiceFactoryTest.java:161)
  while locating cucumber.runtime.java.guice.impl.GuiceFactoryTest$AnnotatedSingletonClass```

Likewise when with changing the custom injector in the integration tests. Will take a fresh look tomorrow.

**Comment by mpkorstanje** [comment]
I'm guessing that you used a refactoring tool to inline the the variable. Then I'm also guessing a test like `shouldGiveNewInstanceOfAnnotatedSingletonClassForEachScenarioWhenOverridingModuleBindingIsScenarioScope` will fail because here SCENARIO should still refer to the same instance. So you'd have to create the scenario module only once to make it work.

```java
    @Test
    public void shouldGiveNewInstanceOfAnnotatedSingletonClassForEachScenarioWhenOverridingModuleBindingIsScenarioScope() {
        factory = new GuiceFactory(injector(CucumberModules.SCENARIO, new AbstractModule() {
            @Override
            protected void configure() {
                bind(AnnotatedSingletonClass.class).in(CucumberScopes.SCENARIO);
            }
        }));
        instancesFromDifferentScenarios = getInstancesFromDifferentScenarios(factory, AnnotatedSingletonClass.class);
        assertThat(instancesFromDifferentScenarios, elementsAreAllUnique());
    }
```

mpkorstanje:
I think it is possible to add `CucumberModules.createModule(ScenarioScope scenarioScope)` to make that use case a bit easier though. 

But as you're actually using Guice. What do you think?

james-bjss:
> At the same time I don't quite understand the purpose of `YourModule`. To my best understanding this use case should be covered by `@ScenarioScoped`. So if this is not a common use case then I'm okay with putting the burden of that complexity on the people who use it.

In our case we use it provide named instances of Wiremock clients which are configured to point to different remote instances. Also our applicaiton development teams use guice and sometimes there are components we reuse in our test code. Have also seen people using guice to instantiate WebDriver and this is configured within their own module. 

I agree though, there is probably not a large portion of the user base using this. I am just happy to have working solution and help others avoid the same issue. At the end of the day it's not something that changes much in the project so complexity is less of an issue, so long as it's documented.

mpkorstanje:
>  As per - 184858c would that be acceptable?

I think it is a good solution! Wasn't aware of this possibility.

james-bjss:
Thanks. Am I Ok to merge once the build is green or should I tidy/squash my commits first?

guice/src/main/java/cucumber/api/guice/CucumberScopes.java
```diff
@@ -9,4 +9,9 @@
  */
 public class CucumberScopes {
     public static final ScenarioScope SCENARIO = new SequentialScenarioScope();
```
mpkorstanje: This one should also be deprecated

guice/src/main/java/cucumber/api/guice/CucumberScopes.java
```diff
@@ -9,4 +9,9 @@
...
+
+    public static ScenarioScope createScenario() {
```
mpkorstanje: This should be consistent with `createScenarioModule`

aslakhellesoy:
Hi @james-bjss,

Aslak Hellesøy
Creator of Cucumber

mpkorstanje:
> I took a different approach in 184858c
> by binding the class instead of an instance of the ScenarioScope.

> Would you preffer I backed that out?

No. That was a good solution. 

Anyway. There should be a 4.1.0 release in a few hours once the build finishes and artefacts are propagated to maven central.

Issue #1493, opened 2018-10-28
Use cucumber-parent pom


mlvandijk:
## Summary

Aslak made [cucumber/cucumber-parent](https://github.com/cucumber/cucumber-parent) which extracts quite a bit of shared settings for all the java modules.

If you like you can start with setting it up in `cucumber-jvm`

## Expected Behavior

Cucumber-jvm builds with cucumber-parent pom
Remove a lot of duplication from cucumber-jvm pom

## Current Behavior

Lots of duplication

## Possible Solution

Cucumber-jvm builds with cucumber-parent pom

## Steps to Reproduce (for bugs)
N/A

## Context & Motivation
Remove duplication

## Your Environment
N/A

mpkorstanje:
Resolved by #1494

PR #1494, opened 2018-10-28
Use cucumber parent

mlvandijk:
## Summary

Use cucumber-parent pom and remove duplication.

## Details

Open issues:
- [x] release updated cucumber-parent and use new version (1.0.2)
- [x] ~~please ensure that all projects or any sub-modules that use jar packaging have project.Automatic-Module-Name set.~~ Update: Cucumber JVM has split packages. So we cant use automatic module names yet. So that needs a work around. Can you disable it for cucumber JVM and keep it for the others? (DONE; added override)
- [x] check config for nexus-staging-maven-plugin (DONE; seems fine)

## Motivation and Context

See  #1493

## Types of changes

## Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -657,16 +657,13 @@
                     </configuration>
                 </plugin>
 
-                <!-- NOTE: parent adds additional configuration:
-                 Please ensure that all projects or any sub-modules that use jar packaging have project.Automatic-Module-Name set.-->
-                    <!--<configuration>-->
-                        <!--<archive>-->
-                            <!--<manifestEntries>-->
-                                <!--<Automatic-Module-Name>${project.Automatic-Module-Name}</Automatic-Module-Name>-->
-                            <!--</manifestEntries>-->
-                        <!--</archive>-->
-                    <!--</configuration>-->
-                <!--</plugin>-->
+
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-jar-plugin</artifactId>
+                    <version>3.0.2</version>
+                    <configuration combine.self="override"/>
+                </plugin>
 
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -788,17 +788,6 @@
                     </execution>
                 </executions>
             </plugin>
-
-            <plugin>
-                <groupId>org.sonatype.plugins</groupId>
-                <artifactId>nexus-staging-maven-plugin</artifactId>
-                <extensions>true</extensions>
-                <configuration>
-                    <serverId>ossrh</serverId>
-                    <nexusUrl>https://oss.sonatype.org/</nexusUrl>
-                    <autoReleaseAfterClose>true</autoReleaseAfterClose>
-                </configuration>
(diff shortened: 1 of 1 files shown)


mlvandijk:
Differences in `nexus-staging-maven-plugin`:

**Before using cucumber-parent pom:**
```
      <pluginManagement>

        <plugins>
          <plugin>
            <groupId>org.sonatype.plugins</groupId>
            <artifactId>nexus-staging-maven-plugin</artifactId>
            <version>1.6.7</version>
          </plugin>
        </plugins>
      </pluginManagement>
```

**After adding cucumber-parent pom:**
```
      <pluginManagement>
        <plugins>
            <groupId>org.sonatype.plugins</groupId>
            <artifactId>nexus-staging-maven-plugin</artifactId>
            <version>1.6.7</version>
            <extensions>true</extensions>
            <configuration>
              <serverId>ossrh</serverId>
              <nexusUrl>https://oss.sonatype.org/</nexusUrl>
              <autoReleaseAfterClose>true</autoReleaseAfterClose>
            </configuration>
          </plugin>
        </plugins>
      </pluginManagement>
```

mpkorstanje:
Does it come from the parent? Do we have any other source for the same settings in this pom or are these defaults?

mlvandijk:
Yes, it comes from the parent, see [line 344](https://github.com/cucumber/cucumber-parent/blob/master/pom.xml#L344)

mpkorstanje:
Are they defaults? If so, they can be removed from the pom

mpkorstanje:
From parent pom

mlvandijk:
From the [Maven documentation](http://maven.apache.org/pom.html#Plugin_Management): "pluginManagement: is an element that is seen along side plugins. Plugin Management contains plugin elements in much the same way, except that rather than configuring plugin information for this particular project build, it is intended to configure project builds that inherit from this one. However, this only configures plugins that are actually referenced within the plugins element in the children or in the current POM. The children have every right to override pluginManagement definitions."

Since we use the same configuration in the plugin as in the pluginmanagement, this seems fine to me. If anything we could remove the additional <plugins><plugin> configuration in the cucumber-jvm pom, as it's already inherited from the cucumber-parent. (Have tried locally; this has no effect on the effective pom)

mlvandijk:
Added override for Automatic-Module-Name as suggested [here](https://stackoverflow.com/questions/26736876/difference-between-plugins-and-pluginmanagement-tag-in-maven-pom-xml/26739966).

pom.xml
```diff
@@ -885,16 +788,7 @@
...
-                    <autoVersionSubmodules>true</autoVersionSubmodules>
-                    <useReleaseProfile>false</useReleaseProfile>
-                    <releaseProfiles>release</releaseProfiles>
-                    <goals>deploy</goals>
-                </configuration>
-            </plugin>
+
             <plugin>
                 <groupId>org.sonatype.plugins</groupId>
                 <artifactId>nexus-staging-maven-plugin</artifactId>
```
mpkorstanje: If this plugin with this configuration is present in the parent, we don't need it here.

mlvandijk: I'll double check at home before (squash and) merging.

Issue #1507, opened 2018-11-26
When running with @CucumberOptions, consider using `src/test/resources/features/` as default for `features`


christianhujer:
## Summary
Make the directory `src/test/resources/features/` the fallback default location for cucumber-jvm to search for feature files in case the current default location has no feature files.

## Expected Behavior
Using the Maven directory structure is probably the most popular directory structure in JVM-based projects. In this directory structure, feature files would go in `src/test/resources/features/`. It should therefore be the default when no options have been specified and no feature files have been put in the package that contains the class which carries the `@CucumberOptions` annotation.

## Current Behavior
The feature files are searched for in the package of the class that carries the `@CucumberOptions` annotation.

## Possible Solution
In `RuntimeOptionsFactory`, change the implementation of `addDefaultFeaturePathIfNoFeaturePathIsSpecified()`.

Instead of
```
    private void addDefaultFeaturePathIfNoFeaturePathIsSpecified(List<String> args, Class clazz) {
        if (!featuresSpecified) {
            args.add(MultiLoader.CLASSPATH_SCHEME + packagePath(clazz));
        }
    }
```
do this:
```
    private void addDefaultFeaturePathIfNoFeaturePathIsSpecified(List<String> args, Class clazz) {
        if (!featuresSpecified) {
            String path = MultiLoader.CLASSPATH_SCHEME + packagePath(clazz);
            if (containsFeatures(path))
                args.add(path);
            else
                args.add("src/test/resources/features/");
        }
    }
```
Warning: I haven't tested it, and I'm sure it's flawed. Also, I haven't provided function `boolean containsFeatures(String path)`.

## Context & Motivation
The maven directory structure is probably the most popular directory structure for Java projects these days. In this directory structure, feature files would be placed in `src/test/resources/features/`. It makes sense to consider that as a default so that users do not have to specify this common path explicitly. In order to not break users which rely on the existing default, using `src/test/resources/features` should only be the second default, if the first (current) default doesn't contain any features.

mpkorstanje:
Thanks for suggesting this. In principle Cucumber should work with a minimal amount of configuration and use sensible defaults. 

However in principle Cucumber should also be framework agnostic. Framework specific integration can be provided by extensions like `cucumber-junit` or `cucumber-testng`. As such project structure specific defaults would be more at home in a Maven or Gradle plugin. 

[...]

PR #1508, opened 2018-11-26
correct method RuntimeOptionsFactory.addDefaultFeaturePathIfNoFeature…

svavka:
…PathIsSpecified - if not defined feature path set package path or absolute src/test/resources/features/ if feature files not found

correct AbstractTestNGCucumberTestsTest.setUp method - replace deprecated use method testNG.addListener
correct test WeldFactoryTest.stopCalledWithoutStart - replaced line wrapping with the default in the OS

## Summary
issue #1507

Make the directory src/test/resources/features/ the fallback default location for cucumber-jvm to search for feature files in case the current default location has no feature files.
## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/runtime/RuntimeOptionsFactory.java
+++ b/core/src/main/java/cucumber/runtime/RuntimeOptionsFactory.java
@@ -10,6 +10,7 @@ import java.util.List;
 import static java.util.Arrays.asList;
 
 public class RuntimeOptionsFactory {
+    private static final String FEATURE_SUFFIX = "feature";
     private final Class clazz;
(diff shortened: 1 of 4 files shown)


mpkorstanje:
@svavka what are your intentions for this? Just to be clear about it, this is not a desirable feature.

svavka:
@mpkorstanje, I implemented the request #1507, sorry that it was not in demand

mpkorstanje:
The intent is much appreciated! Please see my comment in #1507.

svavka:
@mpkorstanje, Thank you very much for the explanation, I'll know. I will try to help you in a really necessary problem.

Issue #1516, opened 2018-12-12
TypeRegistryConfigurer with @Autowired


bojanv55:
Is it possible to use `@Autowired` spring annotation inside `TypeRegistryConfigurer`? 

```
@Component
public class TypeRegistryConfiguration implements TypeRegistryConfigurer {

	private SpecifierMapper specifierMapper;

	public TypeRegistryConfiguration(SpecifierMapper specifierMapper){
		this.specifierMapper = specifierMapper;
	}

	@Override
...
```

Currently I get an error:

```
cucumber.runtime.CucumberException: java.lang.NoSuchMethodException: com.a.b.TypeRegistryConfiguration.<init>()

	at cucumber.runtime.Reflections.newInstance(Reflections.java:53)
	at cucumber.runtime.Reflections.instantiateSubclasses(Reflections.java:35)
	at cucumber.runtime.Reflections.instantiateExactlyOneSubclass(Reflections.java:17)
	at cucumber.runtime.BackendModuleBackendSupplier.loadBackends(BackendModuleBackendSupplier.java:46)
	at cucumber.runtime.BackendModuleBackendSupplier.get(BackendModuleBackendSupplier.java:37)
	at cucumber.runner.ThreadLocalRunnerSupplier.createRunner(ThreadLocalRunnerSupplier.java:44)
	at cucumber.runner.ThreadLocalRunnerSupplier.access$000(ThreadLocalRunnerSupplier.java:15)
	at cucumber.runner.ThreadLocalRunnerSupplier$1.initialValue(ThreadLocalRunnerSupplier.java:24)
	at cucumber.runner.ThreadLocalRunnerSupplier$1.initialValue(ThreadLocalRunnerSupplier.java:21)
	at java.lang.ThreadLocal.setInitialValue(ThreadLocal.java:180)
	at java.lang.ThreadLocal.get(ThreadLocal.java:170)
	at cucumber.runner.ThreadLocalRunnerSupplier.get(ThreadLocalRunnerSupplier.java:40)
	at cucumber.api.junit.Cucumber.<init>(Cucumber.java:108)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
[...]

PR #1518, opened 2018-12-20
configure registry with configurer from objectFactory

bojanv55:
## Types of changes

## Checklist:

### DIFF
--- a/java/src/main/java/io/cucumber/java/JavaBackend.java
+++ b/java/src/main/java/io/cucumber/java/JavaBackend.java
@@ -1,6 +1,17 @@
 package io.cucumber.java;
 
+import static io.cucumber.core.io.MultiLoader.packageName;
+import static java.lang.Thread.currentThread;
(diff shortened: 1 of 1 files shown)


bojanv55:
Related to [this](https://github.com/cucumber/cucumber-jvm/issues/1516). Does look a bit ugly but seems working. Otherwise needs much more refactoring because of circular dependencies...

mpkorstanje:
Nice! I'm afraid I can't accept this without the refactoring though. You can break the cyclic dependency by moving the object factory and its related classes to the core module. Have a look at the way I did this for v5:

https://github.com/cucumber/cucumber-jvm/blob/develop-v5/core/src/main/java/io/cucumber/core/runtime/BackendServiceLoader.java

But if it is too much effort you can also of course wait for v5. :)

mpkorstanje:
I think it's better to revisit this in v5.

PR #1528, opened 2019-01-04
[Core] Cache all matched step defintions 

lsuski:
## Summary

Performance improvements in step definitions searching

## Details

- Allow caching of step definitions with arguments. It was previously disabled because of DocString and Table arguments. I've change it in such way to always create new arguments list from PickleStep but do not search for definition if step text is the same
- create some objects once in `JavaStepDefinition`

## Motivation and Context

Improve performance of most often executed operations - `Glue.stepDefinitionMatches` is executed `step_definitions_count * average_steps_count_per_scenario * scenarios_count`

## How Has This Been Tested?
Tested on my Android project with 800 scenarios

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/runtime/MethodFormat.java
+++ b/core/src/main/java/cucumber/runtime/MethodFormat.java
@@ -12,6 +12,7 @@ import java.util.regex.Pattern;
 public class MethodFormat {
     private static final Pattern METHOD_PATTERN = Pattern.compile("((?:static\\s|public\\s)+)([^\\s]*)\\s\\.?(.*)\\.([^\\(]*)\\(([^\\)]*)\\)(?: throws )?(.*)");
     private static final String PACKAGE_PATTERN = "[^,]*\\.";
+    private static final Pattern COMPILED_PACKAGE_PATTERN = Pattern.compile(PACKAGE_PATTERN);
     private final MessageFormat format;
 
     public static final MethodFormat SHORT = new MethodFormat("%c.%m(%a)");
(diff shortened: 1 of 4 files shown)


lsuski:
Ok, I'll push unit test for that after weekend

mpkorstanje:
Owh btw. Out of curiosity what was the performance before and after?

lsuski:
So, I have 1065 step definitions. On Android 6.0 which is quite slow getting single step from cache takes up to 50 ms and searching through all definitions takes up to 500ms. Before that fix I have about 3-4 seconds gap between scenarios. Now for many scenarios this gap is 0.5s-1s. More steps are cached the better overall speed of all scenarios execution. Improvement is especially visible in  Features with background steps and scenario outlines

core/src/main/java/cucumber/runtime/MethodFormat.java
```diff
@@ -12,6 +12,7 @@
 public class MethodFormat {
     private static final Pattern METHOD_PATTERN = Pattern.compile("((?:static\\s|public\\s)+)([^\\s]*)\\s\\.?(.*)\\.([^\\(]*)\\(([^\\)]*)\\)(?: throws )?(.*)");
     private static final String PACKAGE_PATTERN = "[^,]*\\.";
```
mpkorstanje: It's okay to inline this.

lsuski: Done

lsuski:
That's all for now

mpkorstanje:
There should be new a release in a bit.

PR #1530, opened 2019-01-04
[JUnit] Implement Cucumber as a Junit Platform Engine

mpkorstanje:
## Summary

Adds Cucumber implementation of JUnit Platforms `TestEngine` that will discover, filter and execute scenarios.

Fixes: #1149  

## Motivation and Context

The motivation for this PR is two fold. To reduce the amount of configuration needed to run Cucumber and to facilitate developers of tooling that integrates Cucumber into test runners and IDE's.

#### Facilitating tooling ###

Unlike previous versions of JUnit, JUnit 5 is composed of several different modules from three different sub-projects.
```
JUnit 5 = JUnit Platform + JUnit Jupiter + JUnit Vintage
```
The JUnit Platform serves as a foundation for launching testing frameworks on the JVM. It also defines the TestEngine API for developing a testing framework that runs on the platform. By implementing Cucumber as a `TestEngine` we can benefit from existing implementations that use the JUnit Platform. This will make it easier to develop tooling for Cucumber - any tooling that supports the Junit Platform will automatically support the execution of Cucumber Scenarios. For example: 

image

Note: Support is limited to running cucumber tests. Syntax highlighting, and other editor specific features are not supported.

#### Reduce boiler plate ####

In the ideal situation a project only contains step definitions and feature files. Currently however some additional setup is needed in the form of a runner class annotated with `@CucumberOptions`. By using the JUnit platform the presence of Cucumber on the class path will make it so that tooling can automatically execute features.

Coincidentally the JUnit Platform only allows a single source of configuration. This makes it clear in which configuration cucumber will be executed. This is an improvement over the current hierarchy of environment and annotation based variables.

## Work in progress
 - [x] Map `ConfigurationParameters` to `RuntimeOptions` in  `CucumberEngineExecutionContext`.  This will require some refactoring of `RuntimeOptions`. Note that not all options are valid in this context.  Valid options are: `--dry-run`, `--monochrome`, `--strict`, `--plugin` and `--glue`.
- [x] Document the available configuration parameters and how to use them in maven. Comparable to the [documentation for Junit 5 itself](https://maven.apache.org/surefire/maven-surefire-plugin/examples/junit-platform.html). 
 - [ ] Resolve features suggested by `ModuleSelector` in `DiscoverySelectorResolver`
 - [x] Resolve features suggested by `ClasspathRootSelector` in `DiscoverySelectorResolver`
 - [x] Resolve features suggested by `UniqueIdSelector` in `DiscoverySelectorResolver`
 - [x] Resolve features suggested by `UriSelector` in `DiscoverySelectorResolver`.
 - [x] Implement `getTags` for `PickleDescriptor`, `ScenarioOutlineDescriptor` and `FeatureFileDescriptor`.
 - [x] Ensure IDEA picks up the CucumberTestEngine implementation. See: https://github.com/cucumber/cucumber-jvm/issues/1149#issuecomment-443423653
 - [ ] Submit a patch Maven Surefire the JunitPlatformProvider to support non-class files. See: https://github.com/cucumber/cucumber-jvm/issues/1149#issuecomment-443423653

## Types of changes

## Checklist:


spring/src/test/java/cucumber/runtime/java/spring/threading/RunParallelCucumberTest.java
```diff
@@ -12,7 +12,7 @@
 import java.util.concurrent.Future;
 
 
-public class RunParallelCukesTest {
+public class RunParallelCucumberTest {
```
mlvandijk: +1 :)

aaronols:
I noticed a lot of `cucucumber` package names in the diff. Is that intentional or a spelling mistake?

aslakhellesoy:
That is defiantly a speling misteak!

mpkorstanje:
Nah just a stuter. :stuck_out_tongue:

mpkorstanje:
Bit annoying. IDEA can't doesn't understand `ClasspathResourceSource` locations. Only seems to understand `FileSource` locations.

mpkorstanje:
Ah. Looks like it simply hasn't been implemented: https://github.com/JetBrains/intellij-community/blob/master/plugins/junit5_rt/src/com/intellij/junit5/JUnit5TestExecutionListener.java#L363

aaronols:
I think it would be worth it to open a feature request on the IntelliJ bug tracker to support this:

https://youtrack.jetbrains.com/issues/IDEA

I'm sure there are more frameworks that would benefit from the ability to link from log output lines to classpath resources.

mpkorstanje:
I was thinking about that but might be easier to provide a `CompositeTestSource` with a classpath source and file source.

I'm not sure what the intended use of `CompositeTestSource` is.

aaronols:
`CompositeTestSource` might actually be the way to go:

- "A CompositeTestSource contains one or more TestSources." https://junit.org/junit5/docs/current/api/org/junit/platform/engine/support/descriptor/CompositeTestSource.html
- There is a method to create a `CompositeTestSource` from a list of `TestSource`
https://junit.org/junit5/docs/current/api/org/junit/platform/engine/support/descriptor/CompositeTestSource.html#from(java.util.Collection)
- "[TestSource] is marker interface" and has implementations for all kinds of test sources, including `ClasspathResourceSource`
https://junit.org/junit5/docs/current/api/org/junit/platform/engine/TestSource.html
- A `ClasspathResourceSource` can be created from a `String` or a `URI` and may even have a `FilePosition` that points to the exact line inside the target file.
https://junit.org/junit5/docs/current/api/org/junit/platform/engine/support/descriptor/ClasspathResourceSource.html

This may actually be enough to get the intended functionality working in IntelliJ IDEA.

guenhter:
@mpkorstanje I guess with an own engine, we lose the support for extensions, don't we?

mpkorstanje:
> @mpkorstanje I guess with an own engine, we lose the support for extensions, don't we?

Yes, it will not be possible to use extensions. Though it would be inaccurate to say that we've lost them. JUnit extensions assume that the test context consists of a single class (the one annotated with the extension). Cucumber typically has a test context that consists multiple classes. So it was never possible to use extension in the first place.

Issue #1564, opened 2019-02-21
Windows running feature file throws IllegalArgumentException


jsa34:
## Summary
Running a feature file throws exception:
Exception in thread "main" java.lang.IllegalArgumentException: Unsupported scheme: C:/Users/me/IdeaProjects/automated-tests/src/test/resources/features/subgroup/test.feature
	at cucumber.runtime.io.MultiLoader.resources(MultiLoader.java:28)
	at cucumber.runtime.model.FeatureLoader.loadFromFeaturePath(FeatureLoader.java:41)
	at cucumber.runtime.model.FeatureLoader.load(FeatureLoader.java:35)
	at cucumber.runtime.model.FeatureLoader.load(FeatureLoader.java:21)
	at cucumber.runtime.FeaturePathFeatureSupplier.get(FeaturePathFeatureSupplier.java:23)
	at cucumber.runtime.Runtime.run(Runtime.java:67)
	at cucumber.api.cli.Main.run(Main.java:26)
	at cucumber.api.cli.Main.main(Main.java:8)

This was not a problem in version 4.2.0, but is in 4.2.3.

## Expected Behavior
When running a feature file, it should pick up the classpath.  It is currently not passing any scheme so throws exception

## Current Behavior
In version 4.2.0, when cucumber gets to this method:
```
    public Iterable<Resource> resources(URI path, String suffix) {
        if ("classpath".equals(path.getScheme())) {
            return this.classpath.resources(path, suffix);
        } else if ("file".equals(path.getScheme())) {
            return this.fs.resources(path, suffix);
        } else {
            throw new IllegalArgumentException("Unsupported scheme: " + path);
        }
    }
```
It would pass the arg "path" as "classpath: " followed by the classpath of the variable.

Now, when I evaluate this method, in version 4.2.3 it returns JUST the classpath of the feature file, but not the prefix "classpath: " so wrongly returns for "path.getScheme()" as C because the path = C:/......

## Steps to Reproduce (for bugs)
Run any test in windows

## Your Environment
* Version used: 4.2.3
* Operating System and version: Windows 10
* Link to your project: https://gitlab.developers.cam.ac.uk/jsa34/selenium-java-framework

mpkorstanje:
Thanks for the clear reproducer!

Looks like we're missing a case in [FeaturePath.parse](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/io/cucumber/core/model/FeaturePath.java#L30).

[...]

PR #1568, opened 2019-02-25
[Core] Disambiguate between drive letter and uri scheme

jsa34:
Added windows case where path is being passed with standard file separator in the file location passed.

Fixes: #1564 

## Summary

Update for path in case it is being passed in Windows with the standard file separator, as Intellij appears to do.

### DIFF
--- a/core/src/main/java/io/cucumber/core/model/FeaturePath.java
+++ b/core/src/main/java/io/cucumber/core/model/FeaturePath.java
@@ -3,6 +3,7 @@ package io.cucumber.core.model;
 import java.io.File;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.Locale;
 
 
 /**
@@ -48,7 +49,10 @@ public class FeaturePath {
         return URI.create(featureIdentifier);
     }
     
-    private static boolean isWindowsOS() { return System.getProperty("os.name").contains("Windows");}
+    private static boolean isWindowsOS() { 
+        String osName = System.getProperty("os.name");
+        return normalize(osName).contains("windows");
+    }
     
     private static boolean pathContainsWindowsDrivePattern(String featureIdentifier) {
         return featureIdentifier.matches("^(?:[a-zA-Z]:.*)$");
@@ -81,4 +85,12 @@ public class FeaturePath {
             throw new IllegalArgumentException(e.getMessage(), e);
         }
     }
+    
+    private static String normalize(final String value) {
+        if (value == null) {
+            return "";
+        }
+        return value.toLowerCase(Locale.US).replaceAll("[^a-z0-9]+", "");
+    }
+    
 }
--- a/core/src/main/java/io/cucumber/core/model/FeaturePath.java
+++ b/core/src/main/java/io/cucumber/core/model/FeaturePath.java
@@ -32,6 +32,10 @@ public class FeaturePath {
             String standardized = replaceNonStandardPathSeparator(featureIdentifier);
             return parseAssumeFileScheme(standardized);
         }
+        
+        if (isWindowsOS() && pathContainsWindowsDrivePattern(featureIdentifier)) {
+            return parseAssumeFileScheme(featureIdentifier);
+        }
 
         if (probablyURI(featureIdentifier)) {
             return parseProbableURI(featureIdentifier);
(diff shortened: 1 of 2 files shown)


jsa34:
Tests added

jsa34:
Not sure why this is not happy - running fine in Windows, and assumed overriding the os.name prop would force this to pass through for the test

core/src/main/java/io/cucumber/core/model/FeaturePath.java
```diff
@@ -43,12 +47,17 @@ public static URI parse(String featureIdentifier) {
     private static URI parseProbableURI(String featureIdentifier) {
         return URI.create(featureIdentifier);
     }
-
+    
+    private static boolean isWindowsOS() { return System.getProperty("os.name").contains("Windows");}
```
mpkorstanje: I believe the return value of `System.getProperty("os.name")` should be normalized first using:

```
    private static String normalize(String value) {
        if (value == null) {
            return "";
        }
        return value.toLowerCase(Locale.US).replaceAll("[^a-z0-9]+", "");
}
```

core/src/test/java/io/cucumber/core/model/FeaturePathTest.java
```diff
@@ -82,12 +82,31 @@ public void can_parse_windows_absolute_path_form(){
...
+        assertEquals(file.toURI().getSchemeSpecificPart(), uri.getSchemeSpecificPart());
+    }
+
+    @Test
+    public void can_parse_windows_file_path_with_standard_file_separator_and_non_alpanumeric_chars(){
+        assumeThat(System.getProperty("os.name"), containsString("Windows")); //Requires windows
+        URI uri = FeaturePath.parse("C:/path-to/the_file.feature");
+        assertEquals("file", uri.getScheme());
+        // Use File to work out the drive letter
+        File file = new File("/path-to/the_file.feature");
```
mpkorstanje: I now realize that the test this was copied from declares the current drive letter to be `C`. So File is not needed here, nor in `can_parse_windows_file_path_with_standard_file_separator`, nor in `can_parse_windows_absolute_path_form`.

It would have been useful in a relative path test on Windows but we don't seem to have a test case for that.

jsa34:
Good points!

Thanks - will update :)

jsa34:
Updated! :)

core/src/test/java/io/cucumber/core/model/FeaturePathTest.java
jsa34: Good point - the aim of the test was to make sure that the scheme was being reolved as file, no tthe drive letter.  Updated test accordingly

core/src/main/java/io/cucumber/core/model/FeaturePath.java
jsa34: Good point!  Updated code to normalise as recommended.

core/src/main/java/io/cucumber/core/model/FeaturePath.java
```diff
@@ -43,12 +48,20 @@ public static URI parse(String featureIdentifier) {
...
     }
-
+    
+    private static boolean isWindowsOS() { 
+        String osName = System.getProperty("os.name");
+        return normalize(osName).contains("windows");
+    }
+    
+    private static boolean pathContainsWindowsDrivePattern(String featureIdentifier) {
+        return featureIdentifier.matches("^(?:[a-zA-Z]:.*)$");
```
mpkorstanje: Is there a reason for the capture group?

core/src/test/java/io/cucumber/core/model/FeaturePathTest.java
mpkorstanje: True. But it's good to test the path anyway. The drive letter shouldn't influence it. It should always be `C` because thats what is used in the setup.

mpkorstanje:
I've pushed a few changes. Can you run the tests on a windows machine for me?

jsa34:
Hello - thanks for the updates.

I have run the unit tests can_parse_windows_file_path_with_standard_file_separator and can_parse_windows_path_form - both fail at:
uri.getSchemeSpecificPart() (last assert of each test).

This is returning a prepended / (e.g. /C:/this/is/a/folder.txt)
And the assertion is expecting C:/this/is/a/folder.txt.

jsa34:
It appears this is what is returned at             

FeaturePath.parseAssumeFileScheme

at line 76:
return featureFile.toURI();

jsa34:
there was also an incorrect reference in that test to an extra /the/ in the filepath, which I will now remove, but it still fails due to the prepended /

mpkorstanje:
That makes sense I guess. It is the same as what `File.toUri` would return.

jsa34:
Agreed - now updated assertions for both tests to match output of File.toUri - thanks for your help!

aslakhellesoy:
Hi @jsa34,

Aslak Hellesøy
Creator of Cucumber

mpkorstanje:
Cheers! Thanks for contributing.

PR #1582, opened 2019-03-13
Tests for issue with context caching (#1583)

MarcHauptmann:
## Summary

Cucumber doesn't share its spring context with other junit tests using the same context configuration. Added (failing) tests to reproduce this issue.

## Details

Cucumber uses a ThreadLocal to store a ContextCache for each thread. In contrast the spring test runner uses a static default ContextCache in DefaultCacheAwareContextLoaderDelegate. This leads to creating two spring contexts when running cucumber tests together with plain junit tests.

Tests pass when Line 18 in FixBootstrapUtils is changed to 
```CacheAwareContextLoaderDelegate contextLoader = new DefaultCacheAwareContextLoaderDelegate();```
(calling parameterless constructor).

This is probably not a valid solution as it would break functionality added with #1148.

### DIFF
--- a/spring/src/test/java/cucumber/runtime/java/spring/contextcaching/ContextCachingScenarioTest.java
+++ b/spring/src/test/java/cucumber/runtime/java/spring/contextcaching/ContextCachingScenarioTest.java
@@ -1,0 +1,8 @@
+package cucumber.runtime.java.spring.contextcaching;
+
+import cucumber.api.junit.Cucumber;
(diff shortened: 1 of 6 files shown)


mpkorstanje:
@MarcHauptmann would you mind building https://github.com/cucumber/cucumber-jvm/pull/1604 locally and seeing how that works for you? 

Might be worth noting that it's no longer possible to use constructor dependency injection or to inject step definitions in other step definitions.

PR #1588, opened 2019-03-21
Adding new timestamp field to event and json report

gazler22:
**Summary**
Adding new timestamp field to event and json report

**Detail**
New elapsedTimeInMillis field added to the eventBus as well as the TimeService (and its stub).

This calculates the system time elapsed in millis and is later used in the jsonformatter as a time stamp (after conversion).

This field can be used in all the events for timestamp usages if required (unlike nanos this is an accurate representation of the current timestamp), right now its only being used in TestCaseStarted.

I also updated the JUnits accordinly since the date is calculated dynamically. The JSON formatter takes a DateTimeFormatter input to change how the time will be formatted to string format.

**Motivation and Context**
The timestamp field in the json report allows plugins to correctly calculate the time a TestSuite takes when the TCs are run in parallel. The scenario startTime (which is what i currently added to the report) will be useful in a number of ways both for reporting and getting correct duration of parallel TC execution.

**How Has This Been Tested?**
The current JUnits have been updated to reflect the new timestamp field that is passed on to the report.

**Types of changes**
- [x ] New feature (non-breaking change which adds functionality).

**Checklist:**
I have updated the current TCs but not sure if i need to add new ones solely for testing the addition of this new field. please let me know if i do.

A documentation update is probably not needed for this, but if it is, please let me know and i ll update it.

### DIFF
--- a/core/src/main/java/io/cucumber/core/api/event/TestCaseStarted.java
+++ b/core/src/main/java/io/cucumber/core/api/event/TestCaseStarted.java
@@ -2,10 +2,16 @@ package io.cucumber.core.api.event;
 
 public final class TestCaseStarted extends TestCaseEvent {
     public final TestCase testCase;
+    private final long elapsedTimiMillis; 
 
-    public TestCaseStarted(Long timeStamp, TestCase testCase) {
+    public TestCaseStarted(Long timeStamp, Long elapsedTimiMillis, TestCase testCase) {
         super(timeStamp, testCase);
         this.testCase = testCase;
+        this.elapsedTimiMillis = elapsedTimiMillis;
+    }
+
+    public long getElapsedTimiMillis() {
+        return elapsedTimiMillis;
     }
 
 }
--- a/core/src/main/java/io/cucumber/core/event/EventBus.java
+++ b/core/src/main/java/io/cucumber/core/event/EventBus.java
@@ -7,6 +7,8 @@ public interface EventBus extends EventPublisher {
 
     Long getTime();
 
+    Long getElapsedTimeMillis();
+    
     void send(Event event);
 
     void sendAll(Iterable<Event> queue);
--- a/core/src/main/java/io/cucumber/core/plugin/JSONFormatter.java
+++ b/core/src/main/java/io/cucumber/core/plugin/JSONFormatter.java
@@ -1,21 +1,14 @@
 package io.cucumber.core.plugin;
 
-import io.cucumber.core.api.event.HookTestStep;
-import io.cucumber.core.api.event.HookType;
-import io.cucumber.core.api.event.PickleStepTestStep;
-import io.cucumber.core.api.event.Result;
-import io.cucumber.core.api.event.TestCase;
-import io.cucumber.core.api.event.TestStep;
-import io.cucumber.core.api.event.EmbedEvent;
-import io.cucumber.core.api.event.EventHandler;
-import io.cucumber.core.api.event.EventListener;
-import io.cucumber.core.api.event.EventPublisher;
-import io.cucumber.core.api.event.TestCaseStarted;
-import io.cucumber.core.api.event.TestRunFinished;
-import io.cucumber.core.api.event.TestSourceRead;
-import io.cucumber.core.api.event.TestStepFinished;
-import io.cucumber.core.api.event.TestStepStarted;
-import io.cucumber.core.api.event.WriteEvent;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
 import gherkin.ast.Background;
 import gherkin.ast.Feature;
 import gherkin.ast.ScenarioDefinition;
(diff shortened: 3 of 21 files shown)


core/src/main/java/io/cucumber/core/plugin/JSONFormatter.java
```diff
@@ -212,6 +222,11 @@ private void finishReport() {
         return testCaseMap;
     }
 
+    private String getDateTimeFromTimeStamp(TestCaseStarted event) {
+        LocalDateTime localDateTime = LocalDateTime.ofInstant(Instant.ofEpochMilli(event.getElapsedTimiMillis()), ZoneId.systemDefault());
+        return localDateTime.format(dateTimeFormatter == null ? DateTimeFormatter.ISO_DATE_TIME : dateTimeFormatter);
```
mpkorstanje: Is there a reason the `dateTimeFormatter` is optional?

core/src/main/java/io/cucumber/core/api/event/TestCaseStarted.java
```diff
@@ -2,10 +2,16 @@
...
 
-    public TestCaseStarted(Long timeStamp, TestCase testCase) {
+    public TestCaseStarted(Long timeStamp, Long elapsedTimiMillis, TestCase testCase) {
         super(timeStamp, testCase);
         this.testCase = testCase;
+        this.elapsedTimiMillis = elapsedTimiMillis;
+    }
+
+    public long getElapsedTimiMillis() {
+        return elapsedTimiMillis;
```
mpkorstanje: I don't think `elapsedTimiMillis` is quite right. Perhaps we can rename this to `timeStampMillis`.

core/src/main/java/io/cucumber/core/event/EventBus.java
```diff
@@ -7,6 +7,8 @@
 
     Long getTime();
 
+    Long getElapsedTimeMillis();
```
mpkorstanje: Same here, but `getTimeStampMillis`.

core/src/main/java/io/cucumber/core/runner/TimeService.java
```diff
@@ -2,12 +2,18 @@
 
 public interface TimeService {
     long time();
+    long elapsedTimeMillis();
```
mpkorstanje: Same here `timeMillis`.

core/src/test/java/io/cucumber/core/runner/StepDurationTimeService.java
```diff
@@ -35,4 +35,9 @@ private void handleTestStepStarted(TestStepStarted event) {
         long time = time();
         currentTime.set(time + stepDuration);
     }
+
+    @Override
+    public long elapsedTimeMillis() {
+        return 0L;
```
mpkorstanje: This should have a similar implementation to `time`.

core/src/main/java/io/cucumber/core/plugin/JSONFormatter.java
mpkorstanje: For the JSON output the format should be fixed predictable. So I would think the zone should always be UTC. Using a `ZonedDateTime` may be more appropriate but I'm not sure.

mpkorstanje: And probably good to bear in mind that v4 is still using Java 7. So you can't use the fancy Date stuff yet.

gazler22: And well Java 7 puts away both the localdatetime and zonedatetime options away.
I ll change it to the usual Date.

And shouldn't it just be the default zone of the PC running the TCs? In case they use it to see the timestamp of the TC run in the report.

core/src/test/java/io/cucumber/core/runner/StepDurationTimeService.java
gazler22: > 
> 
> Looks good!
> 
> Just to be sure, you've made a PR against `develop-v5`. This version is still a bit away. You could also make the PR against `master` and we could pick it up in the next release of v4.

Ok. I will create a new PR for master. I wasn't sure if that's how it should have been so i raised one against the develop branch. But yes, its probably better to have this in v4 itself.

core/src/main/java/io/cucumber/core/event/EventBus.java
gazler22: its the elapsed time since epoch in millis so i named it that way. But i ll update the name no problem.

core/src/main/java/io/cucumber/core/plugin/JSONFormatter.java
gazler22: I kept the dateTimeFormatter optional for the sake of flexibility in case we dont want the full time for some TC (we just want the date, etc).

Its actually also being used in the JUnits to bypass the 'dynamic generation' issue, for eg in `RuntimeTest`, `JsonParallelRuntimeTest`. I can make it constant but I ll need some suggestion on how to update the JUnits then. Do u have an idea?

mlvandijk: I'd prefer not to depend on the default timezone of the machine running the tests, as this might be either a local machine with locally running application, or a CI server (i.e. Jenkins) running tests against an actual environment; where the CI server and the test environment might not have the same system datetime / timezone. Being able to specify it is preferable inho

gazler22: Yes that makes sense. I can change it to always be UTC in that case.

But do u mean allow the user to specify what timezone to use before running the tests? From cucumber options or the like?

gazler22:
Closing this PR to raise a new one on master. Will update the changes there.

PR #1591, opened 2019-03-24
[Core] Add timestamp milis to TestCaseStartedEvent and json reporter

gazler22:
Adding new PR for the master branch instead of develop-v5.

Refer :
#1588 

Updated the changes with new name for the field and code changes compatible with Java 7.

Also updated the `JSONFormatter` constructor to include a TimeUtils object for getting the timestamp in string ISO format. This also helps make the timeStamp field testable using mocks.

We can rename the TimeUtils class name to something more generic (as a utility class for the JSONFormatter) for future updates if more util methods are added which need to be mocked for the jsonformatter tests.

Please let me know if any more changes are needed.


core/src/main/java/cucumber/runtime/formatter/JSONFormatter.java
```diff
@@ -48,7 +49,8 @@
     private final Gson gson = new GsonBuilder().setPrettyPrinting().create();
     private final NiceAppendable out;
     private final TestSourcesModel testSources = new TestSourcesModel();
-
+    private final TimeUtils timeUtils;
```
mpkorstanje: I don't think this is needed. The event bus is constructed with a time provider that allows tests to receive deterministic time (kinda like `Clock` from Java 8). If you implement `TimeServiceStub.timeStampMillis` in the same way as `TimeServiceStub.timeStamp`.

core/src/main/java/cucumber/api/event/TestCaseStarted.java
```diff
@@ -4,10 +4,15 @@
 
 public final class TestCaseStarted extends TestCaseEvent {
     public final TestCase testCase;
+    private final long timeStampMillis; 
 
-    public TestCaseStarted(Long timeStamp, TestCase testCase) {
+    public TestCaseStarted(Long timeStamp, Long timeStampMillis, TestCase testCase) {
```
mpkorstanje: Please add a deprecated `public TestCaseStarted(Long timeStamp, TestCase testCase)` constructor that calls `this(timeStamp, 0, testCase)`. The `TestCaseStarted` event is part of the public API so we should deprecate the constructor before removing it.

core/src/main/java/cucumber/runner/TimeService.java
```diff
@@ -2,12 +2,19 @@
...
+    long timeStampMillis();
 
     TimeService SYSTEM = new TimeService() {
         @Override
         public long time() {
             return System.nanoTime();
         }
+
+        @Override
+        public long timeStampMillis()
```
mpkorstanje: K&R brackets. You can probably configure this in your IDE on a project basis.

mpkorstanje:
Almost there!

core/src/main/java/cucumber/runtime/formatter/JSONFormatter.java
gazler22: I imlemented timeStampMillis in TimeServiceStub as well but there are a number of JUnits which don't pass an eventBus to the Runner.build() because they do not expect the 'duration' field in the response. But now they would have the timeStampMillis field since its at the scenario level.

For eg : **RuntimeTest.runs_feature_with_json_formatter()**

For these Junits I had to add the TimeUtils so i could mock it. Should I update the JUnits instead to use an eventBus with TimeServiceStub?

laxersaz:
I am seriously excited about this :-)

core/src/main/java/cucumber/api/event/TestCaseStarted.java
gazler22: Thanks that's useful knowledge :) Updated.

core/src/main/java/cucumber/runner/TimeService.java
gazler22: Really sorry about that. Updated.

core/src/main/java/cucumber/runtime/formatter/JSONFormatter.java
mpkorstanje: Please do! Unless it's too much work. Then we can look into making it easier to use the time service.

gazler22: No I was actually hoping I could do that :) Just wanted to make sure there's no issue if i update them instead.

Let me check and I ll udpate.

gazler22: Done

core/src/test/java/cucumber/runtime/formatter/FormatterBuilder.java
```diff
@@ -1,8 +1,10 @@
 package cucumber.runtime.formatter;
 
+
```
mpkorstanje: This file only has whitespace changes.

core/src/main/java/cucumber/runtime/formatter/JSONFormatter.java
```diff
@@ -379,4 +387,11 @@ private void addEmbeddingToHookMap(byte[] data, String mimeType) {
         }
         return resultMap;
     }
+    
+    private String getDateTimeFromTimeStamp(long timeStampMillis) {
+        SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.XXX");
```
mpkorstanje: This seems to result in dates that look like `1970-01-01T00:00:00.Z`. I think you meant to use `SSS` for milliseconds.

gazler22: Ah yes, you are right. I ll update this.

core/src/test/java/cucumber/runtime/formatter/FormatterBuilder.java
gazler22: I updated this file to be the same from the master. It still shows a diff as a bracket position change. M not sure how to remove that diff if thats what was needed.

mpkorstanje: The old file ended with a new line. That's the only difference left now.

gazler22: Updated :)

sorry it took this long to get this resolved.

mpkorstanje:
Cheers!

PR #1593, opened 2019-03-25
chore: [Core] Improve MessageDigest Algorithm

askrakesh:
## Types of changes

## Checklist:


mpkorstanje:
Thanks, but could you explain why? What problem is being solved and is this the best way to solve it?

PR #1594, opened 2019-03-29
[Core] Add TimeStampMillis to TimeStampedEvent

mpkorstanje:
## Summary

A timestamp is more useful when that timestamp is relative to the epoch.
Unlike `System.currentTimeMillis`, `System.nanoTime` is relative to some
arbitrary point in time.

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/api/event/TimeStampedEvent.java
+++ b/core/src/main/java/cucumber/api/event/TimeStampedEvent.java
@@ -1,14 +1,30 @@
 package cucumber.api.event;
 
 abstract class TimeStampedEvent implements Event {
-    private final Long timeStamp;
 
-    TimeStampedEvent(Long timeStamp) {
+    private final Long timeStamp;
+    private final long timeStampMillis;
+
+    TimeStampedEvent(Long timeStamp, Long timeStampMillis) {
         this.timeStamp = timeStamp;
+        this.timeStampMillis = timeStampMillis;
     }
 
+    /**
+     * {@inheritDoc}
+     */
     @Override
     public Long getTimeStamp() {
         return timeStamp;
+    }
+
+    /**
+     * Returns timestamp in milliseconds of the epoch.
+     *
+     * @return timestamp in milli seconds
+     * @see System#currentTimeMillis()
+     */
+    public long getTimeStampMillis() {
+        return timeStampMillis;
     }
 }
--- a/core/src/main/java/cucumber/runtime/formatter/JSONFormatter.java
+++ b/core/src/main/java/cucumber/runtime/formatter/JSONFormatter.java
@@ -382,14 +382,14 @@ final class JSONFormatter implements EventListener {
         if (result.getErrorMessage() != null) {
             resultMap.put("error_message", result.getErrorMessage());
         }
-        if (result.getDuration() != null && result.getDuration() != 0) {
+        if (result.getDuration() != 0) {
             resultMap.put("duration", result.getDuration());
         }
         return resultMap;
     }
     
     private String getDateTimeFromTimeStamp(long timeStampMillis) {
-        SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS");
+        SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX");
         sdf.setTimeZone(TimeZone.getTimeZone("UTC"));
         
         return sdf.format(new Date(timeStampMillis));
--- a/core/src/test/java/cucumber/runtime/formatter/JSONFormatterTest.java
+++ b/core/src/test/java/cucumber/runtime/formatter/JSONFormatterTest.java
@@ -45,7 +45,7 @@ public class JSONFormatterTest {
     private final List<SimpleEntry<String, Result>> hooks = new ArrayList<>();
     private final List<String> hookLocations = new ArrayList<>();
     private final List<Answer<Object>> hookActions = new ArrayList<>();
-    private Long stepDuration = 0L;
+    private Long stepDurationMillis = 0L;
 
     @Test
     public void featureWithOutlineTest() {
@@ -89,7 +89,7 @@ public class JSONFormatterTest {
             "      {\n" +
             "        \"id\": \"banana-party;monkey-eats-bananas\",\n" +
             "        \"keyword\": \"Scenario\",\n" +
-            "        \"start_timestamp\": \"1970-01-01T00:00:00.000\",\n" +
+            "        \"start_timestamp\": \"1970-01-01T00:00:00.000Z\",\n" +
             "        \"name\": \"Monkey eats bananas\",\n" +
             "        \"line\": 3,\n" +
             "        \"description\": \"\",\n" +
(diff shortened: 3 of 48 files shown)


core/src/test/java/cucumber/runtime/formatter/JSONFormatterTest.java
```diff
@@ -1241,7 +1241,7 @@ private String runFeaturesWithFormatter() {
         return report.toString();
     }
 
-    private Long milliSeconds(int milliSeconds) {
-        return milliSeconds * 1000000L;
+    private long milliSeconds(int milliSeconds) {
+        return milliSeconds;
```
gazler22: I think we can just use `Long.valueOf()` in the callers instead of this method now.

core/src/main/java/cucumber/runtime/formatter/JSONFormatter.java
```diff
@@ -382,14 +384,14 @@ private void addEmbeddingToHookMap(byte[] data, String mimeType) {
         if (result.getErrorMessage() != null) {
             resultMap.put("error_message", result.getErrorMessage());
         }
-        if (result.getDuration() != null && result.getDuration() != 0) {
-            resultMap.put("duration", result.getDuration());
+        if (result.getDuration() != 0) {
+            resultMap.put("duration", MILLISECONDS.toNanos(result.getDuration()));
```
gazler22: Do we still need nanos?

We have lost the precision of nanos anyway, and `toNanos()` basically scales it by **1000000** i think. We could just use millis if there isnt a specific reason for using nanos.

gazler22:
Loos pretty good :) I like that we are using the timeStampMillis in all the other places as well now, it would probably be super useful.

How do u feel about just sending the end and start times in the report and leaving the calculation of the duration to the system using the report instead? (since we have added start time anyway)

core/src/main/java/cucumber/api/event/TimeStampedEvent.java
```diff
@@ -1,14 +1,36 @@
...
+    private final long timeStampMillis;
 
-    TimeStampedEvent(Long timeStamp) {
+    TimeStampedEvent(Long timeStamp, Long timeStampMillis) {
         this.timeStamp = timeStamp;
+        this.timeStampMillis = timeStampMillis;
     }
 
+    /**
+     * Returns timestamp in nano seconds since an arbitrary start time.
```
mlvandijk: Any way to make this clearer? I am now wondering what "arbitrary start time" means...

mlvandijk: Should the JavaDoc here be consistent with the JavaDoc above? "Returns a timestamp in nano seconds."

core/src/main/java/cucumber/api/event/TimeStampedEvent.java
```diff
@@ -1,14 +1,36 @@
...
+     *
+     * @return timestamp in nano seconds
+     */
     @Override
     public Long getTimeStamp() {
         return timeStamp;
     }
+
+    /**
+     * Returns timestamp in milliseconds of the epoch.
```
mlvandijk: I like that this one is consistent with the other one ;)

core/src/main/java/cucumber/api/event/TimeStampedEvent.java
mlvandijk: (Or should it be removed if we are using millis everywhere?)

aslakhellesoy:
> A timestamp is more useful when that timestamp is relative to the epoch.

Useful for what exactly? There are two broad categories:
* Knowing how long something took (24 minutes)
* Knowing when something happened (12 Oct 1957)

Nano time is often best for duration, Millis time is often best for time. See https://stackoverflow.com/questions/351565/system-currenttimemillis-vs-system-nanotime

What are we trying to accomplish with these timestamps?

aslakhellesoy:
I agree we don’t need the nano precision and happy to switch to milliseconds to measure elapsed time. I’m not sure we need to store those times as duration (relative to some start time) or absolute (time since epoch). I’d like to understand what the timestamps are used for first.

core/src/main/java/cucumber/runtime/formatter/JSONFormatter.java
mpkorstanje: Yes. The consumers of the JSON report would still expect the time to be expressed in nano seconds.

gazler22:
@aslakhellesoy The initial reason was to allow reports to know the **concurrent start time** of when a test case (or scenario) started, to allow them to accurately display the actual time it took to run a suite when running TCs in parallel (if we just give the duration, they do not know which TCs were run in parallel).

core/src/main/java/cucumber/runtime/formatter/JSONFormatter.java
gazler22: We could deprecate it to allow consumers to migrate, if that seems feasible.

mpkorstanje:
> What are we trying to accomplish with these timestamps?

1. Correct the current implementation. Timestamps indicate a point in time. This makes nano time the wrong choice for use in `TimeStampedEvent`.

   In the long run we should replace `TimeService` with the java standard `Clock` and `long timeStamp` with `Instant` and `long duration` with `Duration` as these more accurately represent the concepts used here. 

2. Allow cucumbers events to be correlated with other events in other systems. When analyzing test failures of a deployed system it is useful to know when an event happened. Most log systems allow you to filter by a time range. Currently this information is not accessible.

     @laxersaz (the author of Cluecumber) already expressed interest in including this into his reports.

3. Allows execution of multiple (concurrent) scenarios from different runners to be aggregated and displayed in real time.

mpkorstanje:
@gazler22 
> How do u feel about just sending the end and start times in the report and leaving the calculation of the duration to the system using the report instead? (since we have added start time anyway)

Would be nice but we have some existing systems to support. It's also kinda convenient.

mpkorstanje:
I'm going to revert the use of milliseconds for duration. While test cases last milliseconds, it seems that individual steps are more granular then that.

core/src/main/java/cucumber/api/event/TimeStampedEvent.java
mpkorstanje: I don't think I can be much more clear then "arbitrary start time". Arbitrary X means pick an X, we don't care which one. From the linked `System.nanoTime` documentation.

> Returns the current value of the running Java Virtual Machine's high-resolution time source, in nanoseconds. This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time. The value returned represents nanoseconds since some fixed but arbitrary <i>origin</i> time (perhaps in the future, so values may be negative).  The same origin is used by all invocations of this method in an instance of a Java virtual machine; other virtual machine instances are likely to use a different origin.

Can't remove it because it's part of the API so existing plugins will expect it to be present.

mpkorstanje: Good point though. I should have used `{@inheritDoc}`.

core/src/main/java/cucumber/runtime/formatter/JSONFormatter.java
mpkorstanje: I've reverted the use of milliseconds for duration. Individual steps are still quite close to 1ms.

Issue #393, opened 2012-09-14
JUnit Rule annotation is not supported


rlogiacco:
The JUnit Rule annotation support would be very useful to allow to take snapshots on failing tests, but the Cucumber JUnit runner does not support such annotation.
I think that it should be easy to add this support by modifying the runners and introducing the JUnit Statement class usage: I've forked the project and started working on it: please let me know if this is an unwanted feature.

markwinspear:
Hi. I came across this thread because I am trying to put together a simple framework using a Page Object Model patten using Webdriver and using Cucumber as the BDD framework. I'm not a developer first and foremost, however I have taken the advice of Dave Haeffner in his book 'Selenium Guidebook for Java' in setting up a Base test class which uses JUnit's rule to set up an external resource and set capabilities then fire up a browser either locally or in saucelabs.

I have conflict between this and my cucumber annotations and cannot get the two to play nicely together when my base test class is extended by my SearchTestDefinitions class.  I suspect that if the cucumber api supported JUnit rule, I would not get this issue (but as I say, I am not a developer by nature). 

Any advice on how to combine the two if possible would be welcomed

The code is at https://github.com/markwinspear/SeleniumCucumberJavaBasicTest2

Many thanks
Mark

mikeholler:
Rules are important for Android. See #894.

johndeverall:
I just came across this thread - see Serenity issue https://github.com/serenity-bdd/serenity-core/issues/221

It seems this feature would be very helpful for Spring integration, and a lot of people use Spring in the JVM world. Is there any chance this could be looked at again?

alb-i986:
+1

I think the solution would be to have `cucumber.api.junit.Cucumber` extend `BlockJUnit4ClassRunner` instead of `ParentRunner`. In fact, BlockJUnit4ClassRunner encapsulates the Rules handling logic.
I'll see if I can find some time working on it.

johndeverall:
I'm also wanting to put some time into this but haven't looked into it since I last commented as I've been away on holiday. @alb-i986 I'll take a look at what you're suggesting. It sounds like you're ahead of me in conceptualizing a solution. I'll be able to comment more once I take a fresh look at the code.

alb-i986:
@aslakhellesoy 
Would you agree on having `cucumber.api.junit.Cucumber` extend `BlockJUnit4ClassRunner` instead of `ParentRunner`? Can you foresee any issues in doing that?
Thanks

[...]

Issue #1470, opened 2018-09-26
Use TestContextManager in cucumber-spring


mpkorstanje:
## Summary

Support `TestContextManager` in `cucumber-spring`.

## Expected Behavior

When using `cucumber-spring` `@MockBean` (and other annotations) should just work.

```java
public class Example {

    @MockBean
    private User object;

    @Given("there is a User")
    public void there_is_a_User() {
        when(object.toString()).thenReturn("I've been mocked");
    }

    @Then("the User is mocked")
    public void the_user_is_mocked() throws Exception {
        assertThat(object.toString(), equalTo("I've been mocked"));
    }
}
```

## Current Behavior

It is currently possible to use `@Autowired` e.g.:

```java
public class Example {

    @Autowired
    private User object;

    @Then("the user is real")
    public void the_user_is_mocked() throws Exception {
        assertThat(object.toString(), equalTo("I am real"));
    }
}
```
But fields annotated with`@MockBean` are not mocked, the example fails with a null pointer exception.

## Possible Solution

The causes of failure are two fold:

1. There is no spring context declared (`ContextConfiguration`, `BootstrapWith`, ect)  so the `SpringFactory` falls back to some minimal default. This can be resolved by adding `@SpringBootTest` to the step definition.

2. The [SpringFactory](https://github.com/cucumber/cucumber-jvm/blob/master/spring/src/main/java/cucumber/runtime/java/spring/SpringFactory.java) does not call  
 * `TestContextManager.beforeTestClass()` 
 * `TestContextManager.prepareTestInstance()`.  
 * `TestContextManager.beforeTestMethod()` 
 * `TestContextManager.beforeTestExecution()` 
 * `TestContextManager.afterTestExecution()` 
 * `TestContextManager.afterTestMethod()` 
 * `TestContextManager.afterTestClass()` 

Instead step definitions are registered as beans in the Spring context. This had the advantage that step definitions can be auto wired into each other and constructor dependency injection just works. However it also puts the step definitions into the spring context which complicates things.

Given that step definitions calling each other is not a best practice I would not mind step definitions were no longer part of the spring context and instead treated as test instances (see [SpringJUnit4ClassRunner.createTest](https://github.com/spring-projects/spring-framework/blob/master/spring-test/src/main/java/org/springframework/test/context/junit4/SpringJUnit4ClassRunner.java#L225)). Any dependency injection would have to be done via `@Autowired`.

So to implement this properly the following would have to be done:
[...]

PR #1598, opened 2019-04-04
Make Cucumber runner extend BlockJUnit4ClassRunner

nicerloop:
## Summary

This is a Proof-of-Concept to improve cucumber-jam integration with junit 4 by deriving Cucumber runner from BlockJUnit4ClassRunner in order to benefit from natural integration with unit 4 features like scenario rules (https://github.com/cucumber/cucumber-jvm/issues/393) and natural integration with external frameworks (http://pitest.org, https://www.testcontainers.org, https://docs.spring.io/spring/docs/5.1.6.RELEASE/spring-framework-reference/testing.html#testcontext-junit4-rules)

## Details

The change proven here is to derive Cucumber unit runner from BlockJUnit4ClassRunner rather than from ParentRunner, which requires in this version the make the FeatureRunner disappear to directly host scenarios/Pickles at the same level as scenarios/methods in classical unit tests.

The current implementation is really a quick and dirty proof that is can work.

## Types of changes

- [?] New feature (non-breaking change which adds functionality).
- [?] Breaking change (fix or feature that would cause existing functionality to not work as expected).

## Checklist:


mpkorstanje:
I'm impressed by what you've managed to do.

It is unfortunately quite hacky. I mean that in the sense the concepts involved are misaligned. The `BlockJUnit4ClassRunner` makes several assumptions that are not true for Cucumber. To wit it assume that a single method is a single test and that each method is executed on a test context which consists of a single class. 

Neither is true for Cucumber. As a result of this we run into quite a few complications that would make JUnit rules work quite differently from how they'd normally would be expected to work. To wit a brief list of all problem areas:

1. Steps do not have access to the rule instance. It's part of the JUnit test context which is not accessible from Cucumber. This is a requirement for most JUnit rules to work properly (e.g. `TemporaryFolder`). 

2. JUnit rules assume that any exceptions thrown by the method under test bubble up through the hierarchy of `Statements` where it can be caught (e.g. `ExpectedException`). This is again not true for Cucumber. All exceptions thrown in a step will be caught by the runner.

3. Cucumber is generally expected to be executable by different runners. Allowing on runner to do more things then others would results in quite some unexpected situations. For instance running with The IDEA Cucumber Plugin (which uses the CLI) will have different results then when using JUnit. This is unfortunately already the case with `@BeforeClass` and `@AfterClass` and as such their use is discouraged.

So instead of trying to push all the square pegs into round holes I think it would be better to let the user deal with it. For most cases they an do so by initializing the JUnit rule in a step definition and invoke the rules before and after methods by using Cucumbers before and after hooks. It's not quite perfect but it keeps the ambiguity out of the framework.

One exception here is Spring. We currently do support Spring through the `cucumber-spring` module. Unfortunately this is support is less then perfect. Our `SpringObjectFactory` does not use the `TestContextManager` provided by Spring. As a results not all annotations work. There is an open issues (https://github.com/cucumber/cucumber-jvm/issues/1470) to resolve this if you'd be interested in doing so.

That said. I would like to be able to somewhat properly integrate JUnit Rules with Cucumber. To do so however we (much like the JUnit 5 project) would need to make our test context an explicit concept in the framework. This is a non-trivial under taking.

nicerloop:
I was well aware that it is really a hack, just to confirm it is doable. After discussions with my colleagues, I concluded the appropriate mapping between cucumber and junit4 is as follows: Cucumber / Suite, FeatureRunner / BlockJUnit4ClassRunner, PickleRunner / FrameworkMethod.

I implemented this mapping, which requires completely removing the step description feature. This does not seem to be that problematic, as it has been identified to be removed in https://github.com/cucumber/cucumber-jvm/issues/263#issuecomment-14525266 . 

As for the problems you identified, they are real but:

1) junit context (test class) not directly accessible from the cucumber steps, is the same problem as cucumber state management, which can be worked-around by a world object or dependency injection

2) exceptions caught by the runner is not a problem for me for now, as I don't want to use rules for scenario control, but rather for external state setup and teardown in integration tests: temporary folder, database docker container, spring test rules...

3) cucumber may be executed by different runners, but the junit ecosystem is much larger than the cucumber one, and I want to reuse all those junit extensions modelled as rules to pimp my cucumber tests, rather than write and maintain integrations by copy-pasting what is already done, for no effective gain; junit rules and class rules complement cucumber hooks, as @around and global hooks are not available

Just have a look at the pitest extension for cucumber https://github.com/alexvictoor/pitest-cucumber-plugin which contains much copy-paste to upgrade dependencies versions, whereas this improved integration between cucumber and junit brings directly the integration with pitest for no additional cost.

As for using rules from the before and after hooks, not all rules allow that, as the rule design is to decorate statements. In fact, it is not possible with the spring rules. And as the cucumber-spring modules does not provide integration with the testcontextmanager, using junit rules as designed is the easiest way for me.

I really hope to be able to contribute this evolution, as it effectively allows my teams to use cucumber with spring applications, mutation testing and docker containers.

mpkorstanje:
As it stands in v4 the dependency injection context is an implementation detail of the `JavaBackend`. You may be able to apply a decorator pattern with v5 (see the `develop-v5` branch) but I imagine it would be quite clunky.

I don't find your cost argument to be particularly convincing. Your proposed changes would incur additional complexity to Cucumber to allow for functionality that has a limited application. Additionally because this functionality doesn't quite work as expected it will be a constant source of tickets, questions, confusion and complaints. It then seems only appropriate that both the cost and benefit of this functionality are in the same place.

So I don't think that your current approach will be fruitful and I would recommend that you maintain and publish your own runner. We can make this easier by reducing the amount of boiler plate needed though! Comparing `Runtime` with `Cucumber`, `TestNGCucumberRunner` and `CucumberTestUnitFinder` I think there should be some options.

PR #1600, opened 2019-04-07
[Core] Add more details to ParserException

gazler22:
## Summary
Adding the path to the resource for easier investigation during parsing exceptions in cucumber.

## Details
Added resource path to the exception thrown under `FeatureParser.parseResource()` method to allow user to see which file failed to parse.

## Motivation and Context
Better user experience. It is useful to know which file caused the parser exception. Refer slack conversation 

> Nicolas   [Mar 30th at 3:03 AM]
> Hello! I'm starting off with cucumber-jvm and I've noticed that the Parser errors don't specify which `.feature` file is broken. Is there any way to enable this behavior? We're looking at adding multiple feature files at a time in a medium sized QA team and this would help out quite a bit. Thanks!
> 
> mpkorstanje   [8 days ago]
> And if you're willing to invest a little time you could probably submit a patch too:
> 
> You'd have to add some details around here I think:
> 
> core/src/main/java/cucumber/runtime/model/FeatureParser.java:40
>             ```throw new CucumberException(e);```

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/runtime/model/FeatureParser.java
+++ b/core/src/main/java/cucumber/runtime/model/FeatureParser.java
@@ -37,7 +37,7 @@ public class FeatureParser {
             List<PickleEvent> pickleEvents = compilePickles(gherkinDocument, resource);
             return new CucumberFeature(gherkinDocument, path, source, pickleEvents);
         } catch (ParserException e) {
-            throw new CucumberException("Failed to parse resourse : " + path.getPath(), e);
+            throw new CucumberException("Failed to parse resourse at: " + path.toString(), e);
         }
     }
 
--- a/core/src/main/java/cucumber/runtime/model/FeatureParser.java
+++ b/core/src/main/java/cucumber/runtime/model/FeatureParser.java
@@ -37,7 +37,7 @@ public class FeatureParser {
             List<PickleEvent> pickleEvents = compilePickles(gherkinDocument, resource);
             return new CucumberFeature(gherkinDocument, path, source, pickleEvents);
         } catch (ParserException e) {
-            throw new CucumberException("Failed to parse resourse at: " + path.toString(), e);
+            throw new CucumberException("Failed to parse resource at: " + path.toString(), e);
         }
     }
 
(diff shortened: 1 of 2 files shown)


junit/src/test/java/cucumber/runtime/junit/CucumberTest.java
```diff
@@ -68,7 +67,7 @@ public void testThatParsingErrorsIsNicelyReported() throws Exception {
             new Cucumber(LexerErrorFeature.class);
             fail("Expecting error");
         } catch (CucumberException e) {
-            assertThat(e.getMessage(), startsWith("gherkin.ParserException$CompositeParserException: Parser errors:"));
+            assertThat(e.getCause().toString(), containsString("gherkin.ParserException$CompositeParserException: Parser errors:"));
```
mpkorstanje: The try-catch surounding this can be replaced with JUnits [ExpectedException](https://junit.org/junit4/javadoc/4.12/org/junit/rules/ExpectedException.html) rule. It reads a bit nicer.

core/src/main/java/cucumber/runtime/model/FeatureParser.java
```diff
@@ -37,7 +37,7 @@ public static CucumberFeature parseResource(Resource resource) {
             List<PickleEvent> pickleEvents = compilePickles(gherkinDocument, resource);
             return new CucumberFeature(gherkinDocument, path, source, pickleEvents);
         } catch (ParserException e) {
-            throw new CucumberException(e);
+            throw new CucumberException("Failed to parse resourse at: " + path.toString(), e);
```
mpkorstanje:
```suggestion
            throw new CucumberException("Failed to parse resource at: " + path.toString(), e);
```

gazler22: Oh sorry about that.

junit/src/test/java/cucumber/runtime/junit/CucumberTest.java
gazler22: udpated

core/src/main/java/cucumber/runtime/model/FeatureParser.java
mpkorstanje: No worries! I make lots of typos too.

mpkorstanje:
Cheers!

core/src/main/java/cucumber/runtime/model/FeatureParser.java
gazler22: :)

PR #1604, opened 2019-04-07
[Spring] WIP: Use TestContextManager 

mpkorstanje:
## Summary

Spike for fixing #1582, #1582 and  #1470.

Changes:
 * Removes spring transaction hooks. Steps annotated with `@txn` will no longer have a managed transaction. It's recommended to implement this yourself.
 * Removes dependency on `org.springframework:spring-tx`
 * Removes step definition injection. Injecting step definitions into step definitions was already discouraged. The use of glue code is recommended. (Not sure if this is required)
 * Removes constructor dependency injection for step definitions. (Not sure if this is required).

TODO:
 * [ ] Deprecate step definition constructor dependency injection 
 * [ ] Deprecate `txn` steps in cucumber v4
 * [ ] Deprecate the use of `appliction.xml`
 * [ ] Handle absence of annotated class
 * [ ] Fix all the docs! 
 * [ ] Try to bring back constructor dependency injection for step definitions


mpkorstanje:
I'm giving up on this exact appraoch for now #1848 was a much better fix for the parallel execution problem.

PR #1608, opened 2019-04-08
[Core] Refactore usage formatter

mlvandijk:
## Summary

Refactor Usage Formatter, as discussed
Fix bug with premature rounding of calculations.

## How Has This Been Tested?
Added tests for refactoring.
Added tests to reproduce bug and fix it.

## Types of changes

## Checklist:


mpkorstanje:
Starting to look pretty good! Do you think you can make the the containers immutable?

mlvandijk:
I'll see what I can do.

mpkorstanje:
@mlvandijk I've had a look. I think I figured out why you got stuck. It wasn't possible to make the containers immutable without consolidating the two `toSeconds` method invocations first. I've fixed those now.

There is just one thing left to do. `UsageFormatter.usageMap` is package private for testing purposes. It need not be. As you've noticed yourself it is perfectly possible to test this plugin via its inputs (events) and it's out put (json). You can use `assertThat(out.toString(), sameJSONAs(json));` to help with comparing json. While you probably won't immediately see the need for it `sameJSONAs` will ignore attribute order. This is useful as the attribute order is not guaranteed to be the same on every machine.

mpkorstanje:
Also I rebased on master and force pushed that. `git pull --rebase` should solve that for you.

mpkorstanje:
Good enough for now.

mlvandijk:
Sorry, I didn't have time for a while. Thanks for continuing without me.

PR #1615, opened 2019-04-28
[Java] Use ServiceLoader for ObjectFactory

toepi:
backport of commit a03dd01 for cucumber 4.x

## Summary

Backport ServiceLoader for ObjectFactory to 4.x.

## Details
So it is possible to use ServciceLoader for ObjectFactory also with 4.x and must not be change when changed to 5.x. Easier migration to next major-version.

## How Has This Been Tested?

Difficult to test, have not found any test in develop-v5 for it .. and not sure how to provide a usefull test. Had test it with my own ObjectFactory (not public :( )

## Types of changes

## Checklist:

### DIFF
--- a/java/src/main/java/cucumber/api/java/ObjectFactory.java
+++ b/java/src/main/java/cucumber/api/java/ObjectFactory.java
@@ -2,7 +2,10 @@ package cucumber.api.java;
 
 /**
  * Minimal facade for Dependency Injection containers
+ *
+ * @deprecated use @code io.cucumber.core.backend.ObjectFactory instead
  */
+@Deprecated
 public interface ObjectFactory {
 
     /**
--- a/java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
+++ b/java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
@@ -13,10 +13,15 @@ import java.net.URI;
 import java.util.List;
 
 import static java.util.Arrays.asList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Objects;
+import java.util.ServiceLoader;
 
-public class ObjectFactoryLoader {
+public final class ObjectFactoryLoader {
 
-    private static final Logger log = LoggerFactory.getLogger(ObjectFactoryLoader.class);
+    private static final Logger LOG = LoggerFactory.getLogger(ObjectFactoryLoader.class);
 
     private ObjectFactoryLoader() {
     }
@@ -30,27 +35,50 @@ public class ObjectFactoryLoader {
      * @return an instance of {@link ObjectFactory}
      */
     public static ObjectFactory loadObjectFactory(ClassFinder classFinder, String objectFactoryClassName) {
-        ObjectFactory objectFactory;
         try {
-            Reflections reflections = new Reflections(classFinder);
-
-            if (objectFactoryClassName != null) {
-                Class<ObjectFactory> objectFactoryClass = (Class<ObjectFactory>) classFinder.loadClass(objectFactoryClassName);
-                objectFactory = reflections.newInstance(new Class[0], new Object[0], objectFactoryClass);
+            final Reflections reflections = new Reflections(classFinder);
+            if (objectFactoryClassName == null) {
+                return loadSingleObjectFactory(reflections);
             } else {
-                List<URI> packages = asList(URI.create("classpath:cucumber/runtime"));
-                objectFactory = reflections.instantiateExactlyOneSubclass(ObjectFactory.class, packages, new Class[0], new Object[0], null);
+                return loadSelectedObjectFactory(reflections, classFinder, objectFactoryClassName);
             }
         } catch (TooManyInstancesException e) {
-            log.warn(e.getMessage());
-            log.warn(getMultipleObjectFactoryLogMessage());
-            objectFactory = new DefaultJavaObjectFactory();
+            LOG.warn(e.getMessage());
+            LOG.warn(getMultipleObjectFactoryLogMessage());
+            return new DefaultJavaObjectFactory();
         } catch (NoInstancesException e) {
-            objectFactory = new DefaultJavaObjectFactory();
+            return new DefaultJavaObjectFactory();
         } catch (ClassNotFoundException e) {
             throw new CucumberException("Couldn't instantiate custom ObjectFactory", e);
         }
-        return objectFactory;
+    }
+
+    private static ObjectFactory loadSingleObjectFactory(final Reflections reflections) {
+        Iterator<io.cucumber.core.backend.ObjectFactory> serviceLoaderObjectFactories = ServiceLoader.load(io.cucumber.core.backend.ObjectFactory.class).iterator();
+        if (serviceLoaderObjectFactories.hasNext()) {
+            final Collection<io.cucumber.core.backend.ObjectFactory> instances = new HashSet<>();
+            do {
+                instances.add(serviceLoaderObjectFactories.next());
+            } while (serviceLoaderObjectFactories.hasNext());
+            if (instances.size() > 1) {
+                throw new TooManyInstancesException(instances);
+            }
+            return new ObjectFactoryAdapter(instances.iterator().next());
+        } else {
+            LOG.warn("Use deprecated reflections to load ObjectFactory.");
+            final List<URI> packages = asList(URI.create("classpath:cucumber/runtime"));
+            return reflections.instantiateExactlyOneSubclass(ObjectFactory.class, packages, new Class[0], new Object[0], null);
+        }
+    }
+
+    private static ObjectFactory loadSelectedObjectFactory(final Reflections reflections, final ClassFinder classFinder, final String objectFactoryClassName) throws ClassNotFoundException {
+        final Iterator<? extends io.cucumber.core.backend.ObjectFactory> serviceLoaderObjectFactories = ServiceLoader.load(classFinder.<io.cucumber.core.backend.ObjectFactory>loadClass(objectFactoryClassName)).iterator();
+        if (serviceLoaderObjectFactories.hasNext()) {
+            return new ObjectFactoryAdapter(serviceLoaderObjectFactories.next());
+        } else {
+            LOG.warn("Use deprecated reflections to load requested ObjectFactory.");
+            return reflections.newInstance(new Class[0], new Object[0], classFinder.<ObjectFactory>loadClass(objectFactoryClassName));
+        }
     }
 
     private static String getMultipleObjectFactoryLogMessage() {
@@ -62,4 +90,35 @@ public class ObjectFactoryLoader {
         sb.append("In order to enjoy IoC features, please remove the unnecessary dependencies from your classpath.\n");
         return sb.toString();
     }
+
+    private static class ObjectFactoryAdapter implements ObjectFactory {
+
+        private final io.cucumber.core.backend.ObjectFactory delegate;
+
+        public ObjectFactoryAdapter(final io.cucumber.core.backend.ObjectFactory delegate) {
+            this.delegate = Objects.requireNonNull(delegate);
+        }
+
+        @Override
+        public void start() {
+            delegate.start();
+        }
+
+        @Override
+        public void stop() {
+            delegate.stop();
+        }
+
+        @Override
+        public boolean addClass(Class<?> arg0) {
+            return delegate.addClass(arg0);
+        }
+
+        @Override
+        public <T> T getInstance(Class<T> arg0) {
+            return delegate.getInstance(arg0);
+        }
+
+    }
+
 }
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>io.cucumber</groupId>
         <artifactId>cucumber-jvm</artifactId>
-        <version>4.3.2-SNAPSHOT</version>
+        <version>4.4.0-SNAPSHOT</version>
     </parent>
 
     <artifactId>cucumber-core</artifactId>
--- a/core/src/main/java/cucumber/runtime/TooManyInstancesException.java
+++ b/core/src/main/java/cucumber/runtime/TooManyInstancesException.java
@@ -4,7 +4,7 @@ import java.util.Collection;
 
 public class TooManyInstancesException extends CucumberException {
 
-    TooManyInstancesException(Collection instances) {
+    public TooManyInstancesException(Collection instances) {
         super(createMessage(instances));
     }
 
--- a/core/src/main/java/io/cucumber/core/backend/ObjectFactory.java
+++ b/core/src/main/java/io/cucumber/core/backend/ObjectFactory.java
@@ -1,0 +1,40 @@
+package io.cucumber.core.backend;
+
+/**
+ * Minimal facade for Dependency Injection containers
+ */
+public interface ObjectFactory {
+
+    /**
+     * Instantiate glue code <b>before</b> scenario execution. Called once per
+     * scenario.
+     */
+    void start();
+
+    /**
+     * Dispose glue code <b>after</b> scenario execution. Called once per
+     * scenario.
+     */
+    void stop();
+
+    /**
+     * Collects glue classes in the classpath. Called once on init.
+     *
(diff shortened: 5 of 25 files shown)


mpkorstanje:
Thanks for looking to contribute. Much appreciated.  However I have several severe objections to this and I don't think they can be resolved.

1. This implementation requires that `ObjectFactory` has a zero-arg constructor. This seems rather impractical as it renders backends unable to use the type registry and/or resource loader. 

2. This is not an exact backport of the current state of `develop-v5` which uses a `ServiceLoader<BackendProviderService>` instead of `ServiceLoader<ObjectFactory>`.

3. This aspect of `v5` is still in development and may change. Adding backwards compatibility now would be rather premature.

I also don't quite see the advantage of adding this functionality now. Perhaps you can clarify.

mpkorstanje:
> Easier migration to next major-version.

The `ObjectFactory` class will change from `cucumber.runtime.java` to `io.cucumber.core.backend`. Using a `ServiceLoader` now won't make migration easier.

toepi:
Okay let me explain why i think it should be a good idee to have this kind of "functionality". I hope you will not see it as a explanation of how you should do it.
**Background**
We have made the step from cucumber v2 to v4 and is cost us a big amount of time.
but I think you are right in say 'It is a major and not all old stuff must be work' - is perfectly legal. (And v3 has mostly introduce the same issues)
**Motivation**
Now it was nice to have the features in the old version also, so we can use a new (mostly better feature) with a old version and than change to a newer version and must not change all the stuff at once.
Another benefit more user can use this functionality and so test it - see continuous delivery best practices her, the "forbidden" branches and recommend "feature toggles".
**My Opinion**
For me you should introduce new functions in a minor version and create a major when you no longer can provide/maintain a given function. For this feature - support both classpathscanning and serviceloader for objectfactories and in v5 drop classpathscanning?

I know it make some thinks harder to implement - or if you look at my code and your code uglier - but for me as a user easier to use newer version and so for you less user using old versions.
This with the package is stupid for this pr, but I think we will find a solution?

toepi:
I've test this code with a 
- own `ObjectFactory` which implements `cucumber.api.java.ObjectFactory` and does not use ServiceLoader -> use my `ObjectFactory`
- same `ObjectFactory` and use it with `ServiceLoader` -> use my `ObjectFactory`
- same `ObjectFactory` with another `ObjectFactory` with `ServiceLoader` -> use `DefaultFactory` and print Warning
- own `ObjectFactory` and cucumber-weld:4.3.0 into classpath -> use `DefaultFactory` and print Warning
- change interface of my own `ObjectFactory` to new `io.cucmber.core.backend` -> use my `ObjectFactory`

all this test get the expected result's

mpkorstanje:
Thanks for explaining. It makes sense but I am afraid the implementation isnt quite right. I'll try to explain but I am going to be short on time this week.

java/src/main/java/cucumber/api/java/ObjectFactory.java
```diff
@@ -2,34 +2,10 @@
 
 /**
  * Minimal facade for Dependency Injection containers
+ *
+ * @deprecated use @code io.cucumber.core.backend.ObjectFactory instead
  */
-public interface ObjectFactory {
+@Deprecated
+public interface ObjectFactory extends io.cucumber.core.backend.ObjectFactory {
```
mpkorstanje: Given that you've added an `ObjectFactoryAdapter` is it necessary to extend this interface?

java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
```diff
@@ -32,18 +36,16 @@ private ObjectFactoryLoader() {
...
         try {
-            Reflections reflections = new Reflections(classFinder);
-
-            if (objectFactoryClassName != null) {
-                Class<ObjectFactory> objectFactoryClass = (Class<ObjectFactory>) classFinder.loadClass(objectFactoryClassName);
-                objectFactory = reflections.newInstance(new Class[0], new Object[0], objectFactoryClass);
+            final Reflections reflections = new Reflections(classFinder);
+            final ServiceLoader<io.cucumber.core.backend.ObjectFactory> loader = ServiceLoader.load(io.cucumber.core.backend.ObjectFactory.class);
+            if (objectFactoryClassName == null) {
+                objectFactory = new ObjectFactoryAdapter(loadSingleObjectFactory(reflections, loader));
```
mpkorstanje: Adapters should only be applied as needed. Can you push this down to the point it's needed? Same for the one below.

java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
```diff
@@ -53,6 +55,46 @@ public static ObjectFactory loadObjectFactory(ClassFinder classFinder, String ob
...
+            result = objectFactoriesByServiceLoader.get(objectFactoryClassName);
+        } else {
+            final Class<io.cucumber.core.backend.ObjectFactory> objectFactoryClass = (Class<io.cucumber.core.backend.ObjectFactory>) classFinder.loadClass(objectFactoryClassName);
+            result = reflections.newInstance(new Class[0], new Object[0], objectFactoryClass);
+        }
+        return result;
+    }
+
+    private static void loadObjectFactories(final Iterable<? extends io.cucumber.core.backend.ObjectFactory> objectFactories, final Map<String, io.cucumber.core.backend.ObjectFactory> knowObjectFactories) {
+        for (io.cucumber.core.backend.ObjectFactory objectFactory : objectFactories) {
```
mpkorstanje: This will actually instantiate all object factories. This is never desirable.

java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
```diff
@@ -53,6 +55,46 @@ public static ObjectFactory loadObjectFactory(ClassFinder classFinder, String ob
...
+        loadObjectFactories(reflections.instantiateSubclasses(io.cucumber.core.backend.ObjectFactory.class, packages, new Class[0], new Object[0]), knowObjectFactories);
+        loadObjectFactories(loader, knowObjectFactories);
+        return knowObjectFactories;
+    }
+
+    private static io.cucumber.core.backend.ObjectFactory loadSelectedObjectFactory(final Reflections reflections, final ServiceLoader<io.cucumber.core.backend.ObjectFactory> loader, final ClassFinder classFinder, final String objectFactoryClassName) throws ClassNotFoundException {
+        final Map<String, io.cucumber.core.backend.ObjectFactory> objectFactoriesByServiceLoader = new HashMap<>();
+        loadObjectFactories(loader, objectFactoriesByServiceLoader);
+        final io.cucumber.core.backend.ObjectFactory result;
+        if (objectFactoriesByServiceLoader.containsKey(objectFactoryClassName)) {
```
mpkorstanje: If you know the class name you can use https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html#load(java.lang.Class) instead. This avoids instantiation of all the services.

java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
```diff
@@ -53,6 +55,46 @@ public static ObjectFactory loadObjectFactory(ClassFinder classFinder, String ob
         return objectFactory;
     }
 
+    private static io.cucumber.core.backend.ObjectFactory loadSingleObjectFactory(final Reflections reflections, final ServiceLoader<io.cucumber.core.backend.ObjectFactory> loader) {
```
mpkorstanje: Rather then trying both the service loader and class path scanning this method should first try the service loader and then try class path scanning. Doing both and then checking the result makes more complicated then it needs to be.

java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
```diff
@@ -53,6 +55,46 @@ public static ObjectFactory loadObjectFactory(ClassFinder classFinder, String ob
...
+        final Map<String, io.cucumber.core.backend.ObjectFactory> objectFactories = loadObjectFactories(reflections, loader);
+        final io.cucumber.core.backend.ObjectFactory result;
+        if (objectFactories.isEmpty()) {
+            throw new NoInstancesException(io.cucumber.core.backend.ObjectFactory.class);
+        } else if (objectFactories.size() > 1) {
+            throw new TooManyInstancesException(objectFactories.values());
+        } else {
+            result = objectFactories.values().iterator().next();
+        }
+        return result;
```
mpkorstanje: It's okay to use an early return statement once an object factory has been found.

java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
```diff
@@ -53,6 +55,46 @@ public static ObjectFactory loadObjectFactory(ClassFinder classFinder, String ob
...
+        final Map<String, io.cucumber.core.backend.ObjectFactory> objectFactoriesByServiceLoader = new HashMap<>();
+        loadObjectFactories(loader, objectFactoriesByServiceLoader);
+        final io.cucumber.core.backend.ObjectFactory result;
+        if (objectFactoriesByServiceLoader.containsKey(objectFactoryClassName)) {
+            result = objectFactoriesByServiceLoader.get(objectFactoryClassName);
+        } else {
+            final Class<io.cucumber.core.backend.ObjectFactory> objectFactoryClass = (Class<io.cucumber.core.backend.ObjectFactory>) classFinder.loadClass(objectFactoryClassName);
+            result = reflections.newInstance(new Class[0], new Object[0], objectFactoryClass);
+        }
+        return result;
```
mpkorstanje: It's okay to use an early return statement once an object factory has been found.

java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
```diff
@@ -53,6 +55,46 @@ public static ObjectFactory loadObjectFactory(ClassFinder classFinder, String ob
         return objectFactory;
```
mpkorstanje: It's okay to use an early return statement. It's just outside of the scope of your changes but will make this section of the code easier to read.

java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
mpkorstanje: Ah. I see you copied this from the original. I should fix it there too.

java/src/main/java/cucumber/api/java/ObjectFactory.java
toepi: Not sure, but if not we have in both interfaces the same methods? And we need this Adapter (which I've only created not to change JavaBackend in a incompatible way?)
Let me know which solution you prefer:
1. Old Interface empty (current solution)
2. Old and new Interface are identical and only in different packages

java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
toepi: Yeah but why is this a issue? If you have more than one at your path cucumber ignore them an than run all scenarios with the default one with take more time than create unnecessary instances? And if you look at the code in reflector it does the same?

java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
toepi: sry? but I do not know the class? We know the interface?
This is only a workaround to check if service loader and old reflection found the same Classes or found different. Can happen if you migrate a ObjectFactory to ServiceLoader and than put a old one into class path. Okay I can change the impl. so we use ServiceLoader only if Reflection not found one? Do you think this is the better way?

java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
toepi: Was my first idea, but than I've not found at which place?
1. not necessary for:
1.1 old Interfaces-impl. loaded by ServiceLoader
1.2 old interfaces-impl. loaded by Reflection
2. necessary for:
2.1. only new interface-impl. loaded by ServiceLoader
2.2. only new interface-impl. loaded by Reflection, and yes with current impl. is it possible
This means every finder can return a Factory which need this Adapter, and if you look at the impl. the overhead of it is minimal.

java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
toepi: But than we do not detect if more than one ObjectFactory is available - currently it detect if relefection and service loader found different implementations.

toepi: But can change it, but than I think in the other way? First Reflection than ServiceLoader?

java/src/main/java/cucumber/api/java/ObjectFactory.java
mpkorstanje: The duplicate code is not a problem. They're different interfaces. So option #2 would be fine.

java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
mpkorstanje: It's important to keep related bits of functionality together. Makes it easier to understand the code and once ObjectFactory no longer extends io.cucumber.core.backend.ObjectFactory it won't be possible either anyway.

java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
mpkorstanje: It avoids potential side effects. Means there is one less thing to take into account when debugging. Just because it was done in the past, doesn't mean it's good idea to continue it.

java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
mpkorstanje: We do know the class name. It's `objectFactoryClassName`. You can use that to get the class.

We should give preference to the new system over the old one. So if the `ServiceLoader` does not give us any results we fall back to reflection.

mpkorstanje: Reason being that we deprecated the old solution. So it should not be used when not needed.

java/src/main/java/cucumber/runtime/java/ObjectFactoryLoader.java
mpkorstanje: First ServiceLoader then Reflection. We're deprecating the reflection after all.

toepi:
Now we have two distinct ways:
Try ServiceLoader first and if is was success, use it. If not try Reflections. Unsure if we real must check if more than one exists, why not use the first one (more convince?)

toepi:
Question: should i update version to 4.4.0-SNAPSHOT so that travis can build this pr?

mpkorstanje:
> Unsure if we real must check if more than one exists, why not use the first one (more convince?)

Ideally the fail fast principle is applied and we throw an exception when the system is misconfigured rather then continue. We don't do this. But it would be good to do so.

The fail fast principle is useful because it prevents the system from acting unpredictable. Instead it just stops.

Which factory is first would depend on the order the classes are loaded. This is not defined properly. So using the first makes the system unpredictable when misconfigured.

So ideally we do this:

```
ObjectFactory o = load with service loader. // throws when misconfigured

If o != null return o;

ObjectFactory o = load with scanning // logs error when multiple factories are present
If o != null return o;

return default object factory
```

This keeps the code structure really obvious.

mpkorstanje:
> Question: should i update version to 4.4.0-SNAPSHOT so that travis can build this pr?

Please do. There is a maven command for it. Don't do it by hand! ;)

toepi:
before i resolve the conflicts, any other changes needed for this pr?

mpkorstanje:
It's the version numbers. You can rebase on master it should be fine.

mpkorstanje:
No other changes needed.

aslakhellesoy:
Hi @toepi,

Aslak Hellesøy
Creator of Cucumber

aslakhellesoy:
Hi @toepi,

Aslak Hellesøy
Creator of Cucumber

PR #1616, opened 2019-04-28
[DeltaSpike] Add ObjectFactory for Apache DeltaSpike

toepi:
## Summary

objectfactory for cdi with deltaspike

## Details

Deltaspike abstract underlining cdi-container. so it is possible to change cdi-impl. only by replace the dependencies.

## Motivation and Context
No longer to modules for cdi, only one with can replace the old one (weld and openejb) and in the same time support for another cdi-impl (OpenWebBeans).

## How Has This Been Tested?

Add Testcases and profiles to change cdi-impl.

## Types of changes

## Checklist:

### DIFF
--- a/deltaspike/README.md
+++ b/deltaspike/README.md
@@ -16,6 +16,36 @@ Enable cdi support for your steps by adding a (empty) beans.xml into your classp
 </beans>
 ```
 
+To use DependencyInjection add `@Inject` to any field which should be managed by CDI, for more information see [JSR330](https://www.jcp.org/en/jsr/detail?id=330).
+
+```java
+public class BellyStepdefs {
+
+    @Inject
+    private Belly belly;
+
+    //normal step code ...
+```
+
+This ObjectFactory doesn't start or stop any [Scopes](https://docs.oracle.com/javaee/6/tutorial/doc/gjbbk.html), so all beans live inside the default scope (Dependent). Now cucumber requested a instance of your stepdefinitions for every step, which means cdi create a new instance for every step and for all injected fields. This behaviour makes it impossible to share a state inside a szenario.
+
+To bybass this, you must annotate your class(es) with `@javax.inject.Singleton`:
+1. on stepdefintions: now the ojectfactory will creates only one instance include injected fields per scenario and both injected fields and stepdefinitions can be used to share state inside a scenario.
+2. on any other class: now the objectfactory will create a new instance of your stepdefinitions per step and stepdefinitions can not be used to share state inside a scenario, only the annotated classes can be used to share state inside a scenario
+
+you can also combine both approaches.
+
+```java
+@Singleton
+public class BellyStepdefs {
+
+    @Inject
+    private Belly belly;
+
+    //normal step code ...
+```
+It is not possible to use any other scope than Dependent this means alsoi it is not possible to share a state over two or more scenarios, every scenario start with a clean environment.
+
 To enable this objectfactory add the folling dependency to your classpath:
 ```xml
 <dependency>
@@ -97,4 +127,3 @@ or to use it with OpenWebBeans:
 ```
 
 Some containers need that you provide a CDI-API in a given version, but if you develop CDI and use one of the above containers it should already on your path.
-
--- a/deltaspike/README.md
+++ b/deltaspike/README.md
@@ -1,0 +1,100 @@
+Cucumber DeltaSpike
+===================
+
+This module relies on [DeltaSpike Container Control](https://deltaspike.apache.org/documentation/container-control.html) to start/stop supported CDI container.
+
+## Setup
+Enable cdi support for your steps by adding a (empty) beans.xml into your classpath (src/main/resource/META-INF for normal classes or src/test/resources/META-INF for test classes):
+
+```xml
+<beans xmlns="http://java.sun.com/xml/ns/javaee"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+      http://java.sun.com/xml/ns/javaee
+      http://java.sun.com/xml/ns/javaee/beans_1_0.xsd">
+
+</beans>
+```
+
+To enable this objectfactory add the folling dependency to your classpath:
+```xml
+<dependency>
+    <groupId>io.cucumber</groupId>
+    <artifactId>cucumber-deltaspike</artifactId>
+    <version>${cucumber.version}</version>
+    <scope>test</scope>
+</dependency>
+```
+
+and one of the supported cdi-containers.
+
+to use it with Weld:
+
+```xml
+<dependency>
+    <groupId>org.apache.deltaspike.cdictrl</groupId>
+    <artifactId>deltaspike-cdictrl-weld</artifactId>
+    <version>${deltaspike.version}</version>
+    <scope>test</scope>
+</dependency>
+<dependency>
+    <groupId>org.jboss.weld.se</groupId>
+    <artifactId>weld-se-core</artifactId>
+    <version>${weld-se.version}</version>
+    <scope>test</scope>
+</dependency>
+```
+
+or to use it with OpenEJB:
+
+```xml
+<dependency>
+    <groupId>org.apache.deltaspike.cdictrl</groupId>
+    <artifactId>deltaspike-cdictrl-openejb</artifactId>
+    <version>${deltaspike.version}</version>
+    <scope>test</scope>
+</dependency>
+<dependency>
+    <groupId>org.apache.tomee</groupId>
+    <artifactId>openejb-core</artifactId>
+    <version>${openejb-core.version}</version>
+    <scope>test</scope>
+</dependency>
+<dependency>
+    <groupId>javax.xml.bind</groupId>
+    <artifactId>jaxb-api</artifactId>
+    <version>${jaxb-api.version}</version>
+    <scope>test</scope>
+</dependency>
+<dependency>
+    <groupId>org.glassfish.jaxb</groupId>
+    <artifactId>jaxb-runtime</artifactId>
+    <version>${jaxb-api.version}</version>
+    <scope>test</scope>
+</dependency>
+```
+
+or to use it with OpenWebBeans:
+```xml
+<dependency>
+    <groupId>org.apache.deltaspike.cdictrl</groupId>
+    <artifactId>deltaspike-cdictrl-owb</artifactId>
+    <version>${deltaspike.version}</version>
+    <scope>test</scope>
+</dependency>
+<dependency>
+    <groupId>org.apache.openwebbeans</groupId>
+    <artifactId>openwebbeans-impl</artifactId>
+    <version>${owb.version}</version>
+    <scope>test</scope>
+</dependency>
+<dependency>
+    <groupId>javax.annotation</groupId>
+    <artifactId>jsr250-api</artifactId>
+    <version>1.0</version>
+    <scope>test</scope>
+</dependency>
+```
+
+Some containers need that you provide a CDI-API in a given version, but if you develop CDI and use one of the above containers it should already on your path.
+
--- a/deltaspike/pom.xml
+++ b/deltaspike/pom.xml
@@ -17,11 +17,6 @@
     <properties>
         <project.Automatic-Module-Name>io.cucumber.deltaspike</project.Automatic-Module-Name>
         <deltaspike.version>1.9.0</deltaspike.version>
-        <cdi-api.version>1.0-SP4</cdi-api.version>
-        <jaxb-api.version>2.3.1</jaxb-api.version>
-        <weld-se.version>1.1.34.Final</weld-se.version>
-        <openejb-core.version>8.0.0-M1</openejb-core.version>
-        <owb.version>2.0.10</owb.version>
     </properties>
 
     <dependencies>
@@ -37,7 +32,7 @@
         <dependency>
             <groupId>javax.enterprise</groupId>
             <artifactId>cdi-api</artifactId>
-            <version>${cdi-api.version}</version>
+            <version>1.0-SP4</version>
             <scope>provided</scope>
         </dependency>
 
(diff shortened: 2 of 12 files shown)


deltaspike/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+            <artifactId>cdi-api</artifactId>
+            <version>${cdi-api.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>cucumber-junit</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
```
mpkorstanje: You can omit the version here and please avoid the use of properties when not needed.

```
        <dependency>
            <scope>test</scope>
        </dependency>
```

deltaspike/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <profiles>
+        <profile>
+            <id>deltaspike-weld</id>
```
mpkorstanje: Why the different profiles? It would appear they all go unused.

toepi: to show that you can use weld, openejb and openwebbeans with this module? you're right only weld will be used at default. But I'm unsure how to test the other cdi-containers here or to show that the will work? At a later stage we can replace the profiles with to "old" modules?

deltaspike/src/main/java/cucumber/runtime/java/deltaspike/DeltaSpikeObjectFactory.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Override
+    public boolean addClass(final Class<?> clazz) {
+        return true;
+    }
+
+    @Override
+    public <T> T getInstance(final Class<T> type) {
+        return CDI.current().select(type).get();
```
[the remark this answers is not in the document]
toepi: this is the "default" way to get a type for a CDI-Container? The `container` is a deltaspike interface to start and stop containers in java-se it does not contains a easy to use method to get the requested object.

deltaspike/pom.xml
toepi: > Could you add a `README.md` to explain how this module should be used? It doesn't appear to be trivial.

It can be used in the same why you use weld or obenejb module (if i do it right, it should be an drop-in replacement).
Any preference where i should place the readme, in the basedir of the module? And any tips what should i explain? I will try it, but you see my english is not the best.

mpkorstanje: Yes. Base dir of the module should be fine. Don't worry about your English. Others may be able to improve on it.

I general the information included should be sufficient to get some one who's checked out the https://github.com/cucumber/cucumber-java-skeleton to get started.

So listing the required dependencies, the optional dependencies, and including an example of a step definition class that uses injection should be enough. Feel free to add anything else that is note worthy too.

deltaspike/src/main/java/cucumber/runtime/java/deltaspike/DeltaSpikeObjectFactory.java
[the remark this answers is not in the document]
mpkorstanje: I wasn't aware. Cheers.

deltaspike/pom.xml
mpkorstanje: It's nice as a demonstration. However this is an open source project which means that any sort of manual test will soon fall by the wayside. Might be better to remove it up all together.

toepi:
Have found this old pr #971 this should be deprecated with this one?

deltaspike/pom.xml
toepi: okay, what do you think about replace profiles with examples for the different containers? Than I can referent the doc also to this?

mpkorstanje: Just the default profile should be fine I think. Delta Spike is a facade for the different libraries. You can put the examples in the docs.

deltaspike/src/main/java/cucumber/runtime/java/deltaspike/DeltaSpikeObjectFactory.java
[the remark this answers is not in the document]
rmannibucau: CDI comes from cdi 2 where deltaspike usage can be replace by SeContainer. Did you intend to use BeanManager to do the lookup or deltaspike BeanProvider directly?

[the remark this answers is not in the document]
toepi: CDI comes from CDI 1.1? SeContainer with 2.0. Why should I use BeanManager at this place?
It should be possible to extend addClass to register given type to the container, but think CDI default discovery should be good enough her?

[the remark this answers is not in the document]
rmannibucau: Well there are two things

1. As you mentionned cdi comes from spec v1.1 (not 2 as i mentionned) but point is deltaspike api is really portable for cdi >= 1.0 containers so you cant use it together
2. Yes scanning is a good default but even in this case you dont need cdi class since deltaspike has a replacement for it

In other words it looks a hybrid cdi 2 (secontainer+cdi) / deltaspike (cdictrl/beanprovider) impl to me so can be worth cleaning up IMHO.

Does it make sense presented this way?

Indeed i prefer the option with v2 but i agree a v1.0 still makes sense today.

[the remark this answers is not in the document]
toepi: 2 - i do not understand.

to solve the hybrid you prefer beanprovider over CDI?

mpkorstanje:
I'm happy to take both PR's. While CDI containers are used only by a fraction of the users they're relatively low maintenance and have a relatively high number of contributors.

I do however have a rather limited knowledge of either so I'd like both implementations to be correct, in line with the frameworks expected use (so as to limit future questions/issues) and focused on the basic use case. For advanced use cases I expect that people re-implement the factory. At least until we come up with something better.

Also if you can both review and approve each others PR's then I can have some confidence that both implementations are good.

deltaspike/src/main/java/cucumber/runtime/java/deltaspike/DeltaSpikeObjectFactory.java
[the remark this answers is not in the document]
rmannibucau: Yes please

Point 2 was just that CDI is the standard of BeanProvider but nit available for cdictrl based on cdi 1.0 impls.

Side note you can need to handle creational contexts to release them to not leak.

mpkorstanje:
I've pushed some fixes so it will work wit the latest 5.0.0-SNAPSHOT. Would you mind adding a short guide on how to use this module in the read me?

toepi:
sry complete forgotten this pr, will try to add requested guide next week. for 5.0.0 I think this objectfactory should replace weld and openejb do you think I should update this pr to do this also?

mpkorstanje:
Cheers.

Generally speaking its good practice not to remove something withouth a deprecation period. It also doesn't hurt to keep weld around for a bit. So leave it alone for now.

rmannibucau:
Side note: this does not fully superseed openejb due to DD impl but an EJBContainer based impl could happily do it and enable to support tomee at the same time - can also be done behind DS API if you want to do the pr ;).

mpkorstanje:
That'd be quite useful. If you're interested in doing this please do go ahead.

toepi:
after reading cucumber/docs.cucumber.io#369, do you think it will be better to replace `container.boot()` and `container.shutdown()` with somthink to start and stop a given scope (e.g. requestedscope)?

rmannibucau:
Makes sense if configurable cause any setup can break tests from my experience. Often request is started by default but you can need session or need to not have request commonly too. Depends how you write it and which container you use sometimes.

Can make more sense to decorate step instance with an interceptor starting/stopping scope at request (@Scopes on the step ?).

mpkorstanje:
Cucumber currently puts step definitions in the container. Because step definitions generally speaking get dirty the sensible default would be to restart the container completely. 

If this operation is expensive you may also consider removing only the step definition classes (everything provided by `addClass`) and keeping the container alive. However step definitions should be recreated for each scenario.

rmannibucau:
Operation is cheap but it is a state app code can rely on, this is why it must be configurable IMO. That said, can be a "given" statement with implicit close.

mpkorstanje:
I'm not following. Each scenario should run in an clean context. The following feature should always pass:

```
Feature: Don't share state between scenarios
  Scenario: I dirty the context
      Given I set FOO from BAR to BAZ

  Scenario: The dirty context should gone
       Then FOO should be set to BAR again
```

rmannibucau:
Depending how you run or impl each step context can be different. Typically a client step will have a request context per http request whereas a server step can share it state within the scenario. This is why it must be configurable.
Also keep in mind some lib test if the context is active or not - jta for instance - so starting it by default can make the test silently broken.

mpkorstanje:
> Typically a client step will have a request context per http request whereas a server step can share it state within the scenario.

What are sever steps and client steps? Could you write some scenarios to show the difference.

>  Also keep in mind some lib test if the context is active or not - jta for instance - so starting it by default can make the test silently broken.

I don't see how this is relevant. Could you write some scenarios to show when it matters?

rmannibucau:
1. Selenium vs service testing (famous belly one ;))
2. If your logic activates stuff based on the scope activation or not then the app behavuor differs. Typically if you start request scope when it shouldnt, transactional scope lifecycle will differ. It is true for several libs so each scenario should be able to define in which context it runs.
In tomee embedded we prestarted request scope and it broke some users tests so we end up making it fully configurable.
I dont have a computer handy but can try to shape something next week.
We can also let it like that and enhance it at need - users can have a single usage for now :crossing_fingers:.

mpkorstanje:
Let's start with no scope activated (basically selenium). As defaults go that is pretty sensible. 

Then we can work out how to activate other scopes, but given that object factory needs to start the context before the scenario can start I don't think it will be possible. Unless you can change the scope once the factory has started. In that case you tag a scenario with the required scope and used tagged hooks to setup that scope.

toepi:
rebased to master and add a readme, but somebody with better English should take a look.

deltaspike/README.md
```diff
@@ -0,0 +1,100 @@
...
+<dependency>
+    <groupId>javax.annotation</groupId>
+    <artifactId>jsr250-api</artifactId>
+    <version>1.0</version>
+    <scope>test</scope>
+</dependency>
+```
+
+Some containers need that you provide a CDI-API in a given version, but if you develop CDI and use one of the above containers it should already on your path.
+
```
mpkorstanje: Pretty good. I didn't spot any problems with your English.

After reading this I am left with two questions:

1. I know how to set this up, but how do I do the dependency injection thing? Which annotations do I use, where do I use them. This doesn't need to be a full treatise on every available annotation. Just the most obvious `@Inject` one.

2. What is the scope of the injected beans? When I start a second scenario will my steps be able to access information from the previous scenario (they shouldn't), will the application/container be in the same state the previous scenario left it (it shouldn't, but tolerable).

deltaspike/src/main/java/io/cucumber/deltaspike/DeltaSpikeObjectFactory.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.apache.deltaspike.cdise.api.CdiContainer;
+import org.apache.deltaspike.cdise.api.CdiContainerLoader;
+import io.cucumber.core.backend.ObjectFactory;
+import javax.enterprise.context.spi.CreationalContext;
+import javax.enterprise.inject.spi.Bean;
+import javax.enterprise.inject.spi.BeanManager;
+import java.util.Set;
+
+public class DeltaSpikeObjectFactory implements ObjectFactory {
```
mpkorstanje: Note to self: Make final, add @API annotation

deltaspike/README.md
mpkorstanje: Might be good to reference JSR330 by name.

deltaspike/src/main/java/io/cucumber/deltaspike/DeltaSpikeObjectFactory.java
toepi: can you me explain why you need this annotation here? or why you set it to stable and not to internal (other objectfactories)... for me this is not a api is it a implementation and nobody should rely on it?
btw. have done it, push will follow.

deltaspike/README.md
toepi: nice hint, work is in progress :)

deltaspike/src/main/java/io/cucumber/deltaspike/DeltaSpikeObjectFactory.java
mpkorstanje: There are a few use cases where decorating the object factory makes sense. Aslak wrote about one here:

https://cucumber.io/blog/polymorphic-step-definitions/

Unfortunately he used inheritance rather then composition. I want to allow for composition and forbid inheritance. 

The final prevents inheritance. Marking the class as part of the API renders it safe for composition.

mpkorstanje: Ideally we'd have a different system for this.

Something that would be composeable with object factories with out being an object factory. But that is a design challenge that still needs to be solved.

A use case for this would be `PageFactory`. 


https://www.toolsqa.com/selenium-webdriver/page-object-pattern-model-page-factory/

toepi: Thank for the explanation, with cdi some extensions should be possible with 'javax.enterprise.inject.spi.Extension' (use it on a closed project to integrate arquillian-drone)

mpkorstanje: Aah. That also works.

toepi:
Currently I do not see anything missing, please merge it and hope we find some user which get some input what we have forgotten.

mpkorstanje:
All merged. Thanks

PR #1620, opened 2019-05-04
[Core] Replace TimeService with Java Time API 

gazler22:
## Summary
Migrating the current code to use Java 8 Time API fields instead of the currently used primitive fields.

## Details

The long fields `time`,  `timeStampMillis` and `duration` have been replaced by `Instant` and `Duration` respectively.

These values can now be used instead of older primitives to allow for better representation of time and duration in cucumber without having to worry about precision or granularity.

## Motivation and Context

To allow for more consistent time measurements and reduce custom code.

## How Has This Been Tested?

All the JUnits have been updated accordingly to use `Instant` and `Duration`. 

## Types of changes

## Checklist:


core/src/main/java/io/cucumber/core/plugin/JUnitFormatter.java
```diff
@@ -264,7 +264,7 @@ void handleEmptyTestCase(Document doc, Element tc, Result result) {
         private String calculateTotalDurationString(Result result) {
             DecimalFormat numberFormat = (DecimalFormat) NumberFormat.getNumberInstance(Locale.US);
             numberFormat.applyPattern("0.######");
-            return numberFormat.format(((double) result.getDuration()) / 1000000000);
+            return numberFormat.format(((double) result.getTimeDuration().toNanos()));
```
mpkorstanje: I think the division is stil needed.

core/src/main/java/io/cucumber/core/api/event/EmbedEvent.java
```diff
@@ -9,10 +11,17 @@ public EmbedEvent(Long timeStamp, TestCase testCase, byte[] data, String mimeTyp
         this(timeStamp, 0, testCase, data, mimeType);
     }
 
+    @Deprecated
     public EmbedEvent(Long timeStamp, long timeStampMillis, TestCase testCase, byte[] data, String mimeType) {
```
mpkorstanje: No need to deprecate this. It can be removed all together.

mpkorstanje: Same for all the other event constructors.

core/src/main/java/io/cucumber/core/api/event/TimeStampedEvent.java
```diff
@@ -24,7 +37,17 @@ public Long getTimeStamp() {
...
+    @Deprecated
     public long getTimeStampMillis() {
         return timeStampMillis;
     }
+    
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Instant getTimeInstant() {
```
mpkorstanje: Are there other types of instants? If not, just instant will do.

core/src/main/java/io/cucumber/core/plugin/JSONFormatter.java
```diff
@@ -381,16 +382,16 @@ private void addEmbeddingToHookMap(byte[] data, String mimeType) {
...
         return resultMap;
     }
     
-    private String getDateTimeFromTimeStamp(long timeStampMillis) {
+    private String getDateTimeFromTimeStamp(Instant instant) {
         SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX");
         sdf.setTimeZone(TimeZone.getTimeZone("UTC"));
         
-        return sdf.format(new Date(timeStampMillis));
+        return sdf.format(new Date(instant.toEpochMilli()));
```
mpkorstanje: It would be cleaner to directly format the instant. Java 8 has a new formatter for temporals.

core/src/main/java/io/cucumber/core/runner/TimeService.java
```diff
@@ -1,18 +1,16 @@
 package io.cucumber.core.runner;
 
+import java.time.Instant;
+
 public interface TimeService {
```
mpkorstanje: Can you replace this with `Clock`.

gazler22: You mean use `Clock` instead of `Instant` everywhere?

Because that would mean the `TimeService` would return a `clock()` instance instead of `instant()` and same with the `EventBus` and all others (`getClock()` instead of `getInstant()`).

core/src/main/java/io/cucumber/core/api/event/EmbedEvent.java
gazler22: You mean remove all deprecations from `TestCaseStarted`, `TestCaseFinished`, `TimeStampedEvent`, etc  also?

Since they all extend `Event`

gazler22:
> 
> 
> Looks good to me. The JUnitFormatterTest contains changes in the time formatting. That change is not compatible with the previous behavior.is it correct?

Correct. I updated the time formatting to handle  for values received from Duration and Instant instead

core/src/main/java/io/cucumber/core/runner/TimeService.java
mpkorstanje: No. I mean replace `TimeService` with `Clock`. They serve the same purpose.

core/src/main/java/io/cucumber/core/api/event/EmbedEvent.java
mpkorstanje: You can remove all those constructors. This is v5 so no need to deprecate them first.

mpkorstanje:
> Correct. I updated the time formatting to handle for values received from Duration and Instant instead

I think that's because you no longer divide the nanos?

gazler22:
> I think that's because you no longer divide the nanos?

Yes you're right. I had updated the formatter to use nanos directly so I updated the tests as well. But I am updating the format back to millis now and adding the division.

core/src/main/java/io/cucumber/core/runner/TimeService.java
gazler22: I removed `TimeService` and updated all the references to use `Clock.fixed()`or `Clock.systemUTC()` instead. Theres a new `ClockStub` now too. Can you take a look if its fine now.

junit/src/test/java/io/cucumber/junit/api/FeatureRunnerTest.java
```diff
@@ -159,11 +160,21 @@ private FeatureRunner createFeatureRunner(CucumberFeature cucumberFeature, JUnit
...
+
+			@Override
+			public ZoneId getZone() {
+				return null;
+			}
+
+			@Override
+			public Clock withZone(ZoneId zone) {
+				return null;
+			}
```
mpkorstanje: The indentation looks of here.

core/src/main/java/io/cucumber/core/plugin/JUnitFormatter.java
```diff
@@ -264,7 +264,7 @@ void handleEmptyTestCase(Document doc, Element tc, Result result) {
         private String calculateTotalDurationString(Result result) {
             DecimalFormat numberFormat = (DecimalFormat) NumberFormat.getNumberInstance(Locale.US);
             numberFormat.applyPattern("0.######");
-            return numberFormat.format(((double) result.getTimeDuration().toNanos()));
+            return numberFormat.format(((double) result.getDuration().toNanos() / 1000));
```
mpkorstanje: I can't work out why this division is here.

mpkorstanje:
Looks pretty good so far! I'll need a bit more time to check the details.

core/src/main/java/io/cucumber/core/plugin/JUnitFormatter.java
mpkorstanje: Owh wait. I'm looking at a single commit.

mpkorstanje: So the old code was

```
return numberFormat.format(((double) result.getDuration()) / 1000000000);
```

Now it does:

```
return numberFormat.format(((double) result.getDuration().toNanos() / 1000));
```

What are you intending to do here?

core/src/test/java/io/cucumber/core/plugin/JUnitFormatterTest.java
```diff
@@ -76,7 +78,7 @@ public void should_format_passed_scenario() throws Throwable {
         stepsToResult.put("first step", result("passed"));
         stepsToResult.put("second step", result("passed"));
         stepsToResult.put("third step", result("passed"));
-        stepDurationMillis = 1L;
+        stepDuration = Duration.ofNanos(1L);
```
mpkorstanje: Why change from milis to nanos here?

gazler22: Nothing in particular but since I have been using nanos everywhere with the new `Duration` object for the sake of precision I updated this to that as well, along with the formatter too.

You want me to revert it to millis?

core/src/main/java/io/cucumber/core/plugin/JUnitFormatter.java
gazler22: You asked to revert the test case back to the old millis values and keep the division correct? 

I reverted the test case and divided the nanos by `1000` to keep the values returned in millis like they were. If I just use `.toMillis()` it will return just round it to `0.0` instead of the correct `0.003`.

mpkorstanje: I think the divisor should be 1000000000, and if you use millis in the test the expected result should stay the same.

In essence the reported time is in seconds.

junit/src/test/java/io/cucumber/junit/api/FeatureRunnerTest.java
gazler22: This is weird, it shows correctly in my eclipse

gazler22: Its correct here too if I go to edit it directly in github, but the diff shown here makes it look off.

mpkorstanje: Tabs vs spaces?

gazler22: Yea maybe that's it. I ll update it.

core/src/main/java/io/cucumber/core/plugin/JUnitFormatter.java
gazler22: Ah ok. So I updated the TC to use `stepDuration = Duration.ofMillis(1L);` and the above code to do `((double) result.getDuration().toNanos() / 1000000000)` instead.

It keeps the expected result the same.

junit/src/test/java/io/cucumber/junit/api/FeatureRunnerTest.java
gazler22: Yea i think that did it. Thanks

core/src/main/java/io/cucumber/core/plugin/UsageFormatter.java
```diff
@@ -85,45 +87,37 @@ void finishReport() {
...
+            rawDurations.add(stepDuration.duration);
         }
         return rawDurations;
     }
 
     private Gson gson() {
-        return new GsonBuilder().setPrettyPrinting().create();
+        return new GsonBuilder().registerTypeAdapter(Duration.class,
+                (JsonSerializer<Duration>) (duration, returnVal, jsonSerializationContext) -> {
+                    return new Gson().toJsonTree(duration.getSeconds());
```
mpkorstanje: This is pretty neat.

mpkorstanje:
Looks all good. Cheers!

PR #1623, opened 2019-05-06
[Core] Use meaningful thread names

kutzi:
## Summary

When running with --threads, give the the pool threads some meaningful names

## Details

Before it was using the standard threadfactory, which was naming threads like 'pool-x-thread-y'
Now it is 'cucumber-runner-thread-x'

## Motivation and Context

It's better for logging to see which threads belong to cucumber and which are unrelated background threads

### DIFF
--- a/core/src/main/java/cucumber/runtime/Runtime.java
+++ b/core/src/main/java/cucumber/runtime/Runtime.java
@@ -214,26 +214,19 @@ public class Runtime {
 
     private static final class CucumberThreadFactory implements ThreadFactory {
 
+        private static final AtomicInteger poolNumber = new AtomicInteger(1);
         private final AtomicInteger threadNumber = new AtomicInteger(1);
-        private final ThreadGroup group;
+        private final String namePrefix;
 
         CucumberThreadFactory() {
             SecurityManager s = System.getSecurityManager();
             this.group = s != null ? s.getThreadGroup() : Thread.currentThread().getThreadGroup();
+            this.namePrefix = "cucumber-runner-" + poolNumber.getAndIncrement() + "-thread-";
         }
         
         @Override
         public Thread newThread(Runnable r) {
-            Thread t = new Thread(this.group, r, "cucumber-runner-thread-" + this.threadNumber.getAndIncrement(), 0L);
-            if (t.isDaemon()) {
-                t.setDaemon(false);
-            }
-
-            if (t.getPriority() != 5) {
-                t.setPriority(5);
-            }
-
-            return t;
+            return new Thread(r, namePrefix + this.threadNumber.getAndIncrement());
         }
     }
 
--- a/core/src/main/java/cucumber/runtime/Runtime.java
+++ b/core/src/main/java/cucumber/runtime/Runtime.java
@@ -224,7 +224,7 @@ public class Runtime {
         
         @Override
         public Thread newThread(Runnable r) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/cucumber/runtime/Runtime.java
```diff
@@ -210,6 +212,31 @@ public Runtime build() {
...
     }
 
+    private static final class CucumberThreadFactory implements ThreadFactory {
+
+        private final AtomicInteger threadNumber = new AtomicInteger(1);
+        private final ThreadGroup group;
+
+        CucumberThreadFactory() {
+            SecurityManager s = System.getSecurityManager();
+            this.group = s != null ? s.getThreadGroup() : Thread.currentThread().getThreadGroup();
```
mpkorstanje: I don't think it's required nor desirable to manually set the thread group. The `ThreadGroup` for the workers should be the same as the `ThreadGroup` of `Runtime.run`. Currently it is the `ThreadGroup` of `Runtime.Builder.build` which need not be the same. Using a different constructor will automatically use the correct `ThreadGroup`.

core/src/main/java/cucumber/runtime/Runtime.java
```diff
@@ -210,6 +212,31 @@ public Runtime build() {
...
+        }
+        
+        @Override
+        public Thread newThread(Runnable r) {
+            Thread t = new Thread(this.group, r, "cucumber-runner-thread-" + this.threadNumber.getAndIncrement(), 0L);
```
mpkorstanje: Using the `Thread(String name)` constructor should be the safest, least confusing option here.

core/src/main/java/cucumber/runtime/Runtime.java
```diff
@@ -210,6 +212,31 @@ public Runtime build() {
...
+            if (t.isDaemon()) {
+                t.setDaemon(false);
```
mpkorstanje: Is there a reason to override the inherited settings?

core/src/main/java/cucumber/runtime/Runtime.java
```diff
@@ -210,6 +212,31 @@ public Runtime build() {
...
+            }
+
+            if (t.getPriority() != 5) {
+                t.setPriority(5);
```
mpkorstanje: Is there a reason to override the inherited settings?

kutzi:
I just wanted to mirror the behaviour of Executors.defaultThreadFactory() - that's the one which was used effectively before my change.
See https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/util/concurrent/Executors.java#L572

But I'm also fine with removing some of the none strictly necessary stuff (groups, priority)

kutzi:
Actually, we should probably still keep the `pool-x` prefix, as it *is* possible to start multiple cucumber Runtimes in the same JVM

kutzi:
Maybe `cucumber-runner-x-y` with x the pool and y the thread?

mpkorstanje:
I'd say `cucumber-runner-x-thread-y` for clarity.

mpkorstanje:
> I just wanted to mirror the behaviour of Executors.defaultThreadFactory() - that's the one which was used effectively before my change - as closely as possible.

Ah that makes sense. But it's all behavior we don't quite need. So it's better to keep the code simple.

mpkorstanje:
Cheers!

Issue #1628, opened 2019-05-10
Cucumber Runtime swallows exceptions in Executor pool threads when run with --threads > 1


kutzi:
i.e it will log them, dispose the executor thread and then execute the next pickle in a new executor thread.
I have seen this behaviour when there was a bug in my ConcurrentEventListener. It didn't make the cucumber run fail!

See https://github.com/cucumber/cucumber-jvm/pull/1389#discussion_r282840714

PR #1629, opened 2019-05-10
[Core] Handle parallel execution exceptions

kutzi:
See https://github.com/cucumber/cucumber-jvm/issues/1628

That should fix the mentioned issue sufficiently.
There would be some nice-to-have improvements IMO, e.g. reporting all the exceptions if multiple threads failed and failing as early as possible instead of waiting for the termination of all the callables, but that would increase the complexity.

### DIFF
--- a/core/src/main/java/cucumber/runtime/Runtime.java
+++ b/core/src/main/java/cucumber/runtime/Runtime.java
@@ -99,7 +99,9 @@ public class Runtime {
         for (Future f : futures) {
             try {
                 f.get();
-            } catch (ExecutionException | InterruptedException e) {
+            } catch (ExecutionException e){
+                thrown.add(e.getCause());
+            } catch (InterruptedException e) {
                 thrown.add(e);
             }
         }
@@ -112,12 +114,13 @@ public class Runtime {
         bus.send(new TestRunFinished(bus.getTime(), bus.getTimeMillis()));
     }
 
-    class MultipleFailureException extends CucumberException {
+    static class MultipleFailureException extends CucumberException {
         private final List<Throwable> throwables;
 
         MultipleFailureException(List<Throwable> throwables) {
-            super("There ");
+            super("");
             this.throwables = throwables;
+            initCause(throwables.get(0));
         }
 
         public List<Throwable> getFailures() {
@@ -126,6 +129,8 @@ public class Runtime {
 
         public String getMessage() {
             StringBuilder sb = new StringBuilder(String.format("There were %d exceptions:", this.throwables.size()));
+            sb.append("\nThe first exception is shown as cause of this MultipleFailureException.")
+                .append("\nThe other exceptions are these (without stack traces):");
 
             for (Throwable e : this.throwables) {
                 sb.append(String.format("\n  %s(%s)", e.getClass().getName(), e.getMessage()));
--- a/core/src/main/java/cucumber/runtime/Runtime.java
+++ b/core/src/main/java/cucumber/runtime/Runtime.java
@@ -19,11 +19,12 @@ import cucumber.runtime.io.ResourceLoaderClassFinder;
 import cucumber.runtime.model.CucumberFeature;
 import cucumber.runtime.model.FeatureLoader;
 import gherkin.events.PickleEvent;
+import io.cucumber.core.logging.Logger;
+import io.cucumber.core.logging.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.AbstractExecutorService;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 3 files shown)


mpkorstanje:
This is fixing the symptoms. Not the cause. I think the proper fix would be to add try-catch blocks in `AbstractEventPublisher.send`. You can use `io.cucumber.core.logging.LoggerFactory` to obtain a logger and log any exceptions.

kutzi:
These are 2 different issue IMO.

a) That's what you are referring to in your comment, if I get it: make sure a faulty EventHandler doesn't break anything else. I'm not sure if we even want to do that, as I would rather fail hard in that case.

b) make sure that uncaught exceptions are not silently 'swallowed' - i.e. only logged. That's what this PR is about

mpkorstanje:
Fair enough.

mpkorstanje:
I think the run method should look like this:

```java
    public void run() {
        final List<CucumberFeature> features = featureSupplier.get();
        bus.send(new TestRunStarted(bus.getTime(), bus.getTimeMillis()));
            feature.sendTestSourceRead(bus);
        }

        final StepDefinitionReporter stepDefinitionReporter = plugins.stepDefinitionReporter();
        runnerSupplier.get().reportStepDefinitions(stepDefinitionReporter);

        final List<Future<?>> futures = new ArrayList<>();
            for (final PickleEvent pickleEvent : feature.getPickles()) {
                    futures.add(executor.submit(new Runnable() {
                        @Override
                        public void run() {
                        }
                    }));
                }
            }
        }
        executor.shutdown();

        for (Future<?> future : futures) {
            try {
                future.get();
            } catch (InterruptedException | ExecutionException e) {
                throw new CucumberException(e);
            }
        }

        bus.send(new TestRunFinished(bus.getTime(), bus.getTimeMillis()));
    }
```

Using `executor.submit` we immediately get a future from the runnable.

And since we're checking the result of all futures we no longer need to await shutdown.

mpkorstanje:
Regardless of that:

```
        for (Future<?> future : futures) {
            try {
                future.get();
            } catch (InterruptedException | ExecutionException e) {
                throw new CucumberException(e);
            }
        }
```

We now have a separate problem. We are still swallowing exceptions here as we only display the exception thrown by the first pickle. This may not be the exception that killed the system.

kutzi:
right, just using submit(Runnable) is easier. I missed that.
Also, I can remove the await

kutzi:
Regarding your 2nd comment: yes, that's what I mentioned already. We're only re-throwing the 1st exception, but that's still indefinitely better than the current behaviour.

mpkorstanje:
Not in a concurrent system.

There is a race condition between the thread that caused the initial problem in the plugin (for example) and the thread that ran into problems b.c. the plugin is now in an invalid state.

mpkorstanje:
We should at least log all sumitted futures and their results. Then throw a composite exception maybe?

kutzi:
AFAIK, all other errors would still be logged already - as currently.
Can you propose an improvement to the PR?

mpkorstanje:
After `f.get();` throws we ignore subsequent failures.Ill try to think ofan improvement

mpkorstanje:
Done. This should work. Probably needs some improvements but I'm out of time for this week.

core/src/main/java/cucumber/runtime/Runtime.java
```diff
@@ -92,23 +93,48 @@ public void run() {
...
         }
 
         bus.send(new TestRunFinished(bus.getTime(), bus.getTimeMillis()));
     }
 
+    class MultipleFailureException extends CucumberException {
+        private final List<Throwable> throwables;
+
+        MultipleFailureException(List<Throwable> throwables) {
+            super("There ");
```
kutzi: Leftover?

core/src/main/java/cucumber/runtime/Runtime.java
```diff
@@ -75,29 +79,62 @@ public void run() {
...
+
+        public List<Throwable> getFailures() {
+            return Collections.unmodifiableList(this.throwables);
+        }
+
+        public String getMessage() {
+            StringBuilder sb = new StringBuilder(String.format("There were %d exceptions:", this.throwables.size()));
+
+            for (Throwable e : this.throwables) {
+                sb.append(String.format("\n  %s(%s)", e.getClass().getName(), e.getMessage()));
```
kutzi: This doesn't log the stack traces for the original exceptions, does it?
Could make searching for the root cause difficult.

core/src/main/java/cucumber/runtime/Runtime.java
mpkorstanje: CucumberException doesn't have a default constructor (yet).

core/src/main/java/cucumber/runtime/Runtime.java
mpkorstanje: Feel free to push improvements!

mpkorstanje:
I think we're good to go now.

kutzi:
Yes, looks good to me.

mpkorstanje:
Cheers!

Issue #1633, opened 2019-05-21
StepDefinitionReporter fails to report Java8 Lambda StepDefinitions


timtebeek:
Ok so I feel silly asking about this here, but I can't workout why the `StepDefinitionReporter` isn't triggered anymore when I follow the steps to implement my plugin here:
https://blog.jdriven.com/2019/01/cucumber-jvm-plugin/

This is in a new project, using recent (4.2.0, 4.2.6, 4.3.1) versions of cucumber-jvm. Somehow the `@Override public void stepDefinition(StepDefinition stepDefinition) {}` isn't triggered, and I can't find out why. @mpkorstanje we've both seen this work before, any idea why?

The only noticeable difference between this new project and my previous attempts, is that this project uses the `cucumber.api.java8.Nl` together with `#language: nl`.

mpkorstanje:
Does it work with annotation based step definitions?

timtebeek:
Created a sample project here: https://github.com/timtebeek/cucumber-jvm-issues-1633
It appears as though `StepDefinitionReporter#stepDefinition(StepDefinition stepDefinition)` is only ever invoked for the Annotation based step; Language appears not to be a factor (luckily).

So any clue how as to I can have the `StepDefinitionReporter` invoked for Java lambda based step definitions? Or is there any other hook I can use to pick up on those being defined?

mpkorstanje:
So Tim, how do you feel about contributing to open source? I'm afraid this will require a PR.
 :smile: 

Lamda step definitions are only registered when a step definition class is instantiated. This is done in  `JavaBackend.buildWorld`. However we trigger `Runner.reportStepDefinitions` way before that happens. So there you could invoke `Runner.buildWorld` before `glue.reportStepDefinitions`. Problem is that building the world can be a rather expensive operation (e.g. when used with Spring). 

I would prefer that whenever `Glue.addStepDefinition` is invoked we put an event on the bus and deprecate the `StepDefinitionReporter` interface all together. Plumbing wise it shouldn't be too difficult. The `Runner` is already instantiated with an instance of `EventBus` and it instantiates the `Glue` class itself.

timtebeek:
@mpkorstanje Perfect, thanks for the detailed instructions; there's a first attempt in: https://github.com/cucumber/cucumber-jvm/pull/1634
If you agree with the changes there I can add tests if needed.

mpkorstanje:
Looks good! Please continue. Does it work as expected with your test project?

[...]

PR #1634, opened 2019-05-21
[Core] Add StepDefinedEvent

timtebeek:
## Summary

Adds a `StepDefinedEvent` to replace the `StepDefinitionReporter` plugin interface.

## Details

Publish an event from `Glue.addStepDefinition`, as `Runner.reportStepDefinitions` is called before step definition classes are instantiated, which fails to pick up on lambda based steps.

## Motivation and Context

Provide an event to listen for steps being registered, so all steps registrations are available, instead of only the annotation based step definitions.
https://github.com/cucumber/cucumber-jvm/issues/1633

## How Has This Been Tested?

Not yet; Change seems small.

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/api/StepDefinitionReporter.java
+++ b/core/src/main/java/cucumber/api/StepDefinitionReporter.java
@@ -1,7 +1,12 @@
 package cucumber.api;
 
+import cucumber.api.event.StepDefinedEvent;
 import cucumber.runtime.StepDefinition;
 
+/**
+ * @deprecated in favor of {@link StepDefinedEvent}, as Lambda Step Definitions are not reported through this class.
+ */
+@Deprecated
 public interface StepDefinitionReporter extends Plugin {
     /**
      * Called when a step definition is defined
--- a/core/src/main/java/cucumber/api/event/CanonicalEventOrder.java
+++ b/core/src/main/java/cucumber/api/event/CanonicalEventOrder.java
@@ -31,6 +31,7 @@ final class CanonicalEventOrder implements Comparator<Event> {
                 TestRunStarted.class,
             TestSourceRead.class,
             SnippetsSuggestedEvent.class,
+            StepDefinedEvent.class,
             TestCaseEvent.class,
             TestRunFinished.class
         );
--- a/core/src/main/java/cucumber/api/event/EventPublisher.java
+++ b/core/src/main/java/cucumber/api/event/EventPublisher.java
@@ -11,6 +11,7 @@ public interface EventPublisher {
      * <li>{@link TestRunStarted} - the first event sent.
      * <li>{@link TestSourceRead} - sent for each feature file read, contains the feature file source.
      * <li>{@link SnippetsSuggestedEvent} - sent for each step that could not be matched to a step definition, contains the raw snippets for the step.
+     * <li>{@link StepDefinedEvent} - sent for each step definition as it is loaded, contains the StepDefinition
      * <li>{@link TestCaseStarted} - sent before starting the execution of a Test Case(/Pickle/Scenario), contains the Test Case
      * <li>{@link TestStepStarted} - sent before starting the execution of a Test Step, contains the Test Step
      * <li>{@link EmbedEvent} - calling scenario.embed in a hook triggers this event.
--- a/core/src/main/java/cucumber/api/event/StepDefinedEvent.java
+++ b/core/src/main/java/cucumber/api/event/StepDefinedEvent.java
@@ -1,0 +1,13 @@
+package cucumber.api.event;
+
+import cucumber.runtime.StepDefinition;
+
+public class StepDefinedEvent extends TimeStampedEvent {
+    public final StepDefinition stepDefinition;
+
+    public StepDefinedEvent(Long time, Long timeMillis, StepDefinition stepDefinition) {
+        super(time, timeMillis);
+        this.stepDefinition = stepDefinition;
+    }
+
+}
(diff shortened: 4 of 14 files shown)


timtebeek:
Hmm; trying to see if I understand you correctly; You want to (for instance) add a `disposeScenarioScope` method in `Java8StepDefinition` which is called in `Glue.removeScenariosScopedStepDefinitions(Map<String, StepDefinition>)` ? And have that method assign `null` to the `body` field to remove the reference?

I'll be off for a few days, so will only be able to pick this up on the weekend..

timtebeek:
As for the test coverage; Agree it's "parse". Functionally the only new line is: `bus.send(new StepDefinedEvent(bus.getTime(), bus.getTimeMillis(), stepDefinition));`, but I didn't think it sensible to call it in `Glue` and verify that it's called in `GlueTest`. We've both seen waaay too much of that test pattern applied in our shared working history. ;)

mpkorstanje:
I was thinking of an integration test that creates a backend with 2 steps. Then run two scenarios with those two steps and check that we see 3 events emitted. There is utility class that makes it relatively easy to setup.

But I'll have to check.

> Hmm; trying to see if I understand you correctly; You want to (for instance) add a `disposeScenarioScope` method in `Java8StepDefinition` which is called in `Glue.removeScenariosScopedStepDefinitions(Map<String, StepDefinition>)` ? And have that method assign `null` to the `body` field to remove the reference?

Exactly.

mpkorstanje:
For testing this functionality you can find some examples in `RuntimeTest`. I reckon something like this should do. It's a bit messy but should serve to illustrate the point.

```
        final CucumberFeature feature = feature("test.feature", "" +
            "Feature: feature name\n" +
            "  Scenario: scenario name\n" +
            "    When s\n");

        Plugin plugin = new ConcurrentEventListener() {
            @Override
            public void setEventPublisher(EventPublisher publisher) {
                // Capture some events here.                
            }
        };
        ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
        final Backend mock = new Backend() {
            
            @Override
            public void loadGlue(Glue glue, List<URI> gluePaths) {
                // Register some regular glue here        
            }

            @Override
            public void buildWorld() {
                // Register some scenario scoped glue here.
            }

            @Override
            public void disposeWorld() {

            }

            @Override
            public List<String> getSnippet(PickleStep step, String keyword, FunctionNameGenerator functionNameGenerator) {
                return emptyList();
            }
        };

        BackendSupplier backendSupplier = new BackendSupplier() {
            @Override
            public Collection<? extends Backend> get() {
                return singletonList(mock);
            }
        };
        FeatureSupplier featureSupplier = new FeatureSupplier() {
            @Override
            public List<CucumberFeature> get() {
                return singletonList(feature);
            }
        };
        Runtime.builder()
            .withBackendSupplier(backendSupplier)
            .withAdditionalPlugins(plugin)
            .withResourceLoader(new ClasspathResourceLoader(classLoader))
            .withEventBus(new TimeServiceEventBus(new TimeServiceStub(0)))
            .withFeatureSupplier(featureSupplier)
            .build()
            .run();
           // Check if the plugin captured the expected events here.
```

timtebeek:
@mpkorstanje Thanks for the hint on testing. I'm still having some trouble filling in the blank here:
`public void buildWorld() { // Register some scenario scoped glue here.`
The only examples of registering Lambda based steps I see are in the Java/Java8 modules; the approach there is hard to replicate in `core` due to missing classes. Any suggestions there?

I've added ScenarioScoped changes in https://github.com/cucumber/cucumber-jvm/pull/1634/commits/81f2c9849d931fbd3623491ece1d8ed43f652207 , I hope that's what you intended!

mpkorstanje:
A mocked step definition will do. As long as it `isScenarioScoped` returns true.

timtebeek:
Hmm; Looked into it again, but really stuck on the `new Backend() { ... }` details; Just too unfamiliar with the (test) internals I suppose. Would love to see this work; but I'm lost as to how.

mpkorstanje:
I've pushed a commit. It's a bit of a hack job but should get the point across.

mpkorstanje:
I should look into making some proper test implementations of the components I just mocked.

timtebeek:
Awesome, thanks! Had been eyeing the `TestBackendSupplier`, but wouldn't have worked out I'd need anything like `MockedScenarioScopedStepDefinition` or `MockedStepDefinition` on my own.. thanks a lot!
Anything else we need to cross off before we can consider a merge?

mpkorstanje:
Its good as is. But I am going to try and sneak in a redactor or two. There is an emerging pattern here that calls for some reusable test components.

You know the ratchet mechanism bring and keep quality up. May take a bit of time. It's been busy since we last spoke. There is less time for OSS now.

timtebeek:
Ok thanks for helping out; If I can help in any way let me know; all the best there! Miss working together in Amsterdam.. :)

aslakhellesoy:
Hi @timtebeek,

Aslak Hellesøy
Creator of Cucumber

mpkorstanje:
Cheers. No worries!

timtebeek:
Awesome, great to see this merged! I'll update my unused steps plugin blog post to include these Java 8 steps. Should I start a PR to contribute that as well? It's small, has no dependencies and could be useful plus serve as example.

mpkorstanje:
Would not mind seeing it. Though the json format is problematic. Almost impossible to evolve on.

PR #1645, opened 2019-05-27
[Core] Add reverse and random scenario execution order

grasshopper7:
## Summary

Allows scenarios to be executed in random or reverse order. Custom execution orders can also be created.

## Details

All scenarios can be run in random order. 
`java <classpath> cucumber.api.cli.Main --order random -g <glue path> <feature file path>`
The number of scenarios to be run in random can also be mentioned.
`java <classpath> cucumber.api.cli.Main --order random --count <number> -g <glue> <feature file>`

All scenarios can be run in reverse order.
`java <classpath> cucumber.api.cli.Main --order reverse -g <glue path> <feature file path>`

Custom orders can be created by implementing the `PickleOrder` interface in the `cucumber.runtime.order` package. 

## Motivation and Context

Scenarios can be executed in random order. Aids in finding out if their is any dependency between scenarios.

## How Has This Been Tested?

Tests for `RuntimeOptions` are added in `RuntimeOptionsTest` for testing options in the command. `OrderTypeFactoryTest` tests the `OrderTypeFactory` class. `RandomOrderTypeTest` tests the `RandomOrderType` class. No test for `ReverseOrderType` as it will test the `reverse` method of `Collections` class.

## Types of changes

- [x ] New feature (non-breaking change which adds functionality).

## Checklist:

- [x ] I've added tests for my code.


core/src/main/java/cucumber/runtime/RuntimeOptions.java
```diff
@@ -180,6 +184,10 @@ private void parse(List<String> args) {
                 parsedJunitOptions.addAll(asList(arg.substring("--junit,".length()).split(",")));
             } else if (arg.equals("--wip") || arg.equals("-w")) {
                 wip = true;
+            } else if (arg.equals("--order")) {
+            	orderTypeData.put(OrderType.TYPE_NAME, args.remove(0));
```
mpkorstanje: The names of the CLI commands are specific to the runtime options parser. It is not necessary for the `OrderType` to know what it is named on the CLI. It is okay to inline this string here.

core/src/main/java/cucumber/runtime/RuntimeOptions.java
```diff
@@ -214,6 +222,10 @@ private void parse(List<String> args) {
             junitOptions.clear();
             junitOptions.addAll(parsedJunitOptions);
         }
+        if(orderTypeData.containsKey(OrderType.TYPE_NAME)) {
+        	orderType = OrderTypeFactory.getOrderType(orderTypeData);
+        	orderType.checkVariableValues(orderTypeData);
```
mpkorstanje: The exact syntax of CLI commands is also specific to the runtime options parser. Checking if the input is valid should be done here.

Additionally any time when `--count` is used it's argument should be a number and should non-negative. Tying the validation to the order type makes this more complex then needed. 

I think it might be better to make `--count` and `--order` independent variables and simply ignore `--count` when not relevant.

core/src/main/java/cucumber/runtime/order/OrderType.java
```diff
@@ -0,0 +1,26 @@
+package cucumber.runtime.order;
+
+import java.util.List;
+import java.util.Map;
+
+import gherkin.events.PickleEvent;
+
+public abstract class OrderType {
+
+	public static final String NONE_ORDER_TYPE = "None";
```
mpkorstanje: Type names should be lower case or case insensitive.

core/src/main/java/cucumber/runtime/order/OrderType.java
```diff
@@ -0,0 +1,26 @@
+package cucumber.runtime.order;
+
+import java.util.List;
+import java.util.Map;
+
+import gherkin.events.PickleEvent;
+
+public abstract class OrderType {
```
mpkorstanje: I think this class can be turned into an interface. The default implementation for `orderPickleEvents` is only used by `none` order.

core/src/main/java/cucumber/runtime/order/OrderTypeFactory.java
```diff
@@ -0,0 +1,28 @@
+package cucumber.runtime.order;
+
+import java.util.Map;
+
+import cucumber.runtime.CucumberException;
+
+public class OrderTypeFactory {
```
mpkorstanje: This mapping belongs in `RuntimeOptions`. Additionally I think the different orders, could be turned into an enum that implements/extends the `OrderType` interface/class.

core/src/main/java/cucumber/runtime/order/RandomOrderType.java
```diff
@@ -0,0 +1,39 @@
+package cucumber.runtime.order;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import cucumber.runtime.CucumberException;
+import gherkin.events.PickleEvent;
+
+public class RandomOrderType extends OrderType {
```
mpkorstanje: This should be an Enum that implements/extends `OrderType`.

core/src/main/java/cucumber/runtime/order/RandomOrderType.java
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	public RandomOrderType(int count) {
+		this.count = count;
+	}
+
+	@Override
+	public List<PickleEvent> orderPickleEvents(List<PickleEvent> pickleEvents) {	
+		Collections.shuffle(pickleEvents);
+		return pickleEvents.subList(0, ((count > pickleEvents.size() || count == 0) ?
```
mpkorstanje: Nifty but way to hard to follow. 

I think this class is doing too much. It's both ordering and selecting pickles. It should only order pickles. By making `--order` and `--count` independent of each other you can remove this complexity all together.

core/src/main/java/cucumber/runtime/Runtime.java
```diff
@@ -81,20 +81,25 @@ public void run() {
...
-                            runnerSupplier.get().runPickle(pickleEvent);
-                        }
-                    }));
+                	filteredEvents.add(pickleEvent);
                 }
             }
         }
+        
+        final List<Future<?>> executingPickles = new ArrayList<>();
+        for(final PickleEvent pickleEvent : runtimeOptions.getOrderType().orderPickleEvents(filteredEvents)) {
```
mpkorstanje: Ideally the order is injected into the `Runtime` via the builder. This allows us and other implementations to create `Runtime` instances without passing in CLI based `RuntimeOptions`.

This will in the long run let us separate the parsing of CLI options from executing pickles.

core/src/main/java/cucumber/runtime/order/RandomOrderType.java
mpkorstanje: https://stackoverflow.com/questions/2709593/why-would-an-enum-implement-an-interface

core/src/main/java/cucumber/runtime/Runtime.java
mpkorstanje: For example:

```
        List<PickleEvent> allPickles = new ArrayList<>();
            allPickles.addAll(feature.getPickles());
        }

        List<PickleEvent> orderedPickles = pickleOrder.order(allPickles);
        List<PickleEvent> filteredPickles = pickleFilter.filter(orderedPickles);

        final List<Future<?>> executingPickles = new ArrayList<>();
        for (PickleEvent filtredPickle : filteredPickles) {
                executingPickles.add(executor.submit(new Runnable() {
                    @Override
                    public void run() {
                    }
                }));
            }
        }
```

core/src/main/java/cucumber/runtime/RuntimeOptions.java
grasshopper7: Will change it.

core/src/main/java/cucumber/runtime/RuntimeOptions.java
grasshopper7: The `--count`  is used only for the `RandomOrderType` thus the check was specific to that type. That required the test for number etc to be done later.

Idea of making `--count` applicable for all types, to be used if required, does simplifies things. If so then the `orderTypeData` map is also redundant and can be removed. 

Later on if a future `OrderType` requires an additional variable from CLI it needs to be handled accordingly.

core/src/main/java/cucumber/runtime/order/OrderType.java
grasshopper7: Will change it.

core/src/main/java/cucumber/runtime/order/RandomOrderType.java
grasshopper7: Yeah my bad. Though this logic needs to remain, as need to make sure that the `count` is less or equal to the size of the incoming PickleEvents to avoid exception.

core/src/main/java/cucumber/runtime/Runtime.java
grasshopper7: To implement the above the existing iteration for the filters will need to be moved to the `Filters` class. Are you suggesting above that the pickles need to be ordered first and then filtered?

core/src/main/java/cucumber/runtime/order/RandomOrderType.java
grasshopper7: Will convert into enum.  No need of Factory.

core/src/main/java/cucumber/runtime/order/RandomOrderType.java
mpkorstanje: No worries!

core/src/main/java/cucumber/runtime/Runtime.java
mpkorstanje: Yes. It'd be much easier to implement this with a Java8 `Stream` but we don't have those at the moment. So I don't mind too much where the iteration ends up.

grasshopper7:
Implemented `OrderType` as an enumeration implementing `OrderPickleEvents` interface. 

Separated `--order` and `--count` CLI options to work independently. The `--order` logic is handled by `OrderType` enum. While the `--count` is handled by the `limitPickleEvents` method in `Filters` class. 

Updated the test classes accordingly.

mpkorstanje:
Looks good! I'll need a bit more time to give this a proper once over.

core/src/main/java/cucumber/runtime/Runtime.java
```diff
@@ -52,14 +53,16 @@
...
 
     public Runtime(final Plugins plugins,
                    final RuntimeOptions runtimeOptions,
                    final EventBus bus,
                    final Filters filters,
                    final RunnerSupplier runnerSupplier,
                    final FeatureSupplier featureSupplier,
-                   final ExecutorService executor) {
+                   final ExecutorService executor,
+                   final OrderType orderType) {
```
mpkorstanje: Did you mean to use `OrderPickleEvents` here?

core/src/main/java/cucumber/runtime/order/OrderType.java
```diff
@@ -0,0 +1,41 @@
...
+		public List<PickleEvent> orderPickleEvents(List<PickleEvent> pickleEvents) {
+			Collections.shuffle(pickleEvents);
+			return pickleEvents;
+		}		
+	};
+
+
+     public static OrderType getOrderType(String name) {
+        for (OrderType orderType : OrderType.values()) {
+            if (name.equalsIgnoreCase(orderType.name())) {
```
mpkorstanje: Ideally we parse the different order names in runtime options too. This would avoid having to support both lower and uppercase options.

PR #1648, opened 2019-06-03
[Core] Add UnusedStepsSummaryPrinter

timtebeek:
## Summary

Adds a plugin to find and report on unused steps.

## Details

As discussed in https://github.com/cucumber/cucumber-jvm/pull/1634 it uses the new StepDefinedEvent to locate all registered steps, and prints a summary of unused steps to the argument file.

## Motivation and Context

Allows to easily find unused steps, which might be indicative of removed functionality or missed test coverage.

## How Has This Been Tested?

Not yet; as @mpkorstanje indicated some hesitation on whether this would warrant adoption. This PR is to record that discussion, and will be updated depending on the outcome. Same goes for the current Java8 baseline, so it might need to target a future version rather than the next.

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/runtime/formatter/UnusedStepsSummaryPrinter.java
+++ b/core/src/main/java/cucumber/runtime/formatter/UnusedStepsSummaryPrinter.java
@@ -1,39 +1,76 @@
 package cucumber.runtime.formatter;
 
 import java.util.Map;
-import java.util.Optional;
+import java.util.Map.Entry;
 import java.util.TreeMap;
 
+import cucumber.api.Result;
 import cucumber.api.SummaryPrinter;
 import cucumber.api.event.*;
+import cucumber.api.formatter.ColorAware;
 import cucumber.api.formatter.NiceAppendable;
 
-public class UnusedStepsSummaryPrinter implements EventListener, SummaryPrinter {
+public class UnusedStepsSummaryPrinter implements ColorAware, EventListener, SummaryPrinter {
 
-	private final Map<String, String> unusedSteps = new TreeMap<>();
+	private EventHandler<StepDefinedEvent> stepDefinedHandler = new EventHandler<StepDefinedEvent>() {
+		@Override
+		public void receive(StepDefinedEvent event) {
+			unusedSteps.put(event.stepDefinition.getLocation(false), event.stepDefinition.getPattern());
+		}
+	};
+	private EventHandler<TestStepFinished> testStepFinishedHandler = new EventHandler<TestStepFinished>() {
+		@Override
+		public void receive(TestStepFinished event) {
+			String codeLocation = event.testStep.getCodeLocation();
+			if (codeLocation != null) {
+				unusedSteps.remove(codeLocation);
+			}
+		}
+	};
+	private EventHandler<TestRunFinished> testRunFinishedhandler = new EventHandler<TestRunFinished>() {
+		@Override
+		public void receive(TestRunFinished event) {
+			if (unusedSteps.isEmpty()) {
+				return;
+			}
 
-	private NiceAppendable out;
+			Format format = formats.get(Result.Type.UNUSED.lowerCaseName());
+			out.println(format.text(unusedSteps.size() + " Unused steps:"));
+
+			// Output results when done
+			for (Entry<String, String> entry : unusedSteps.entrySet()) {
+				String location = entry.getKey();
+				String pattern = entry.getValue();
+				out.println(format.text(location + " # " + pattern));
+			}
+		}
+	};
+
+	final Map<String, String> unusedSteps = new TreeMap<>();
+	final NiceAppendable out;
+	Formats formats;
 
 	@SuppressWarnings("WeakerAccess") // Used by PluginFactory
-	public UnusedStepsPlugin(Appendable out) {
+	public UnusedStepsSummaryPrinter(Appendable out) {
 		this.out = new NiceAppendable(out);
 	}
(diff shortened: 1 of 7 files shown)


mpkorstanje:
Could you show me an example of the output?

timtebeek:
The current output is:
```
2 Unused steps:
SampleSteps.java:17: this step is unused
SampleSteps.maar(): another unused step
```
For a sample project here: https://github.com/timtebeek/cucumber-jvm-issues-1633/tree/step-defined-event/src/test/java/com/github/timtebeek

mpkorstanje:
Looks good. I was initially thinking we should use the same format as the summary printer, but we may as well add it to the summary printer. That way there is no need for an additional `--dry-run` or separate configuration (for most use cases).

So let's add it to the default summary printer! 

For example:

```
Undefined scenarios:
io/cucumber/skeleton/belly.feature:3 # a few cukes
io/cucumber/skeleton/belly2.feature:3 # a few cukes

Unused steps:
SampleSteps.java:17 # this step is unused
SampleSteps.maar() # another unused step

2 Scenarios (2 undefined)
6 Steps (4 undefined, 2 passed)
2 Unused steps 
0m0.138s
```

* The separator is a hash rather then then a colon. 
* If there are no unused steps those sections need not be printed. 
* The locations should be printed in yellow unless is enabled.

image

mpkorstanje:
Okay. That was a terrible idea. Its nice when running the full suite but a nuisance when running a selection.

Let's not add it, but let's use the same format.

timtebeek:
Running a selection is indeed why I had implemented this as a separate plugin so far; thanks for the suggestions; I've updated the plugin accordingly. Not really sure how to cover this with tests, and/or if we're still on track for an eventual merge..

mpkorstanje:
We are good. I think a simple unit test where events go in and and the right output string is produced should be sufficient.

timtebeek:
Perfect; added `UnusedStepsSummaryPrinterTest` in https://github.com/cucumber/cucumber-jvm/pull/1648/commits/679dd3f5992ee3b3d06ac0a6d579dbe71f2caa63 , hope that matches what you were looking for. Anything else before we can consider a merge?

core/src/main/java/cucumber/runtime/formatter/UnusedStepsSummaryPrinter.java
```diff
@@ -0,0 +1,76 @@
...
+				String location = entry.getKey();
+				String pattern = entry.getValue();
+				out.println(format.text(location + " # " + pattern));
+			}
+		}
+	};
+
+	final Map<String, String> unusedSteps = new TreeMap<>();
+	final NiceAppendable out;
+	Formats formats = new MonochromeFormats();
```
mpkorstanje: `private` is missing I guess.

core/src/main/java/cucumber/runtime/formatter/UnusedStepsSummaryPrinter.java
```diff
@@ -0,0 +1,76 @@
...
+			}
+
+			Format format = formats.get(Result.Type.UNUSED.lowerCaseName());
+			out.println(format.text(unusedSteps.size() + " Unused steps:"));
+
+			// Output results when done
+			for (Entry<String, String> entry : unusedSteps.entrySet()) {
+				String location = entry.getKey();
+				String pattern = entry.getValue();
+				out.println(format.text(location + " # " + pattern));
```
mpkorstanje:
```suggestion
				out.println(format.text(location) + " # " + pattern);
```

Only the location is in yellow.

core/src/main/java/cucumber/runtime/formatter/UnusedStepsSummaryPrinter.java
```diff
@@ -0,0 +1,76 @@
...
+			}
+		}
+	};
+
+	final Map<String, String> unusedSteps = new TreeMap<>();
```
mpkorstanje:
```suggestion
	private final Map<String, String> unusedSteps = new TreeMap<>();
```

core/src/main/java/cucumber/runtime/formatter/UnusedStepsSummaryPrinter.java
```diff
@@ -0,0 +1,76 @@
...
+	final NiceAppendable out;
```
mpkorstanje:
```suggestion
	private final NiceAppendable out;
```

core/src/main/java/cucumber/runtime/formatter/UnusedStepsSummaryPrinter.java
mpkorstanje:
```suggestion
	private Formats formats = new MonochromeFormats();
```

timtebeek:
Thanks for directly applying your own suggestions.. as a note: I'd used package private modifier as the inner class event handlers otherwise need to use syntactic accessors to access the private fields (or so my IDE tells me), but the code should work fine as it is now, and probably clearer.

Shall I merge? Or feel free to merge of that's faster.. :) great to see this added, saves me copy and pasting it where I need it.

mpkorstanje:
It's a new github feature. Really nice. 

I've merged it. I've been using a format that makes it look reasonably nice. But I still need to document it so other people can use it too.

timtebeek:
Perfect, thanks for the review/guidance/merge! :)

mpkorstanje:
You're welcome. Anytime.

PR #1656, opened 2019-06-06
[Core] Allow runner to register bus as concurrent or serial event source

### LINKED ISSUE #1515: Output is generated only after running all features
## Output is only generated after running all features

Hi,
I am using cucumber-jvm version **4.2.0**
When I am running all features via **JUnit** and **Maven**, no output with the running scenarios steps is displayed. The features,  scenarios and the corresponding steps results are displayed only when all cucumber tests finish.
After changing cucumber-jvm version to **3.0.2**, I saw that the output from Cucumber (running features, scenarios and steps) is displayed in real time.

timtebeek:
Fixes #1515

## Summary

As discussed in #1515 the pretty text output for cli/junit/testNg is also delayed for users running on a single thread from their IDE or commandline. This small change restores immediate text output when running on a single thread, only delaying when multiple threads are involved.

A better fix would be an overhaul of the PrettyFormatter, but that's considerably more work, whereas this would fix a lot of the same use cases.

## Details

Do not create or use the CanonicalOrderEventPublisher for plugins when running on a single thread.

## Motivation and Context

It ensures `non-ConcurrentEventListener` plugins get events immediately when running on a single thread, rather than delayed until all tests are run.

## How Has This Been Tested?

Not yet, as it might be considered too hacky.

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/io/cucumber/core/options/PluginOptions.java
+++ b/core/src/main/java/io/cucumber/core/options/PluginOptions.java
@@ -9,4 +9,5 @@ public interface PluginOptions {
 
     boolean isMonochrome();
 
+    boolean assumeEventsInOrder();
 }
--- a/core/src/main/java/io/cucumber/core/options/PluginOptions.java
+++ b/core/src/main/java/io/cucumber/core/options/PluginOptions.java
@@ -8,6 +8,4 @@ public interface PluginOptions {
     boolean isStrict();
 
     boolean isMonochrome();
-
-    boolean assumeEventsInOrder();
 }
--- a/core/src/main/java/io/cucumber/core/options/PluginOptions.java
+++ b/core/src/main/java/io/cucumber/core/options/PluginOptions.java
@@ -8,4 +8,5 @@ public interface PluginOptions {
     boolean isStrict();
 
     boolean isMonochrome();
+
 }
--- a/core/src/test/java/cucumber/runtime/formatter/PluginsTest.java
+++ b/core/src/test/java/cucumber/runtime/formatter/PluginsTest.java
@@ -71,8 +71,18 @@ public class PluginsTest {
 
 
     @Test
-    public void shouldSetNonConcurrentEventListener() {
+    public void shouldSetConcurrentEventListenerForSingleThread() {
         RuntimeOptions runtimeOptions = new RuntimeOptions(Collections.<String>emptyList());
+        Plugins plugins = new Plugins(getSystemClassLoader(), pluginFactory, rootEventPublisher, runtimeOptions);
+        EventListener plugin = Mockito.mock(EventListener.class);
+        plugins.addPlugin(plugin);
+        verify(plugin, times(1)).setEventPublisher(rootEventPublisher);
+    }
+
+
+    @Test
+    public void shouldSetNonConcurrentEventListenerForMultiThread() {
+        RuntimeOptions runtimeOptions = new RuntimeOptions("--threads 2");
         Plugins plugins = new Plugins(getSystemClassLoader(), pluginFactory, rootEventPublisher, runtimeOptions);
         EventListener plugin = Mockito.mock(EventListener.class);
         plugins.addPlugin(plugin);
@@ -83,7 +93,7 @@ public class PluginsTest {
 
     @Test
     public void shouldRegisterCanonicalOrderEventPublisherWithRootEventPublisher() {
-        RuntimeOptions runtimeOptions = new RuntimeOptions(Collections.<String>emptyList());
+        RuntimeOptions runtimeOptions = new RuntimeOptions("--threads 2");
         Plugins plugins = new Plugins(getSystemClassLoader(), pluginFactory, rootEventPublisher, runtimeOptions);
         EventListener plugin = Mockito.mock(EventListener.class);
         plugins.addPlugin(plugin);
--- a/core/src/test/java/cucumber/runtime/formatter/PluginsTest.java
+++ b/core/src/test/java/cucumber/runtime/formatter/PluginsTest.java
@@ -73,6 +73,7 @@ public class PluginsTest {
     @Test
     public void shouldSetConcurrentEventListenerForSingleThread() {
         RuntimeOptions runtimeOptions = new RuntimeOptions(Collections.<String>emptyList());
+        runtimeOptions.setAssumeEventsInOrder(true);
         Plugins plugins = new Plugins(getSystemClassLoader(), pluginFactory, rootEventPublisher, runtimeOptions);
         EventListener plugin = Mockito.mock(EventListener.class);
         plugins.addPlugin(plugin);
--- a/core/src/test/java/cucumber/runtime/formatter/PluginsTest.java
+++ b/core/src/test/java/cucumber/runtime/formatter/PluginsTest.java
@@ -46,7 +46,6 @@ public class PluginsTest {
         Plugins plugins = new Plugins(getSystemClassLoader(), pluginFactory, runtimeOptions);
         StrictAware plugin = Mockito.mock(StrictAware.class);
         plugins.addPlugin(plugin);
-        plugins.setSerialEventBusOnEventListenerPlugins(rootEventPublisher);
         verify(plugin).setStrict(true);
     }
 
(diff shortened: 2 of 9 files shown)


timtebeek:
No idea why the some JDKs suddenly get ```
[ERROR] Errors: 
[ERROR]   class gherkin.ast.Scenario cannot be cast to class gherkin.ast.Step (gherkin.ast.Scenario and gherkin.ast.Step are in unnamed module of loader 'app')
[ERROR]   class gherkin.ast.Background cannot be cast to class gherkin.ast.Step (gherkin.ast.Background and gherkin.ast.Step are in unnamed module of loader 'app')
```
but perhaps we can debug that once the rest of the code is acceptable.

**Comment by mpkorstanje** [comment]
That happens when event's are send to the pretty formatter in the wrong order.

The junit runner and test ng runner are reimplementations of the runtime. They don't use the `--threads ` option because their frameworks provide support for it. That's probably what is going wrong.

Not sure how to detect when they are being ran in parallel.

**Comment by timtebeek** [comment]
Hmm; also can't seem to find here how to detect the number of threads JUnit4 will use; So I'm not sure this approach will work without specific support for JUnit and TestNG. I'll let it rest for a while.

**Comment by mpkorstanje** [comment]
We don't need to know the exact number of threads. Only if the execution is not guaranteed to be serial. 

For JUnit `Cucumber` extends `ParentRunner` and `ParentRunner` has a method named `setScheduler`. I would assume that when a scheduler is set the execution is no longer guaranteed to be serial.

I'm at a loss for TestNG as well. But we could go with the assumption that it runs in parallel by default.

**Comment by mpkorstanje** [comment]
Pushed a thing. I'm rather unhappy about the structure of it, but it works.

It is beocmming apparent that immediately subscribing the plugins to the event bus is problematic. Would need to think about the structure.

**Comment by timtebeek** [comment]
Cool! So guess this means the PR as is at least fixes the immediate pretty text output for CLI and JUnit, and for TestNG anyone that needs it can come along and fix that once reported.

You mention you're unhappy about the structure; Is that something you still want to tackle in this PR, or make a note of in an item and tackle later?

I noticed there's a lot of work amassed in the `develop-v5` and `junit-jupiter` that would touch upon some of these same classes and issues. Would it be acceptable to merge this as it and tackle them once those branches have been merged? As any refactoring now will complicate the work in those branches, where there might be more value in having those branches brought closer to a merge as compared to the unsatisfying feeling around the current implementation here.

**Comment by mpkorstanje** [comment]
The primary difference between v4 and v5 is the package structure. Refactorings have been rather trivial to port over so far. After a few do overs you get the hang of it quite quickly. So that hasn't been a bother yet.

I was planning to EOL 4.x.x in June this year to burn out the last few bugs but there has been a splurge of contributions and fixes recently. So I'm going to reset that timer. Additionally the upgrade to V5 a significant breaking change. I'd like to give the world a heads up and a chance to contribute some deprecations to allow for a graceful migration.

I'm holding of on JUnit 5 until then. I would like to overhaul Cucumbers resource loading/class path scanning first. Which will be much easier to do once we can use Java 8. By then it will probably be easier to redo the JUnit 5 implementation from scratch.

**Comment by mpkorstanje** [comment]
And to answer the question. Yes, this is something I would like to tackle. 

There are a few requirements:

1. The plugins class should be created once along with all other components. 
2. The plugins class should be a repository of plugin objects.
3. The plugin objects should not subscribe to the event bus in the constructor.
4. The plugin objects should be subscribe to the event bus prior to sending the first event.
5. `assumeEventsInOrder` is not a plugin option nor a runtime option and should be be used as such

So I'm thinking `Plugins.setEventBusOnEventListenerPlugins(bus)` and `Plugins.setSerialEventBusOnEventListenerPlugins(bus)`.

It's mostly a plumbing job.

**Comment by mpkorstanje** [comment]
We've got a slack btw. https://cucumberbdd-slack-invite.herokuapp.com/

**Comment by timtebeek** [comment]
@mpkorstanje think I've updated the MR according to your suggestions; could you check if this matches what you were aiming for?

File: core/src/main/java/io/cucumber/core/options/PluginOptions.java [hunk 1]
```diff
@@ -8,5 +8,4 @@
     boolean isStrict();
 
     boolean isMonochrome();
```
mpkorstanje:
```suggestion
    boolean isMonochrome();

```

core/src/test/java/cucumber/runtime/formatter/PluginsTest.java
```diff
@@ -43,39 +43,54 @@
     @Test
     public void shouldSetStrictOnPlugin() {
         RuntimeOptions runtimeOptions = new RuntimeOptions("--strict");
-        Plugins plugins = new Plugins(getSystemClassLoader(), pluginFactory, rootEventPublisher, runtimeOptions);
+        Plugins plugins = new Plugins(getSystemClassLoader(), pluginFactory, runtimeOptions);
         StrictAware plugin = Mockito.mock(StrictAware.class);
         plugins.addPlugin(plugin);
+        plugins.setSerialEventBusOnEventListenerPlugins(rootEventPublisher);
```
mpkorstanje:
```suggestion
```

core/src/test/java/cucumber/runtime/formatter/PluginsTest.java
```diff
@@ -43,39 +43,54 @@
...
 
 
     @Test
     public void shouldSetMonochromeOnPlugin() {
         RuntimeOptions runtimeOptions = new RuntimeOptions("--monochrome");
-        Plugins plugins = new Plugins(getSystemClassLoader(), pluginFactory, rootEventPublisher, runtimeOptions);
+        Plugins plugins = new Plugins(getSystemClassLoader(), pluginFactory, runtimeOptions);
         ColorAware plugin = Mockito.mock(ColorAware.class);
         plugins.addPlugin(plugin);
+        plugins.setSerialEventBusOnEventListenerPlugins(rootEventPublisher);
```
mpkorstanje:
```suggestion
```

core/src/test/java/cucumber/runtime/formatter/PluginsTest.java
```diff
@@ -43,39 +43,54 @@
...
 
 
     @Test
     public void shouldSetConcurrentEventListener() {
         RuntimeOptions runtimeOptions = new RuntimeOptions(Collections.<String>emptyList());
-        Plugins plugins = new Plugins(getSystemClassLoader(), pluginFactory, rootEventPublisher, runtimeOptions);
+        Plugins plugins = new Plugins(getSystemClassLoader(), pluginFactory, runtimeOptions);
         ConcurrentEventListener plugin = Mockito.mock(ConcurrentEventListener.class);
         plugins.addPlugin(plugin);
+        plugins.setSerialEventBusOnEventListenerPlugins(rootEventPublisher);
```
mpkorstanje:
```suggestion
```

core/src/test/java/cucumber/runtime/formatter/PluginsTest.java
```diff
@@ -43,39 +43,54 @@
...
 
     @Test
-    public void shouldSetNonConcurrentEventListener() {
+    public void shouldSetConcurrentEventListenerForSingleThread() {
         RuntimeOptions runtimeOptions = new RuntimeOptions(Collections.<String>emptyList());
-        Plugins plugins = new Plugins(getSystemClassLoader(), pluginFactory, rootEventPublisher, runtimeOptions);
+        Plugins plugins = new Plugins(getSystemClassLoader(), pluginFactory, runtimeOptions);
         EventListener plugin = Mockito.mock(EventListener.class);
         plugins.addPlugin(plugin);
+        plugins.setEventBusOnEventListenerPlugins(rootEventPublisher);
```
mpkorstanje:
```suggestion
```

core/src/test/java/cucumber/runtime/formatter/PluginsTest.java
```diff
@@ -43,39 +43,54 @@
...
+    }
+
+
+    @Test
+    public void shouldSetNonConcurrentEventListenerForMultiThread() {
+        RuntimeOptions runtimeOptions = new RuntimeOptions(Collections.<String>emptyList());
+        Plugins plugins = new Plugins(getSystemClassLoader(), pluginFactory, runtimeOptions);
+        EventListener plugin = Mockito.mock(EventListener.class);
+        plugins.addPlugin(plugin);
+        plugins.setSerialEventBusOnEventListenerPlugins(rootEventPublisher);
```
mpkorstanje:
```suggestion
```

mpkorstanje:
Pushed some minor edits and reduced the diff via github. Lets see if the test pass. :laughing:

mpkorstanje:
All passed. Thanks for putting the time into this.

timtebeek:
My pleasure! :)

PR #1657, opened 2019-06-07
Added Named output hook

RobertoPegoraro:
## Summary
Capability to add name on outputs

## Details
Created new implementation to allow add name on outputs
Now is possible improve reports with beautiful output like this:
```json
"after":[
   {
      "output":[
         {
            "My Named Output":[
               "REQUEST - Executing POST on",
               "REQUEST Body: {}",
               "RESPONSE - ResponseType",
               "RESPONSE -Body: {}"
            ]
         }
      ],
      "result":{
         "duration":619511,
         "status":"passed"
      },
      "match":{
         "location":"Hooks.afterStep(Scenario)"
      }
   }
]
```

Usage:
```java 
@After
public void after(Scenario scenario){
    scenario.write("My Named Output", "message");
}
```
## Motivation and Context
I need to be able show personalized output in reports.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -193,6 +193,16 @@
                                 <elementKind>method</elementKind>
                                 <justification>This class is part of Cucumbers internal API. The change adds a vararg argument.</justification>
                             </item>
+                            <item>
+                                <code>java.method.addedToInterface</code>
+                                <new>method void cucumber.api.Scenario::write(java.lang.String, java.lang.String)</new>
+                                <package>cucumber.api</package>
+                                <classQualifiedName>cucumber.api.Scenario</classQualifiedName>
+                                <classSimpleName>Scenario</classSimpleName>
+                                <methodName>write</methodName>
+                                <elementKind>method</elementKind>
+                                <justification>This interface can be used by users of Cucumber.</justification>
+                            </item>
                         </revapi.ignore>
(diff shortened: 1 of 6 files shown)


mpkorstanje:
I think this is a good addition. However the JSON output is shared between different implementations of Cucumber. Adding it unilaterally in Cucumber-JVM would hurt the ecosystem. Please start a discussion in in the main Cucumber repo. https://github.com/cucumber/cucumber

mpkorstanje:
So given @aslakhellesoy  reply I think this unlikely to happen. You may have to maintain your own implementation of the JSON formatter.

https://github.com/cucumber/cucumber-jvm/issues/1661#issuecomment-501259804
> We're going to deprecate the JSON formatter in favour of a new message protocol:
> 
>     * https://github.com/cucumber/cucumber/tree/master/cucumber-messages
> 
>     * https://docs.google.com/document/d/12Y3Ut-SVSGdw9L_aW6NGQY9JaCcn3z0alY4d7kzOndU/edit#
> 
> 
> We might build a tool that can consume the new message protocol and create a cucumber JSON file, but we'll encourage people to start using the new message protocol instead.
> 
> It's probably 6-12 months away before we see the protocol implemented in all Cucumbers, but I wouldn't recommend putting more work into the current JSON formatter at this point.

PR #1689, opened 2019-07-05
[Core] Restore Scenario#getSourceTagNames()

tommywo:
## Summary

Restore `Scenario#getSourceTagNames()` as people are using it in their hooks.

## Motivation and Context
People are using hooks for variant settings, case insensitivity, posting information about tests execution into issue trackers using tags similar to [described in docs](https://cucumber.io/docs/cucumber/api/#link-to-other-documents). it would be nice to still have the ability to that.

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/io/cucumber/core/api/Scenario.java
+++ b/core/src/main/java/io/cucumber/core/api/Scenario.java
@@ -1,6 +1,7 @@
 package io.cucumber.core.api;
 
 import io.cucumber.core.event.Status;
+import java.util.Collection;
 import org.apiguardian.api.API;
 
 /**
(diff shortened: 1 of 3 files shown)


mpkorstanje:
As explained in Slack I am happy to add this method back in provide it has a use case that isn't covered by some other functionality such as the plug-in system or tagged hooks. So far only case insensitivity may be a potential use case but I haven't seen that use case described in any sort of detail yet.

mpkorstanje:
Found a usecase! When located in a package on the glue path hooks can be used used to provide drop in functionality that is automatically enabled. Each plugin still has to be enabled explicitly.

borbamartin:
This is great news :)

Still, I'd like to encourage everyone reading this conversation to give a second look at your code, same way as I did thanks to @mpkorstanje. There are features that you can use to achieve the same results and perhaps you're not considering how you can use them.

In my case I was able to achieve the same results of using `getSourceTagNames()` by using tagged hooks, lazy WebDriver initialization and filtering the scenarios I was feeding to TestNG-Cucumber in a custom runner.

I'm not saying this will be the same for everyone, but there's always room for improvement and this is also a chance to improve you're knowledge on the library :)

Issue #1690, opened 2019-07-08
Allow to specify different ObjectFactory classes via annotation


ralphkar:
## Summary
Specifying the ObjectFactory in CucumberOptions would be highly needed.

## Description
I have all the relevant Cucumber steps injected by a Guice injector. This injector also needs all of our project specific modules (yes, we are heavily relying on Guice). So far this was ok.
But now I have to support additional features that are only available when configured. Therefore I need to create a differently configured injector for some tests. I have been looking around for a while in order to find out what is the best way to do this in Cucumber.
I found the following (old) post on StackOverflow:
https://stackoverflow.com/questions/35315963/cucumber-with-guice-multiple-guice-injector/40429150#40429150
With this approach it would be possible for me to provide different instances of InjectorSource. Unfortunately the createRuntime()method has been removed from the Cucumber runner.

## Possible Solution
Now how could I do this with the current Cucumber (I am using version 4.3.1)? Unless there is another workaround (that I am not aware of), it would be most helpful to allow to specify an ObjectFactory within the CucumberOptions annotation.

mpkorstanje:
Currently I don't believe there is a sound solution to this problem. 

The problem is two fold. First of all the `GuiceFactory` uses properties stored in the static variable `Env.INSTANCE` to determine the name of the `InjectorSourceFactory`. It is somewhat inaccessible.

```java
    public GuiceFactory() {
        this(new InjectorSourceFactory(Env.INSTANCE).create().getInjector());
    }
```

Secondly the JavaBackend creates the object factory.

```java
    private JavaBackend(ClassFinder classFinder, TypeRegistry typeRegistry) {
        this(loadObjectFactory(classFinder, Env.INSTANCE.get(ObjectFactory.class.getName())), classFinder, typeRegistry);
    }
```

So for now you may be best forking `cucumber-junit` and hack your own needs into it. For a long term solution, I guess looking for a way to pass properties into the `GuiceFactory` rather then pulling them from a static variables would be the way to go.

mpkorstanje:
This is going to be pretty much impossible to fix in v4. But in v5 you could implement an `ObjectFactoryProviderService` with a single method `ObjectFactory create(Properties properties);` This could then be wired up in the various runners.

Would you be willing and able to send a PR?

ralphkar:
Yes, I could give it a try.
But, I just checked out cucumber-jvm:master but I can't even get it to work
in Eclipse.

Ralph

[...]

PR #1710, opened 2019-07-29
[Core] Add object-factory option to CLI and CucumberOptions

ralphkar:
## Summary

In order to provide different test configurations it might be useful to use different object factories when working with a DI framework. This is described in issue #1690.

## Details

- The annotations for JUnit and TestNG now allow to specify the class to instantiate as object factory.
- A CLI flag has been added to allow the specification of an object factory on the command line (--object-factory).
- Some minor refactorings along the way.

## Motivation and Context

- It is now possible to use different object factories in a Cucumber test suite. Features need to be separated (using tags) and can be grouped as separate tests that are annotated with different CucumberOptions.
- For each specification of CucumberOptions a different objectFactory can be specified. That way different feature tests can e.g. use differently configured injectors.
- It is no longer necessary to specify an object factory in `cucumber.properties`.
- Object factories are loaded via the ServiceLoader and need to be specified in the file `META-INF/services/io.cucumber.core.backend.ObjectFactory`. Multiple implementations can be given here.

## How Has This Been Tested?

Unit tests have been added and a production test has been executed with our product.

## Types of changes

## Checklist:


core/src/main/java/io/cucumber/core/options/RuntimeOptionsParser.java
```diff
@@ -105,6 +105,9 @@ RuntimeOptionsBuilder parse(List<String> args) {
                     throw new CucumberException("--count must be > 0");
                 }
                 parsedOptions.setCount(count);
+            } else if (arg.equals("--object-factory")) {
+                String objectFactoryClassName = args.remove(0).trim();
```
mpkorstanje: Is the trim needed?

junit/src/main/java/io/cucumber/junit/JUnitCucumberOptionsProvider.java
```diff
@@ -67,14 +68,12 @@ public boolean monochrome() {
...
         public SnippetType snippets() {
-            switch (annotation.snippets()) {
-                case UNDERSCORE:
-                    return SnippetType.UNDERSCORE;
-                case CAMELCASE:
-                    return SnippetType.CAMELCASE;
-                default:
-                    throw new IllegalArgumentException("" + annotation.snippets());
-            }
+            return annotation.snippets();
```
mpkorstanje: Why did you also change this?

core/src/test/java/io/cucumber/core/options/CucumberPropertiesParserTest.java
```diff
@@ -44,14 +43,9 @@ public boolean addClass(Class<?> glueClass) {
         }
 
         @Override
-        public void start() {
-
-        }
+        public void start() {}
```
mpkorstanje: Please keep these separated. The next auto format with IDEA will just revert it.

junit/src/main/java/io/cucumber/junit/JUnitCucumberOptionsProvider.java
mpkorstanje: Ah. I see. `io.cucumber.core.snippets.SnippetType` is still part of the public API. It shouldn't be. The local duplicates keep these sort of implementation details away from the user.

mpkorstanje: Can you instead revert this change and remove the `@API` annotation from `io.cucumber.core.snippets.SnippetType`?

mpkorstanje:
Cheers looks good. Just a few small things.

core/src/test/java/io/cucumber/core/options/CucumberPropertiesParserTest.java
ralphkar: Done

core/src/main/java/io/cucumber/core/options/RuntimeOptionsParser.java
ralphkar: Probably not ... I removed it. Just trying to be overly safe ;-)

junit/src/main/java/io/cucumber/junit/JUnitCucumberOptionsProvider.java
ralphkar: Not sure why this is better. Anyway, I reverted the changes for the JUnit and TestNG CucumberOptions annotations (the part about snippets) and removed the `@API` annotation from `io.cucumber.core.snippets.SnippetType`.

ralphkar:
I have made the requested changes. I think the PR is ready. Maybe at some point in the future someone should write a short abstract on how to use object factories in v5.

junit/src/main/java/io/cucumber/junit/JUnitCucumberOptionsProvider.java
mpkorstanje: Cheers.

Currently a user needs to declare two dependencies to user cucumber. For example `cucumber-java` and `cucumber-junit`. The transitive dependencies are automatically included by dependency management tools. 

When designing a library it is a good practice not to expose classes from transitive dependencies to the consumer.  This makes versioning easier. `SnippetType` is defined in `core`. By having an alias in `cucumber-junit` we avoid exposing core.

mpkorstanje:
Cheers!

> Maybe at some point in the future someone should write a short abstract on how to use object factories in v5.

You wouldn't be happen to volunteer? It doesn't need to be a full blown doc. A short brief in a readme of the core module will do.

ralphkar:
> > Maybe at some point in the future someone should write a short abstract on how to use object factories in v5.
> 
> You wouldn't be happen to volunteer? It doesn't need to be a full blown doc. A short brief in a readme of the core module will do.

As requested -> #1715

PR #1726, opened 2019-08-03
[All] Replace ExpectedException with assertThrows

nhojpatrick:
## Summary

assertThrows replacing ExpectedException

## Details

Only the quick wins done, some outstanding to be convert in future PR

## How Has This Been Tested?

mvn clean install passed

## Types of changes

## Checklist:

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -68,6 +68,12 @@
         </dependency>
 
         <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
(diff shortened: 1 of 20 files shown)


core/pom.xml
```diff
@@ -67,6 +67,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
```
mpkorstanje: This should probebely be the junit 5 api. But we can fix that later.

PR #1727, opened 2019-08-03
[All] Use Junit 5 API

nhojpatrick:
## Summary

Next step, org.junit.Test to org.junit.jupiter.api.Test

## Details

Easy changes, not touch cucumber-junit because unsure if that should be changed? e.g. should we have cucumber-junit4 and cucumber-junit-jupiter... so i've left cucumber-junit as org.junit.Test

## How Has This Been Tested?

mvn clean install

## Types of changes

## Checklist:


mpkorstanje:
While `cucumber-junit` can be tested with Junit 5 it might be worth while to have a test implementation that uses JUnit 4 rather then the Vintage implementation of Junit 5. So let's leave `cucumber-junit` entirely on Junit 4.

Issue #1718, opened 2019-07-31
Use JUnit 5 internally


mpkorstanje:
While Cucumber does not support JUnit5 yet. We should at least use it internally.

 * [x] We can use `junit-vintage` to keep the `@RunWith(Cucumber.class)` annnotation working 
 * [x] We should replace all `ExpectedException` rules with `Assertions.assertThrows`.
 * [x] We should remove all other `@RunWith` and `@Rules` annotations.
 * [x] Use the right dependencies in the poms
 * [x] There is a test dependency not marked with scope test (fix it!).
 * [x] Get rid of AssertJ

nhojpatrick:
@mpkorstanje I don't mind picking this up and leading it. I assume develop-v5 to branch from.

mpkorstanje:
Please do! You can create pull requests directly against `master`. I've merged the `develop-v5` branch into master a few days ago.

nhojpatrick:
Going to do them as quicker seperate branches building off each other so smaller and faster to get back into master. Everything that uses @Rule ExpectedException which converted quickly has been done.

nhojpatrick:
I'm making multiple branches, each off the last pull request, to try and make each change easier to check. If you just want to check it all in one go, pick the highest pr number associated with 1718 and that will be the last one.

mpkorstanje:
Cheers! I've merged yesterdays batch. Please beware that I'm squashing all commits to keep the commit history readable. You may have to rebase but you can resolve all conflicts using your version.

nhojpatrick:
Only "@RunWith(Cucumber.class)" is left

mpkorstanje:
Nice!

PR #1733, opened 2019-08-04
[All] Use JUnit 5 internally

nhojpatrick:
## Summary

Move fixes and tweaks for junit v5 #1718

## Details

Null checking values being passed into Exception methods

## Types of changes

## Checklist:


core/src/main/java/io/cucumber/core/exception/CompositeCucumberException.java
```diff
@@ -1,25 +1,32 @@
...
     private final List<Throwable> causes;
 
-    public CompositeCucumberException(List<Throwable> causes) {
-        super(String.format("There were %d exceptions:", causes.size()));
-        this.causes = causes;
+    public CompositeCucumberException(final List<Throwable> causes) {
+        super(String.format("There were %d exceptions:",
+            Objects.isNull(causes) ? 0 : causes.size()
+        ));
+        this.causes = Objects.isNull(causes) ? new ArrayList<>() : causes;
```
mpkorstanje: This exception shouldn't be used in scenarios where there is only one exception. Throwing an `IllegalArgument` exception when an empty is provided would be better.

It is also preferable to use `== null` for regular null checks. `Objects::isNull` is intended for use in lambda expressions.

core/src/test/java/io/cucumber/core/exception/CucumberExceptionTest.java
```diff
@@ -0,0 +1,94 @@
...
+import static org.hamcrest.core.Is.isA;
+import static org.hamcrest.core.IsEqual.equalTo;
+import static org.hamcrest.core.IsNull.nullValue;
+import static org.junit.jupiter.api.Assertions.assertAll;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class CucumberExceptionTest {
+
+    @TestFactory
+    public Collection<DynamicTest> exceptions() {
```
mpkorstanje: Are these tests really dynamic though? It looks like a regular list of tests.

core/src/test/java/io/cucumber/core/plugin/URLOutputStreamTest.java
```diff
@@ -65,15 +64,15 @@ private static URL createUrl(String s) {
     private final List<File> tmpFiles = new ArrayList<File>();
     private final List<String> threadErrors = new ArrayList<String>();
 
-    @Rule
-    public TemporaryFolder tempFolder = new TemporaryFolder();
+    @TempDir
+    Path tempDir;
```
mpkorstanje: Cool!

java/src/main/java/io/cucumber/java/InvalidMethodSignatureException.java
```diff
@@ -6,14 +6,19 @@
...
-    private InvalidMethodSignatureException(String message) {
+    private InvalidMethodSignatureException(final String message) {
         super(message);
     }
 
-    static InvalidMethodSignatureExceptionBuilder builder(Method method) {
+    static InvalidMethodSignatureExceptionBuilder builder(final Method method) {
+        if (Objects.isNull(method)) {
+            throw new IllegalArgumentException("Supplied Method can't be null for InvalidMethodSignatureException");
+        }
```
mpkorstanje: The standard java pattern hand for this is `Objects.requireNonNull` with a static import..

 No need to pass in a message as the line should make it obvious which value was null.

java/src/main/java/io/cucumber/java/InvalidMethodException.java
```diff
@@ -3,17 +3,24 @@
 import io.cucumber.core.exception.CucumberException;
 
 import java.lang.reflect.Method;
+import java.util.Objects;
 
-final class InvalidMethodException extends CucumberException {
+final class InvalidMethodException
+    extends CucumberException {
```
mpkorstanje: There seems to be no need for a line break here.

java/src/main/java/io/cucumber/java/InvalidMethodSignatureException.java
```diff
@@ -54,34 +58,35 @@ public InvalidMethodSignatureException build() {
         }
 
         private String describeNote() {
-            return String.join("\n", notes);
+            return String.join("\n", this.notes);
```
mpkorstanje: With the exception of constructors, inner classes and setters the `this` key word isn't usually needed.

java/src/main/java/io/cucumber/java/InvalidMethodSignatureException.java
mpkorstanje: I guess I should show this as an example:

```java
    public EmbedEvent(Instant timeInstant, TestCase testCase, byte[] data, String mimeType) {
        super(timeInstant, testCase);
        this.data = Objects.requireNonNull(data);
        this.mimeType = Objects.requireNonNull(mimeType);
        this.name = null;
    }
```

The pass through is really use full.

The exception to the no message thing this is when the message involves user input. You can find some examples of this in the `FeaturePath` 

```java
    public static URI parse(String featureIdentifier) {
        requireNonNull(featureIdentifier, "featureIdentifier may not be null");
        if(featureIdentifier.isEmpty()){
            throw new IllegalArgumentException("featureIdentifier may not be empty");
        }
```
Though I must admit it is not consistently applied.

java/src/main/java/io/cucumber/java/InvalidMethodException.java
nhojpatrick: habbit, do it without thinking, will change back for cucumber stuff

core/src/main/java/io/cucumber/core/exception/CompositeCucumberException.java
```diff
@@ -1,25 +1,32 @@
...
 
     public List<Throwable> getCauses() {
         return Collections.unmodifiableList(this.causes);
     }
 
     public String getMessage() {
-        StringBuilder sb = new StringBuilder(super.getMessage());
-        for (Throwable e : this.causes) {
+        final StringBuilder sb = new StringBuilder(super.getMessage());
+        for (final Throwable e : this.causes) {
```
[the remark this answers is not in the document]
nhojpatrick: kept having issues with people reusing variables, and/or code merges, and due to variables being final the compiler errors instead of waiting for unit tests, or cucumber testing or manually regression testing so it's auto pilot for me. again will not add in cucumber code base

java/src/main/java/io/cucumber/java/InvalidMethodSignatureException.java
nhojpatrick: yep, like my other comments, due to previous projects it's easier to do this as having had bugs introduced by senior/experience developers who refactored code and created local variables with the same name as member variables so code then breaks, won't change existing code without a need to have this.

nhojpatrick: reverted back

java/src/main/java/io/cucumber/java/InvalidMethodSignatureException.java
nhojpatrick: will do tomorrow

core/src/main/java/io/cucumber/core/exception/CompositeCucumberException.java
nhojpatrick: will do tomorrow

timtebeek:
@nhojpatrick Fixed some import conflicts in RuntimeTest; just so you're aware and pull before you continue. :)

mpkorstanje:
I've fixed a few a dynamic tests and removed the others. It gets a bit tedious fixing them all and I'd like to get the improvements merged.

It's also good to note that most exceptions are only used in specific scenarios. Too many instances and duplicate step definition exceptions are never used for 0 or 1 elements. So adding tests for this is a bit over kill.

PR #1734, opened 2019-08-09
[Core] Fix UnusedStepSummaryPrinter and consistently handle plugin events

timtebeek:
## Summary

Consistently handle events in plugins, after the initial push towards using lambdas where possible.
Also fixes an issue with the UnusedStepSummaryPrinter which failed to take into account that steps are registered at each scenario invocation. Fixed and improved test to verify.

## Details

Replace EventHandler instances with method references and apply consistent naming after the event handled.

## Motivation and Context

Consistency.

## How Has This Been Tested?

JUnit test suite.

## Types of changes

- [x] Refactoring (non-breaking change which fixes an issue).


mpkorstanje:
Looks good! Cheers!

If you want to merge it please use the squash merge option (from the drop down). You can use the summary section from this issue as the commit message.

PR #1736, opened 2019-08-09
[Core] Replace PicklePredicate with Predicate<PickleEvent>

### LINKED ISSUE #1646: User Streams to filter pickles in Runtime
After #1645 is merged `develop-v5` should refactor the `Runtime` implementation and builder to use streams. The builder should accept a list of `Predicates` which are used to filter the pickles. The existing `Filters` class should be used as one of these predicates.

timtebeek:
Fix #1646 ; Work in progress. Original ticket could use some more details.


timtebeek:
@mpkorstanje Not sure how far to take this conversion; can you provide some guidance/clarification?

junit/src/main/java/io/cucumber/junit/FeatureRunner.java
```diff
@@ -78,23 +77,21 @@ protected void runChild(PickleRunner child, RunNotifier notifier) {
...
+                } else {
+                    PickleRunner picklePickleRunner;
+                    picklePickleRunner = withNoStepDescriptions(cucumberFeature.getName(), runnerSupplier, pickleEvent, jUnitOptions);
+                    children.add(picklePickleRunner);
                 }
+            } catch (InitializationError e) {
+                throw new CucumberException("Failed to create scenario runner", e);
             }
-        }
+        });
```
mpkorstanje: Changing the if block so that it becomes a map step would allow this to be rewritten to a function that returns the `children` list instead of adding into it. Then in the constructor this list can be assigned. Considering the arguments involved I would even considering in lining this in the constructor.

testng/src/main/java/io/cucumber/testng/TestNGCucumberRunner.java
```diff
@@ -125,7 +125,7 @@ public void finish() {
             List<CucumberFeature> features = getFeatures();
             for (CucumberFeature feature : features) {
                 for (PickleEvent pickle : feature.getPickles()) {
-                    if (filters.matchesFilters(pickle)) {
+                    if (filters.test(pickle)) {
                         scenarios.add(new Object[]{new PickleEventWrapperImpl(pickle),
                             new CucumberFeatureWrapperImpl(feature)});
                     }
```
mpkorstanje: This could be entirely done as a stream.

core/src/main/java/io/cucumber/core/filter/Filters.java
```diff
@@ -3,49 +3,45 @@
...
-    	if (count > pickleEvents.size() || count < 1) {
-    		return pickleEvents;
-    	}
-		return pickleEvents.subList(0, count);
-	}
+        if (count > pickleEvents.size() || count < 1) {
+            return pickleEvents;
+        }
+        return pickleEvents.subList(0, count);
+    }
```
mpkorstanje: Ideally where ever `Filters` is used we now use `Predicate<PickleEvent> `.  So might be worth extracting `limitPickleEvents` to some other class.

timtebeek:
@mpkorstanje Think this MR already brings the code closer to your vision; Feel free to make further changes where needed. Tried for a while to test the new stream block in Runtime, but that proved difficult due to the amount of setup required, and lack of assertion options aside from listening to events. Since it's declarative and likely covered through other paths as well, can we leave it at this for now?

core/src/main/java/io/cucumber/core/runtime/Runtime.java
```diff
@@ -87,22 +93,14 @@ public void run() {
...
-            executingPickles.add(executor.submit(() -> runnerSupplier.get().runPickle(pickleEvent)));
-        }
+        final List<Future<?>> executingPickles = features.stream()
+                .flatMap(feature -> feature.getPickles().stream())
+                .filter(filter)
+                .collect(Collectors.collectingAndThen(Collectors.toList(),
+                        list -> pickleOrder.orderPickleEvents(list).stream()))
+                .limit(limit > 0 ? limit : Integer.MAX_VALUE)
+                .map(pickleEvent -> executor.submit(() -> runnerSupplier.get().runPickle(pickleEvent)))
+                .collect(Collectors.toList());
```
mpkorstanje: `collectingAndThen` neat!

timtebeek: Yes that one was new to me as well! Needed it since the random sort order with seed couldn't use a `Comparator`, so learned a new trick. :)

mpkorstanje: Using comparator wouldn't have worked anyway. Java will complain if the comparator isn't consistent.

junit/src/main/java/io/cucumber/junit/FeatureRunner.java
```diff
@@ -20,26 +20,31 @@
...
-                }
-            }
-            return withNoStepDescriptions(cucumberFeature.getName(), runnerSupplier, pickleEvent, jUnitOptions);
-        }).collect(Collectors.toList());
+        this.cucumberFeature = feature;
+        this.children = feature.getPickles().stream()
+            .filter(filter).
+                map(pickleEvent -> options.stepNotifications()
+                    ? withStepDescriptions(runners, pickleEvent, options)
+                    : withNoStepDescriptions(feature.getName(), runners, pickleEvent, options))
```
timtebeek: Good call getting the try/catch blocks out of the stream operations; wasn't sure how much I could change to make it less ugly before. :+1:

mpkorstanje: Generally speaking if code shifts around inside a class it is never a problem.

PR #1739, opened 2019-08-15
[Core] Upgrade to `datatable:3.0.0`

mpkorstanje:
## Summary

 Upgrade to `datatable:3.0.0`. Changes to value type from `Class<?>` to `Type` and return type to `Object` in `TableEntryByTypeTransformer` and `TableCellByTypeTransformer`. This allows lambdas to be used to define these transformers.

## Types of changes

## Checklist:


examples/java-calculator/src/test/java/io/cucumber/examples/java/ParameterTypes.java
```diff
@@ -1,6 +1,6 @@
 package io.cucumber.examples.java;
 
-import io.cucumber.datatable.dependency.com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectMapper;
```
aslakhellesoy: What's the rationale behind changing from the shaded class?

pom.xml
```diff
@@ -44,7 +44,7 @@
         <!--Dependencies -->
         <gherkin.version>5.1.0</gherkin.version>
         <cucumber-expressions.version>8.0.0</cucumber-expressions.version>
-        <datatable.version>2.0.0</datatable.version>
+        <datatable.version>3.0.0-SNAPSHOT</datatable.version>
```
aslakhellesoy: I see the build is failing because there isn't a 3.0.0-SNAPSHOT published.

You should be able to do this easily with `make deploy` from `datatable/java`

examples/java-calculator/src/test/java/io/cucumber/examples/java/ParameterTypes.java
mpkorstanje: It'll be removed with https://github.com/cucumber/cucumber/pull/682.

This makes Jackson a complete drop in that can be managed by the user. Does however also require us to include it in the examples.

pom.xml
aslakhellesoy: I just did a `make deploy` from the `reimplement-type-factory` branch, which I'm guessing is the only branch targetting `3.0.0-SNAPSHOT`.

It's a little tedious not to have snapshots continuously deployed. OTOH it's a bit more secure, and it's relatively easy to do a manual snapshot release when needed.

Then I triggered a new build, which [now passes](https://travis-ci.org/cucumber/cucumber-jvm/builds/572511667) 💚 🥒

mpkorstanje: Cheers!

Issue #1596, opened 2019-04-02
Cannot pass tag expression to maven-surefire-plugin


Mike-Now:
## Summary

The new multi-token expression syntax makes it impossible to declare multi token tag expression in the --tags flag inside the -Dcucumber.options flag in maven-surefire-plugin configuration.

## Expected Behavior

It should be possible possible to pass multi token tag expression in the -Dcucumber.options flag.

## Current Behavior

maven-surefire-plugin crashes throwing either VM crash, IllegalArgumentException or doesn't run any tests depending on the input.

## Possible Solution

Make it possible to make single token tag expression without any whitespace characters.

## Steps to Reproduce (for bugs)

1. Create a basic cucumber project using maven with maven-surefire-plugin
2. In the pom.xml define argLine tag inside maven-surefire-plugin configuration
3. Inside argLine tag add -Dcucumber.options=" --tags 'not @ignored'" or -Dcucumber.options=' --tags "not @ignored"'
4. run `mvn test`

## Context & Motivation

My use case is injecting tag expression from parent pom.

## Your Environment

* Version used: 3.0.2
* Operating System and version: Linux Mint 19

mpkorstanje:
Try 
Try:
```
<argLine>-Dcucumber.options="--tags 'not @ignored'"</argLine>
```

Mike-Now:
I entered that flag slopily in the description.
I have already tried your suggestion, but it doesn't work. If you don't believe me, I can make a minimal example project.

mpkorstanje:
Please do. I copied that out of an existing project.

Mike-Now:
Bad call - I cannot reproduce it in a minimal example. I must a have problem in a configuration then. 
Thank you very much.

mpkorstanje:
Any chance you were using multiple tags?

Mike-Now:
Just a negation.

mpkorstanje:
You probably have to report this to surefire. The command line is valid.

Mike-Now:
I made it work for maven-surefire-plugin:2.22.1 using verbatim command from the 2.12.4 version. If anybody wonders, here's the solution:

```<argLine>-Dcucumber.options="--tags '\''not @ignored'\''"```

mpkorstanje:
Perhaps we should also consider making tags work without explict `@` symbols.

mpkorstanje:
[...]

PR #1741, opened 2019-08-18
[Core] Add property based variant for each runtime option

### LINKED ISSUE #1675: Provide a property name variant for each runtime option
## Summary

It is possible to pass properties to cucumber using CLI arguments in a property. E.g.

```
mvn clean test -Dcucumber.options="--strict --monochrome"
``` 
This is rather complicate, esp when multiple layers are involved and qoutes get confusing (e.x: https://github.com/cucumber/cucumber-jvm/issues/1596). 

```
mvn clean test -Dcucumber.options='--strict --monochrome --tags "not @ignored"'
``` 

A better way to do this would be to provide each option individually:

```
mvn clean test -Dcucumber.strict=true -Dcucumber.monochrome=true -Dcucumber.tags="not @ignored"
```

Once this is available we deprecate `cucumber.options`.

## Possible Solution

In `develop-v5` the `CucumberPropertiesParser.parse` method is provided with a map containing  properties from either either a `cucumber.properties` file, environment variables, or system properties.

After parsing `cucumber.options` an options builder is created. When a property is present the builder can be used to set the matching option to the right value.

Supported properties:
```
cucumber.glue=<separated list of paths>  #using OS appropriate seperator
cucumber.features=<separated list of paths>   #using OS appropriate seperator
cucumber.tags=<single tag expression>
cucumber.name=<regex>
cucumber.ansi-colors.disabled=<boolean>
cucumber.execution.dry-run=<boolean>
cucumber.execution.strict=<boolean>
cucumber.execution.parallel.config.fixed.parallelism=<int>
cucumber.execution.limit=<positive integer>
cucumber.execution.order=<reverse|random|random:seed>
cucumber.snippet-type=<underscore|camelcase>
cucumber.wip=<boolean>
```

Property names where possible should match those used by JUnit 5: https://github.com/cucumber/cucumber-jvm/blob/junit-jupiter/junit-platform-engine/src/main/java/io/cucucumber/jupiter/engine/Constants.java

To document this properly with Java doc please take hints from:

https://junit.org/junit5/docs/current/api/constant-values.html
https://junit.org/junit5/docs/current/api/org/junit/jupiter/engine/Constants.html

mpkorstanje:
## Summary

Adds properties for all CLI options. Removes the ability to load rerun files from the class-path. There appears to be no use case and it significantly simplifies the implementation.

## Details

Adds:

```
cucumber.ansi-colors.disabled
cucumber.execution.dry-run
cucumber.execution.limit
cucumber.execution.order
cucumber.execution.parallel.config.fixed.parallelism
cucumber.execution.strict
cucumber.execution.wip
cucumber.feature
cucumber.rerun-file
cucumber.filter.name
cucumber.filter.tags
cucumber.glue
cucumber.object-factory
cucumber.plugin
cucumber.snippet-type
```

## Motivation and Context

It is possible to pass properties to cucumber using CLI arguments in a property. E.g.

```
mvn clean test -Dcucumber.options="--strict --monochrome"
``` 
This is rather complicate, esp when multiple layers are involved and quotes get confusing (e.x: https://github.com/cucumber/cucumber-jvm/issues/1596). 

```
mvn clean test -Dcucumber.options='--strict --monochrome --tags "not @ignored"'
``` 

A better way to do this would be to provide each option individually:

```
mvn clean test -Dcucumber.strict=true -Dcucumber.ansi-colors.disabled=true -Dcucumber.filter.tags="not @ignored"
```

Closes #1675

## Types of changes

## Checklist:


mpkorstanje:
The `rerun-file` and `glue-file` options don't work well at all. Properties are single values. They don't work well at all with maps and the semantics of combining them are messy.

Best to leave them out for now and revisit them once the JUnit 5 implementation comes along.

mpkorstanje:
Looks like the common practice is to use comma separated lists and use strict overriding. Which is exactly what we do right now.

A complicating factor is that JUnit and TestNG will always override `cucumber.glue` and `cucumber.features` with the package of the runner. But from a hierarchical perspective that makes sense.

salunkhe-ravi:
For some reason "cucumber.execution.limit=1" doesn't work for me...
Can anyone share me some examples and/or usage documentation please?

mpkorstanje:
Not all runners support all options. Limit is only supported by the CLI.

Issue #1705, opened 2019-07-21
Use dedicated DocStringTypeRegistry for doc string transforms


mpkorstanje:
## Summary

Cucumber uses the `DatatTableTypeRegistry` to convert doc strings. This is rather confusing and ignores the type provided with the doc string.

## Current Behavior

Given some gherkin

```feature
Given a doc string
"""json
{hello:"world"}
"""
```

And a step definition:

```java
@Given("a doc string")
public void a_doc_string(JsonNode json){

}
```

Cucumber will try to transform this string into JSON by turning the docstring into a 1x1 table and using the `DataTableTypeRegistry` to convert it to an object of the right target type.

```java
DocStringTransformer<?> docStringTransform = new DocStringTransformer<Object>() {
    @Override
    public Object transform(String docString) {
        Type targetType = tableOrDocStringType.resolve();
        if (Object.class.equals(targetType)) {
            return docString;
        }

        List<List<String>> raw = singletonList(singletonList(docString));
        return DataTable.create(raw, StepExpressionFactory.this.tableConverter).convert(targetType, transpose);
    }
};
```

## Expected Behavior

Cucumber uses a `DocStringTypeRegistry` in which transformations for each document type can be registered or where a default doc string transformer (document, type) -> Object can be registered.

PR #1745, opened 2019-08-19
[Core] Add dedicated DocStringTypeRegistry

rasklaad:
## Summary

Moved docstring transformation from 1x1 DataTable to separate class

## Motivation and Context

https://github.com/cucumber/cucumber-jvm/issues/1705

## Types of changes

## Checklist:


core/src/main/java/io/cucumber/core/stepexpression/DocStringType.java
```diff
@@ -0,0 +1,44 @@
+package io.cucumber.core.stepexpression;
+
+
+import java.lang.reflect.Type;
+
+public class DocStringType {
```
mpkorstanje:
```suggestion
@API(status = API.Status.STABLE)
public final class DocStringType {
```

And you may have to fix the import manually.

core/src/main/java/io/cucumber/core/stepexpression/DocStringType.java
```diff
@@ -0,0 +1,44 @@
...
+    private final Type targetType;
+    private final String contentType;
+    private final Converter converter;
+
+    public Converter getConverter() {
```
mpkorstanje:
```suggestion
    Converter getConverter() {
```

core/src/main/java/io/cucumber/core/stepexpression/DocStringType.java
```diff
@@ -0,0 +1,44 @@
...
+        return converter;
+    }
+
+    public String getContentType() {
```
mpkorstanje:
```suggestion
    String getContentType() {
```

core/src/main/java/io/cucumber/core/stepexpression/DocStringType.java
```diff
@@ -0,0 +1,44 @@
...
+        return contentType;
+    }
+
+    public Type getTargetType() {
```
mpkorstanje:
```suggestion
    Type getTargetType() {
```

core/src/main/java/io/cucumber/core/stepexpression/DocStringType.java
```diff
@@ -0,0 +1,44 @@
...
+        if (contentType == null) {
+            this.contentType = "";
+        }
+        else {
+            this.contentType = contentType;
+        }
+        this.converter = converter;
+    }
+
+    public interface Converter<T> {
```
mpkorstanje:
```suggestion
    @FunctionalInterface
    public interface Converter<T> {
```

rasklaad:
Hi @mpkorstanje , thank you for the review.
I almost fixed everything you pointed, and I have a question how DocStringConverter should behave in this situation:
Feature:
```
Given doc string step
   """my-content-type
   really long docstring
   """
```
Step def:
```
@Given("doc string step")
public void step(String docString) {
    System.out.println(docString);
}
```
I have a two options:
1. DocStringConverter throw exception, because `my-content-type` is not registered.
 This will probably broke a lot of the user code, if user use content type without registering custom converter.
2. DocStringConverter ignore content-type, because target type is String.
This will not allow user register something like this:
```
registry.defineDocStringType(new DocStringType(String.class, "my-content-type", (String s) -> s.replaceAll("\\s", "-")))
```

mpkorstanje:
I think that by searching for possible converter in order of priority of name, then class, should solve that problem.

```
If a doc string type is provided in the gherkin lookup a converter for that doc string type
   if a converter for that doc string type could be found
        convert the value using the the found converter
        return the converted value

if a converter for the type could be found
  convert the value using the the found converter
  return the converted value

throw exception no converter could be found by doc string type, type and no default converter was available.
```

core/src/test/java/io/cucumber/core/stepexpression/StepExpressionFactoryTest.java
```diff
@@ -103,13 +104,39 @@ void unknown_target_type_does_no_transform_data_table() {
...
+        StepExpression expression = new StepExpressionFactory(registry).createExpression("Given some stuff:", JsonNode.class);
+        List<Argument> match = expression.match("Given some stuff:", docString, contentType);
+        JsonNode node = (JsonNode) match.get(0).getValue();
+        assertThat(node.asText(), equalTo(docString));
+    }
+
+    @Test
+    void docstring_expression_transform_anonymous_doc_string_type() {
+        String docString = "A rather long and boring string of documentation";
+        String contentType = null;
```
mpkorstanje: This should result in an error. The content type must always be a string. And the string may not be empty. With the exception of the nameless doc string type we predefined.

mpkorstanje:
Added a few bits to make all these variations work:

```java
    @Given("a doc string 1")
    public void docstr(Object d) {

    }

    @Given("a doc string 2")
    public void docstr(DocString d) {
        d.convert(JsonNode.class);
    }

    @Given("a doc string 3")
    public void docstr(String d) {

    }

    @Given("a doc string 4")
    public void docstr(JsonNode d) {

    }
```

mpkorstanje:
It's currently possible to define parameter types by annotation:

```
public class Steps {

    @Given("today is {iso8601Date}")
    public void today_is(LocalDate date) {

    }
}
```

We should make the same possible for doc strings. I've added the annotation but it's not added to the glue yet. I'm out of time for now. Feel free to pick that up if you have some time.

rasklaad:
> It's currently possible to define data table types by annotation:
> 
> ```
>    @DataTableType
>     public Author authorEntryTransformer(Map<String, String> entry) {
>         return new Author(
>             entry.get("firstName"),
>             entry.get("lastName"),
>             entry.get("birthDate"));
>     }
>     
> ```
> 
> We should make the same possible for doc strings. I've added the annotation but it's not added to the glue yet. I'm out of time for now. Feel free to pick that up if you have some time.

Sure, I'll do it, but it will take some time, to figure out how @DataTableType works.

mpkorstanje:
Good enough I think. Cheers!

Issue #1746, opened 2019-08-20
Support title case data table headers


mpkorstanje:
## Summary

When [using Jackson as a default data table entry transformer](https://cucumber.ghost.io/blog/announcing-cucumber-jvm-4-2-0/) it is possible to automatically convert data tables to lists of specific objects. Doing so however requires that headers are written in camel case. They must match the property names of the object.

## Expected Behavior

When using Jacksons `ObjectMapper` as a  default table entry converter then some data table with title case headers should be convertible to a list of authors. 

```gherkin
| First Name | Last Name | Day of Birth |
| Astrid     | Lindgren  |  1907-12-14  | 
```

```java
public class Author {
  public String firstName;
  public String lastName;
  public LocalDate dayOfBirth;
}
```

## Current Behavior

Jackson will complain about 

```
Caused by: java.lang.IllegalArgumentException: Unrecognized field "First Name" (class io.cucumber.skeleton.Stepdefs$Author), not marked as ignorable (2 known properties: "previousDayOfWeek", "a"])
 at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: io.cucumber.skeleton.Stepdefs$Bean["Previous Day Of Week"])
```

## Workaround 1

Configure Jackson with a space case naming strategy.

A limitation of this strategy is that it converts property names to data table names.
This means that for a given property name there is exactly one correct data table
way of writing a data table header. Converting in the other direction like the cucumber
1/2.x implementation did, made writing headers in title case possible e.g:

Jackson:

 - Greetings From Earth --> greetingsFromEarth

Cucumber:

 - Greetings From Earth --> greetingsFromEarth
 - Greetings from Earth --> greetingsFromEarth

```java
package io.cucumber.skeleton;

import cucumber.api.TypeRegistry;
import cucumber.api.TypeRegistryConfigurer;
import io.cucumber.datatable.TableCellByTypeTransformer;
import io.cucumber.datatable.TableEntryByTypeTransformer;
import com.fasterxml.jackson.databind.ObjectMapper;
import com.fasterxml.jackson.databind.PropertyNamingStrategy;

import java.util.Locale;
import java.util.Map;

import static java.lang.Character.toUpperCase;
import static java.util.Arrays.stream;
import static java.util.stream.Collectors.joining;

public class TypeConfiguration implements TypeRegistryConfigurer {
    @Override
    public Locale locale() {
        return Locale.ENGLISH;
    }

[...]

PR #1751, opened 2019-08-26
[Java] Convert title case headers to property names

rasklaad:
## Summary
User can convert data table title case header, using default data table transformer

## Motivation and Context
https://github.com/cucumber/cucumber-jvm/issues/1746
## Types of changes

## Checklist:


java/src/main/java/io/cucumber/java/DefaultDataTableEntryTransformer.java
```diff
@@ -25,5 +25,5 @@
 @Target(ElementType.METHOD)
 @API(status = API.Status.STABLE)
 public @interface DefaultDataTableEntryTransformer {
-
+    boolean titleCaseHeaders() default false;
```
mpkorstanje: As this will go into the v5 release and as `@DefaultDataTableEntryTransformer` will be introduced in the same realse we can set this to true and provide this as a new feature by default. That way we'll require minimal amounts of configuration to get a working system.

I'm not sure about the name. We're not creating title case headers. We're turning headers into property names. Maybe `convertHeadersToProperties`?  I don't have a good suggestion right now though.

java/src/main/java/io/cucumber/java/JavaDefaultDataTableCapitalCaseTransformerDefinition.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Override
+    Object execute(Map<String, String> fromValue, Type toValueType, TableCellByTypeTransformer cellTransformer) throws Throwable {
+        Object[] args;
+        if (method.getParameterTypes().length == 3) {
+            args = new Object[]{fromValue, toValueType, cellTransformer};
+        } else {
+            args = new Object[]{fromValue, toValueType};
+        }
+        args[0] = convertToCamelCase((Map<String, Object>)args[0]);
```
mpkorstanje: I suggested earlier to merge these classes. But had that not been the case you could reduce the duplication here a bit as follows:

You've duplicated this method to modify `arg[0]`. However `arg[0]` will always be `fromValue`. So you can modify `fromValue` and then called `super.execute` with the modified `fromValue`.

java/src/main/java/io/cucumber/java/JavaDefaultDataTableCapitalCaseTransformerDefinition.java
```diff
@@ -0,0 +1,77 @@
...
+        return Invoker.invoke(lookup.getInstance(method.getDeclaringClass()), method, args);
+    }
+
+    private Map<String, Object> convertToCamelCase(Map<String, Object> map) {
+        Map<String, Object> newMap = new HashMap<>();
+        CamelCaseStringConverter converter = new CamelCaseStringConverter();
+        for (Map.Entry<String, Object> entry : map.entrySet()) {
+            newMap.put(converter.map(entry.getKey()), entry.getValue());
```
mpkorstanje: What will happen when I have a table that looks like this?

```
| Example Header | example header | example Header |
|  2             |  3             |  4             | 
```

java/src/main/java/io/cucumber/java/GlueAdaptor.java
```diff
@@ -54,7 +54,12 @@ void addDefinition(Method method, Annotation annotation) {
         } else if (annotation.annotationType().equals(DefaultParameterTransformer.class)) {
             glue.addDefaultParameterTransformer(new JavaDefaultParameterTransformerDefinition(method, lookup));
         } else if (annotation.annotationType().equals(DefaultDataTableEntryTransformer.class)) {
-            glue.addDefaultDataTableEntryTransformer(new JavaDefaultDataTableEntryTransformerDefinition(method, lookup));
+            if (method.getAnnotation(DefaultDataTableEntryTransformer.class).titleCaseHeaders()) {
```
mpkorstanje: Rather then introducing a new class it might be better to add a constructor parameter to the existing `JavaDefaultDataTableEntryTransformerDefinition`. It avoids the need to create a subclass and duplicate all the tests for the original class and the sub class.

java/src/main/java/io/cucumber/java/JavaDefaultDataTableCapitalCaseTransformerDefinition.java
```diff
@@ -0,0 +1,77 @@
...
+            }
+            return sb.toString();
+        }
+
+        private String normalizeSpace(String originalHeaderName) {
+            return WHITESPACE_PATTERN.matcher(originalHeaderName.trim()).replaceAll(WHITESPACE);
+        }
+
+        private String capitalize(String string) {
+            return new StringBuilder(string.length()).append(Character.toTitleCase(string.charAt(0))).append(string.substring(1)).toString();
```
mpkorstanje: Some new lines might be in order to keep the formatting readable.

java/src/main/java/io/cucumber/java/JavaDefaultDataTableCapitalCaseTransformerDefinition.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.regex.Pattern;
+
+public class JavaDefaultDataTableCapitalCaseTransformerDefinition extends JavaDefaultDataTableEntryTransformerDefinition {
+    JavaDefaultDataTableCapitalCaseTransformerDefinition(Method method, Lookup lookup) {
+        super(method, lookup);
+    }
+
+    @Override
+    public TableEntryByTypeTransformer tableEntryByTypeTransformer() {
+        return super.tableEntryByTypeTransformer();
```
mpkorstanje: Because you're only calling `super` there is no need to override this method.

java/src/test/java/io/cucumber/java/defaultstransformer/DataTableSteps.java
```diff
@@ -31,6 +31,11 @@ public void aListOfAuthorsInATable(List<Author> authors) {
         assertTrue(authors.contains(expectedAuthor));
     }
 
+    @Given("a table with capital case headers")
+    public void aTableWithCapitalCaseHeaders(List<Author> authors) {
+        assertTrue(authors.contains(expectedAuthor));
```
mpkorstanje: I think you meant 'title case`.

java/src/main/java/io/cucumber/java/JavaDefaultDataTableCapitalCaseTransformerDefinition.java
mpkorstanje: Given the complexity of this function and given that it doesn't have any tests yet and given that in its current state is rather hard to test you may want to consider reorganizing this method and and the `CamelCaseStringConverter` so you have a single test able unit.

java/src/test/java/io/cucumber/java/JavaDefaultDataTableEntryTransformerDefinitionTest.java
```diff
@@ -131,6 +136,33 @@ void method_must_have_cell_transformer_as_optional_third_argument() throws Throw
...
+        String actualHeader = toCamelCase(header);
+        Assert.assertThat(actualHeader, equalTo(expectedHeader));
+    }
+
+    private String toCamelCase(String header) throws Throwable{
+        Constructor constructor = JavaDefaultDataTableEntryTransformerDefinition.class.getDeclaredClasses()[0].getDeclaredConstructor();
+        constructor.setAccessible(true);
+        Object camelCaseStringConverter = constructor.newInstance();
+        Method method = camelCaseStringConverter.getClass().getDeclaredMethod("toCamelCase", String.class);
+        return (String) method.invoke(camelCaseStringConverter, header);
```
mpkorstanje: I don't think is necessary to use violence to test the `CamelCaseStringConverter`.

You may want to consider moving `convertToCamelCase` into the `CamelCaseStringConverter` and making the class a package private member of the `io.cucumber.java` package. That way you can focus on testing the title case to camel case edge cases withouth having to construct the `JavaDefaultDataTableEntryTransformerDefinition.`

mpkorstanje:
Added an exception to deal with collisions when converting headers and moved the camel case conversion to core so the different backends don't have to deal with individually.

mpkorstanje:
And with that I think we're done. @rasklaad is there anything you'd think we should add?

rasklaad:
Nah, I think we are fine

mpkorstanje:
Cool. Cheers. Much appreciated!

gurpiarbassi:
@mpkorstanje I've been trying to piece together what needs to be done in order to make use of this lovely functionality. I've got a table with columns in title case e.g. "Order Id", "Customer Id"

I've got a object like:
```
Order {
 String orderId;
 String customerId;
}
```
I've got a step definition

which takes `Order` as an argument.

Do I need to register default table transformer or should it work out the box?

mpkorstanje:
> Do I need to register default table transformer

Yes. Using the annotation.

gurpiarbassi:
sorry, may sound like a stupid question but where do you register this?
I'm new to cucumber and only just learnt about registration by implementing TypeRegistryConfigurer.

Are you saying the step needs to be annotated?

(By the way it's a Java 8 style step)

gurpiarbassi:
It's alright. I worked it out.
Should have read the readme thoroughly.

PR #1754, opened 2019-08-28
v5.x JPMS Build

nhojpatrick:
## Summary

Getting Java 11 module-info.java into the code base

## Details

Know things/issues;
- things in ./pom.xml probably need to be pushed into cucumber-parent/pom.xml
- OpenEJB fails because of javax.ee issues and split package issues in dependencies.

## Motivation and Context

Here to get early insight and see what passes on travis.

## Types of changes

## Checklist:


nhojpatrick:
fyi the travis build failure is the, build failure i'm expecting when executed on Java 11

java/src/main/java11/module-info.java
```diff
@@ -0,0 +1,79 @@
...
+    exports io.cucumber.java.tl;
+    exports io.cucumber.java.tlh;
+    exports io.cucumber.java.tr;
+    exports io.cucumber.java.tt;
+    exports io.cucumber.java.uk;
+    exports io.cucumber.java.ur;
+    exports io.cucumber.java.uz;
+    exports io.cucumber.java.vi;
+    exports io.cucumber.java.zh_cn;
+    exports io.cucumber.java.zh_tw;
```
mpkorstanje: I hope this list is auto generated. :laughing:

java/src/main/java11/module-info.java
```diff
@@ -0,0 +1,79 @@
...
+    requires io.cucumber.core;
```
mpkorstanje: Because of `io.cucumber.core.api.Scenario` shouldn't this be `requires transitive`? 

Not sure if I'd like that. Perhaps we should wrap `Scenario` in a to be created `io.cucumber.java.Scenario`?

core/src/main/java11/module-info.java
```diff
@@ -0,0 +1,26 @@
...
+    exports io.cucumber.core.options;
+    exports io.cucumber.core.plugin;
+    exports io.cucumber.core.reflection;
+    exports io.cucumber.core.runner;
+    exports io.cucumber.core.runtime;
+    exports io.cucumber.core.snippets;
+    exports io.cucumber.core.stepexpression;
+    requires gherkin; // FIXME filename-based automodules detected
+    requires io.cucumber.cucumberexpressions;
+    requires io.cucumber.datatable;
```
mpkorstanje: This too should be `require transitive`

mpkorstanje:
> Here to get early insight and see what passes on travis.

I missed that. Ignore my comments for now! It's a good start!

java/src/main/java11/module-info.java
nhojpatrick: noooooooooooooooope ... feel my pain

mpkorstanje: Okay. So there is a groovy script in `cucumber-java` that generates the annotations. May well have it generate these next time. :D

java/src/main/java11/module-info.java
nhojpatrick: at the moment i'm just looking at `requires`, before looking cleaning up more and sorting out `requires transitive`

pom.xml
```diff
@@ -235,6 +235,40 @@
...
+                <jdk>[11,)</jdk>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>build-helper-maven-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <id>add-source-Java_Platform_Module_System</id>
```
mpkorstanje: Note to self: We should lift this up to the parent pom.

pom.xml
```diff
@@ -235,6 +235,40 @@
...
+                                <phase>generate-sources</phase>
+                                <goals>
+                                    <goal>add-source</goal>
+                                </goals>
+                                <configuration>
+                                    <sources>
+                                        <source>${project.basedir}/src/main/java11</source>
```
mpkorstanje: As `module-info.java` was introduced in 9 should we add it starting from 9?

pom.xml
```diff
@@ -235,6 +235,40 @@
...
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+            <properties>
+                <project.custom.java.version>11</project.custom.java.version>
+                <maven.compiler.release>11</maven.compiler.release>
+            </properties>
```
mpkorstanje: Note to self: This could be a `<configuration>` section with a `<release>`  element. 
Sort out what `project.custom.java.version` does.

pom.xml
```diff
@@ -199,7 +199,7 @@
         <module>spring</module>
         <module>guice</module>
         <module>weld</module>
-        <module>openejb</module>
+<!--        <module>openejb</module>-->
```
mpkorstanje: I've pushed deprecating OpenEJB forward to V5.x so we can do JPMS in V6.

https://github.com/cucumber/cucumber-jvm/issues/1763

mpkorstanje:
@nhojpatrick I know you were looking to do the entire project at once including the dependencies. I think that is going to be more complicated then we anticipated. So instead I would like to start by turning the `cucumber-junit-platform-engine` into a JPMS module. Feature wise it is also a project that could immediately benefit from module support because the JUnit API already supports modules. 

Starting at the leaf modules of Cucumber should work right?

mpkorstanje:
Looking at this. It might work:

http://www.russgold.net/sw/2018/04/easier-than-it-looks/
https://github.com/meterware/multirelease-parent/blob/master/pom.xml

aslakhellesoy:
@nhojpatrick do you plan to finish this?

mpkorstanje:
I'd say we cancel this for now. With the current tooling it is a bit of a pain to get right. I'd rather wait until Cucumber JVM uses 11+. Then we get all the modules without workarounds.

mpkorstanje:
Though looks like Java 8 will be supported until 2030. Not sure if I would want to wait that long. Maybe tooling will have improved by then.

nhojpatrick:
Closing now until i've more time to restart this

PR #1758, opened 2019-08-28
travis openjdk13

nhojpatrick:
## Summary

See how travis builds cucumber using openjdk 12

## Types of changes

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ jobs:
     script: mvn -q verify
   - jdk: openjdk11
     script: mvn -q verify
+  - jdk: openjdk13
+    script: mvn -q verify
 
   # 1.3 Check java doc
   - jdk: openjdk8


aslakhellesoy:
We've recently moved the cucumber/cucumber monorepo off TravisCI and onto CircleCI. The main reason for this is performance (shorter queues, faster execution, better caching).

I'd rather not invest more in TravisCI.

aslakhellesoy:
But until we do migrate, it's fine to add this I think. We should probably allow failures on some versions we don't support yet.

mpkorstanje:
```
ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:1.4.1:enforce (enforce) on project cucumber-jvm: Execution enforce of goal org.apache.maven.plugins:maven-enforcer-plugin:1.4.1:enforce failed: An API incompatibility was encountered while executing org.apache.maven.plugins:maven-enforcer-plugin:1.4.1:enforce: java.lang.ExceptionInInitializerError: null

[ERROR] ------------------------
```

Don't think this is Cucumber but rather our build tools.

PR #1761, opened 2019-08-30
Fixing junit categories

vtatai:
## Summary

This tries to fix the issue described https://groups.google.com/forum/#!searchin/cukes/junit$20categories%7Csort:date/cukes/jVqbTdwoAPo/SbyYmoeWEAAJ and previously addressed on the denied PR https://github.com/cucumber/cucumber-jvm/pull/1021

## Details

I take a different approach here than in the aforementioned PR, passing down the annotations to the `FeatureRunner` and `PickleRunners` so that they pass the filtering process.

Notice that my intent for this PR is primarily to get feedback - there are comments which would be removed / addressed as needed prior to merge. I also have an ugly hack to preserve the Pickle id and the annotations, which at this point I don't see an easy fix for without an upstream JUnit fix.

## How Has This Been Tested?

There are unit tests, and I've manually tested on a Java project (closed source unfortunately).

## Types of changes

## Checklist:

### DIFF
--- a/junit/src/main/java/io/cucumber/junit/Cucumber.java
+++ b/junit/src/main/java/io/cucumber/junit/Cucumber.java
@@ -40,6 +40,7 @@ import org.junit.runners.model.InitializationError;
 import org.junit.runners.model.RunnerScheduler;
 import org.junit.runners.model.Statement;
 
+import java.lang.annotation.Annotation;
 import java.time.Clock;
 import java.util.List;
 import java.util.function.Predicate;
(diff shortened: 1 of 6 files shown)


vtatai:
@mpkorstanje doesn't the change you suggested have the same issue you pointed out on https://github.com/cucumber/cucumber-jvm/pull/1021 ?
Also, I will check with the jUnit folks and get back here with more news.

mpkorstanje:
I don't believe this has the same issues.

Unlike #1021 regardless of outcome it always invokes super. This ensures that the state in the `ParentRunner` remains correct.

However I do not know if this breaks some other part of the contract. Hence the need to test this with various tools.

vtatai:
I am working on the proposed solution above, should have the code pushed soon.

mpkorstanje:
@vtatai have you considered using the `cucumber-junit-platform-engine` instead. The JUnit 5 Platform has build in tag support and it works with Maven too. I've been using it for a while now and it works quite well.

Note that you can't tag the `@Cucumber` annotated runner class. But every Feature and Scenario has it's tags translated to JUnit tags.

https://github.com/cucumber/cucumber-jvm/tree/master/junit-platform-engine#tags

Issue #1764, opened 2019-08-31
Add lambda equivalent for @ParameterType annotation


mpkorstanje:
# Summary

In https://github.com/cucumber/cucumber-jvm/pull/1677 `@ParameterType` , `@DataTableType` and various other annotations were introduced as an improvement over using the `TypeRegistryConfigurer`. However these annotations are part of `cucumber-java`. The lambda based `cucumber-java8` module doesn't have an equivalent yet. 

To make this possible additional methods have to be added to `io.cucumber.java8.LambdaGlue` as well as java 8 equivalents of the the `Java.*Definition` classes and some plumbing to register them with the glue . 

## Expected Behavior

Using `@DataTableType` rather then `@ParameterType` as an example

```java
public class RpnCalculatorSteps implements En {

    public RpnCalculatorSteps() {

        DataTableType((Map<String, String> entry) -> new RpnCalculatorSteps.Entry(
            Integer.valueOf(entry.get("first")),
            Integer.valueOf(entry.get("second")),
            entry.get("operation")
        ));

        DataTableType((Map<String, String> row) -> new ShoppingSteps.Grocery(
            row.get("name"),
            ShoppingSteps.Price.fromString(row.get("price"))
        ));

    }
}
```

## Current Behavior

Using `@DataTableType` rather then `@ParameterType` as an example

```java
public class TypeRegistryConfiguration implements TypeRegistryConfigurer {

    @Override
    public Locale locale() {
        return ENGLISH;
    }

    @Override
    public void configureTypeRegistry(TypeRegistry typeRegistry) {
        typeRegistry.defineDataTableType(new DataTableType(
            RpnCalculatorSteps.Entry.class,
            (Map<String, String> row) -> new RpnCalculatorSteps.Entry(
                Integer.valueOf(row.get("first")),
                Integer.valueOf(row.get("second")),
                row.get("operation")
            )
        ));

        typeRegistry.defineDataTableType(new DataTableType(
            ShoppingSteps.Grocery.class,
            (Map<String, String> row) -> new ShoppingSteps.Grocery(
                row.get("name"),
                ShoppingSteps.Price.fromString(row.get("price"))
            )
        ));
    }
}
````

PR #1768, opened 2019-09-10
[Java8] Add lambda parameter type annotations

rasklaad:
## Details

- [x] @DocStringType lambda implementation
- [x] @ParameterType lambda implementation
- [x] @DataTableType lambda inplementation
- [x]  ParameterType with multiple capture groups
- [x] Javadocs

## Motivation and Context
Closes: https://github.com/cucumber/cucumber-jvm/issues/1764

## How Has This Been Tested?

- [x] docstring tests
- [x] parametertype tests
- [x] datatable type tests

## Types of changes

## Checklist:


rasklaad:
I am kinda stuck with DataTableType implementation, can't get lambda method return type, to register it with LambdaGlueRegistry.
There is a test which represent this feature:
```
Scenario: define data table type by lambda
    Given data table, defined with lambda
      |name  | surname       | famousBook          |
      |Fedor | Dostoevsky    |Crime and Punishment |

DataTableType((Map<String, String> entry) -> {
            return new Author(entry.get("name"), entry.get("surname"), entry.get("famousBook"));
        });

Given("data table, defined with lambda", (Author author) -> {
            assertThat(author.name, equalTo("Fedor"));
            assertThat(author.surname, equalTo("Dostoevsky"));
            assertThat(author.famousBook, equalTo("Crime and Punishment"));
        });
```
Also, I still don't know how to implement ParameterType with multiple capture groups, without creating a lot of functional interfaces with different amount of String arguments.
I mean something like
```
class ParameterTypeDefinitionBody1<T> {
    T accept(String arg) throws Throwable;
}
class ParameterTypeDefinitionBody2<T> {
    T accept(String arg1, String arg2) throws Throwable;
}
class ParameterTypeDefinitionBody3<T> {
    T accept(String arg1, String arg2, String arg3) throws Throwable;
}
```
etc

mpkorstanje:
> can't get lambda method return type, to register it with LambdaGlueRegistry.


You can't get at the return type from `Method.getGenericReturnType` because `DataTableDefinitionBody<T>` is generic. So the class only knows that it has a generic `T` type variable.

However if you have an instance of that class you can get resolved types by using the `type-tools` dependency with `TypeResolver.resolveRawArguments.resolveRawArguments(instance, instance.getClass())`. You can find an example of this in `Java8StepDefinition`. 

Internally it looks into the constant types pool which is a JVM implementation detail but currently also the only way to provide this functionality. 

> Also, I still don't know how to implement ParameterType with multiple capture groups, without creating a lot of functional interfaces with different amount of String arguments.

There is no other way in Java. Best thing you can do is provide up to 9 parameters as we've done in `StepdefBody`.

timtebeek:
Work continued on #1782 ; Aside from merging master, here are my changes isolated from @rasklaad :
https://github.com/cucumber/cucumber-jvm/compare/edb0b77579405903c48c00cb52dd82a8d5aac6d5...lambda-type-annotations-continued
So mostly changing `Java8ParameterTypeDefinition` to function as `CaptureGroupTransformer` for `String[]` arguments, to support the additional parameter types with up to 9 arguments. Also: JavaDocs + tests.

java8/src/main/java/io/cucumber/java8/Java8DataTableTypeDefinition.java
```diff
@@ -0,0 +1,48 @@
...
+        if (Map.class.equals(parameterType)) {
+            return new DataTableType(
+                returnType,
+                (TableEntryTransformer<Object>) this::execute
+            );
+        }
+        else {
+            return new DataTableType(
+                returnType,
+                (TableTransformer<Object>) this::execute
```
mpkorstanje: [Coveralls indicates there is no coverage for this branch](https://coveralls.io/builds/25987334/source?filename=java8/src/main/java/io/cucumber/java8/Java8DataTableTypeDefinition.java#L33). And looking at [`JavaDataTableTypeDefinition.createDataTableType](https://github.com/cucumber/cucumber-jvm/blob/master/java/src/main/java/io/cucumber/java/JavaDataTableTypeDefinition.java#L71) I think there are some options missing.

java8/src/main/java/io/cucumber/java8/Java8DocStringTypeDefinition.java
```diff
@@ -0,0 +1,28 @@
...
+    public DocStringType docStringType() {
+        return docStringType;
+    }
+
+    Java8DocStringTypeDefinition(Object body, String contentType) {
+        super(body, new Exception().getStackTrace()[3]);
+        Class returnType = TypeResolver.resolveRawArguments(DocStringDefinitionBody.class, body.getClass())[0];
+        this.docStringType = new DocStringType(
+            this.method.getReturnType(),
+            contentType.isEmpty() ? method.getName() : contentType,
```
mpkorstanje: This was copied from `JavaDocStringTypeDefinition` but I don't think it makes sense here. Lambdas have no meaningfull method name so the content type is always mandatory. This should throw an error when content type is empty or null.

java8/src/main/java/io/cucumber/java8/DataTableDefinitionBody.java
```diff
@@ -0,0 +1,8 @@
+package io.cucumber.java8;
+
+import java.util.Map;
+
+@FunctionalInterface
+public interface DataTableDefinitionBody<T> {
```
mpkorstanje: This would be a `DataTableEntryDefinitionBody`. Not to be confused with the `TableRowDefinitionBody`, `TableCellDefinitionBody` and actual `DataTableDefinitionBody`. 

See [`JavaDataTableTypeDefinition`](https://github.com/cucumber/cucumber-jvm/blob/master/java/src/main/java/io/cucumber/java/JavaDataTableTypeDefinition.java#L71) for the full set.

java8/src/test/java/io/cucumber/java8/TypeDefinitionsStepdefs.java
```diff
@@ -0,0 +1,90 @@
...
+
+import io.cucumber.datatable.DataTable;
+
+import java.awt.Point;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
+
+import javax.xml.crypto.dsig.keyinfo.KeyValue;
```
mpkorstanje: Unsued?

java8/src/test/java/io/cucumber/java8/TypeDefinitionsStepdefs.java
```diff
@@ -0,0 +1,90 @@
+package io.cucumber.java8;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.junit.Assert.assertThat;
+
+import io.cucumber.datatable.DataTable;
+
+import java.awt.Point;
```
mpkorstanje: Looking forwards to Java 12+ it might be good to define a tuple of integers ourselves. That way we won't have to import the whole of awt.

timtebeek:
Addressed a few of the comments already; don't think I'll find time over the next few days to fix the remaining issues.. Anyone welcome to pick those up.

mpkorstanje:
Whoo all done! Thanks @rasklaad and @timtebeek! Now I've got to update the release notes again! :laughing:

rasklaad:
@timtebeek Thank you for the participation

timtebeek:
@rasklaad No problem at all; glad to help, and thanks for the final push to finish the work! :)

PR #1769, opened 2019-09-11
[Core] Fix timeline formatter

### LINKED ISSUE #1759: Remove or fix timeline formatter
A user reported the following via direct email:

---

Issue: Older version of JQuery library within io.cucumber (cucumber-core) is red flagged during nexus scan and identified as security vulnerable and need to be upgraded.
 
Details:
We are currently leverage  few of the MAVEN dependencies from io.cucumber for our BDD framework and  we are currently noticing a nexus scan error at The Hartford (Policy issue) because of older version of JQuery library used in the cucumber library.

Also, refer the below screenshot for the library where we are having this issue and we tried to update to the latest cucumber-core library and it is still referring to the old version of the JQuery library.

image

---

Now, we could easily upgrade to the latest jQuery (3.4.1 as of this writing), but how would we test that this doesn't break the formatter? From what I can tell we don't have any automated tests for the timeline formatter's JavaScript code.

I would suggest we delete the timeline formatter from Cucumber-JVM. It can be resurrected as a standalone formatter, similar to the new [cucumber-html-formatter](https://github.com/cucumber/cucumber/tree/master/html-formatter) (based on React, with a better test suite).

I also suggest we remove the old html-formatter - it uses an even older version of jQuery.

aslakhellesoy:
## Summary

* Upgrade jquery to 3.4.1
* Upgrade jquery-chosen to 1.8.7
* Deprecate the timeline formatter

This fixes #1759 

## Motivation and Context

The jquery upgrade is needed to prevent nexus and other malware-scanning systems from detecting an XSS vulnerability in the previous jquery (3.3.1).

The deprecation is motivated by the fact that the timeline formatter had regressed in two different ways without being caught by any automated tests:

* https://github.com/cucumber/cucumber-jvm/commit/4d34d6e2e37e0c0d31f0962319687d48f14af269 accidental modification of `vis.min.js` during a rename refactoring
* https://github.com/cucumber/cucumber-jvm/commit/4bb8876 changing `startTime` and `endTime` from `long` to `Instant`

## How Has This Been Tested?

Because there are no automated tests that will detect the modifications made, this has been tested manually:

* Build `cucumber-core`
* Run `cucumber-skeleton-java` with local modifications:
  * Multiple scenarios in multiple feature files
  * Surefire in parallel
  * Enable `timeline` formatter
  * Eyeball the result

## Types of changes

## Checklist:


mpkorstanje:
If you're going to merge this, please use squash merge. Keeps the history easy to ready.

aslakhellesoy:
Good point regarding the deprecation warning. I’ll remove it and only add it back once the new html formatter is usable from cucumber-jvm.

PR #1784, opened 2019-09-30
Print comment before and after StepText

alexandreonterroso:
## Summary

Print some comments before and after the stepText in the console.

## Details

Add the possibility to print custom comments before and after stepText in the console. 
```
new CommentBeforeStepText("This is an example comment before stepText");
new CommentAfterStepText("This is an example comment after stepText");
```

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/cucumber/runner/PickleStepTestStep.java
+++ b/core/src/main/java/cucumber/runner/PickleStepTestStep.java
@@ -1,6 +1,7 @@
 package cucumber.runner;
 
 import cucumber.api.TestCase;
+import cucumber.api.event.CommentEventSender;
 import cucumber.runtime.DefinitionArgument;
 import gherkin.pickles.PickleStep;
 
@@ -39,6 +40,10 @@ class PickleStepTestStep extends TestStep implements cucumber.api.PickleStepTest
         for (HookTestStep before : beforeStepHookSteps) {
             skipNextStep |= before.run(testCase, bus,  scenario, skipSteps);
         }
+        
+        CommentEventSender.getInstance().setBus(bus);
+        CommentEventSender.getInstance().setTestCase(testCase);
+        CommentEventSender.getInstance().setStepText(step.getText());
 
         skipNextStep |= super.run(testCase, bus,  scenario, skipNextStep);
 
--- a/core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
+++ b/core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
@@ -5,6 +5,8 @@ import cucumber.api.PickleStepTestStep;
 import cucumber.api.Result;
 import cucumber.api.TestCase;
 import cucumber.api.TestStep;
+import cucumber.api.event.CommentAfterStepTextEvent;
+import cucumber.api.event.CommentBeforeStepTextEvent;
 import cucumber.api.event.ConcurrentEventListener;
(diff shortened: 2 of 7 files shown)


core/src/main/java/cucumber/runner/PickleStepTestStep.java
```diff
@@ -39,6 +40,10 @@ boolean run(TestCase testCase, EventBus bus, Scenario scenario, boolean skipStep
         for (HookTestStep before : beforeStepHookSteps) {
             skipNextStep |= before.run(testCase, bus,  scenario, skipSteps);
         }
+
```
aslakhellesoy: We should avoid using singletons and setters in this codebase and use constructor parameters instead.

core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
```diff
@@ -43,6 +47,8 @@
     private ScenarioOutline currentScenarioOutline;
     private Examples currentExamples;
     private int locationIndentation;
+    private HashMap<String, ArrayList<String>> commentsBeforeStepText = new HashMap<String, ArrayList<String>>();
```
aslakhellesoy: Make fields `final` if they are not reassigned.

aslakhellesoy:
I'm not sure I understand the purpose of this feature. Can you give some examples of comments you want to print? We already have an API to attach/embed text and images. Did you consider using this?

mpkorstanje:
As implemented I would consider this feature to be non-viable. It will not work when executed in parallel. Like Aslak I'm also a bit puzzled about its purpose.

c64zottel:
Hello,
I work together with alexandreonterroso and he implemented the feature for me.
We have an application accessing another application via REST. Our testers want to have the REST requests/responses logged and eventually visible in the Living Documentation.

This patch allows us to store additional data (for us the REST communication) in the cucumber log.

mpkorstanje:
Have you considered using `Scenario.write` for this instead?

aslakhellesoy:
I see. It sounds like you have re-implemented a feature that is already present - embedding information in the test reports using `embed` and/or `write` from [Scenario](https://github.com/cucumber/cucumber-jvm/blob/b35f6a16e7aeb8ad8ff34cfb867815f5348a651d/java/src/main/java/io/cucumber/java/Scenario.java#L35-L41) - you can get a reference to it from a `Before` hook.

I think it would be better to change the pretty formatter to render this (it currently ignores it I think)

c64zottel:
Yeah, thats exactly what we were looking for. Since my colleage invested already some time, maybe he can add it in the pretty formater with an on/off switch?

aslakhellesoy:
Our only criterion for accepting code is the quality of the code - how well it solves the problem and how easy it will be to maintain. How much time was spent writing the code is not relevant I'm afraid.

In this case I think it's best to start over. I appreciate your contribution, but we cannot accept this as-is. It's always a good idea to get feedback on what you're about to implement before you implement it.

mpkorstanje:
> I think it would be better to change the pretty formatter to render this (it currently ignores it I think)

[It is included in the output](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/io/cucumber/core/plugin/PrettyFormatter.java#L103) but it ignores the steps indentation so it doesn't look so pretty. I would welcome some improvements there.

Issue #1797, opened 2019-10-09
Deprecate TypeRegistryConfigurer


mpkorstanje:
With the implementation of #1764 and #1677 done there is no need to support the `TypeRegistryConfigurer`. We should deprecate and eventually remove it.

mlvandijk:
Should we deprecate in v5 so we can remove in v6?

mpkorstanje:
Yes. I was planning to do this in in v5.x rather then v5.0 because it would make updating the documentation a bit easier but since we're working with release candidates now I think we can prepare the MR for v5.0 along with the release of v5.0.

PR #1799, opened 2019-10-11
[Core] Deprecate TypeRegistryConfigurer

rasklaad:
Deprecated TypeRegistryConfigurer, added more examples, how to use new configuration.

## Motivation and Context
https://github.com/cucumber/cucumber-jvm/issues/1797

## Types of changes

## Checklist:


core/src/main/java/io/cucumber/core/api/TypeRegistryConfigurer.java
```diff
@@ -6,8 +6,12 @@
 
 /**
  * The type registry configurer allows to configure a new type registry and the locale.
+ * @deprecated Please use annotation based configuration.
+ * See <a href="https://github.com/cucumber/cucumber-jvm/blob/master/examples/java-calculator/src/test/java/io/cucumber/examples/java/ShoppingSteps.java">Annotation based example</a>
+ * See <a href="https://github.com/cucumber/cucumber-jvm/blob/master/examples/java8-calculator/src/test/java/io/cucumber/examples/java8/ShoppingSteps.java">Lambda based example</a>
  */
-@API(status = API.Status.STABLE)
+@API(status = API.Status.DEPRECATED)
+@Deprecated
```
mpkorstanje: There is a subtle difference between the mean of these two deprecations. The API Guardian deprecation documents that a method annotated this way be removed at any time. The java deprecation makes no such statements. 

Since we only remove deprecated methods at major release we should still use the `Status.Stable` here.

examples/java-calculator/src/test/java/io/cucumber/examples/java/ShoppingSteps.java
```diff
@@ -19,9 +45,9 @@ public void the_following_groceries(List<Grocery> groceries) {
         }
     }
 
-    @When("I pay {}")
-    public void i_pay(int amount) {
-        calc.push(amount);
+    @When("I pay {price}")
+    public void i_pay(Price amount) {
```
mpkorstanje: I think the object should be called `Amount`. Perhaps we can make this example a bit more powerful by making `Amount` a tuple of `Currency` and `BigDecimal`.

core/src/main/java/io/cucumber/core/api/TypeRegistryConfigurer.java
```diff
@@ -6,8 +6,12 @@
 
 /**
  * The type registry configurer allows to configure a new type registry and the locale.
+ * @deprecated Please use annotation based configuration.
+ * See <a href="https://github.com/cucumber/cucumber-jvm/blob/master/examples/java-calculator/src/test/java/io/cucumber/examples/java/ShoppingSteps.java">Annotation based example</a>
+ * See <a href="https://github.com/cucumber/cucumber-jvm/blob/master/examples/java8-calculator/src/test/java/io/cucumber/examples/java8/ShoppingSteps.java">Lambda based example</a>
```
mpkorstanje: Ideally this should link to the documentation pages rather then github examples.

rasklaad: I can't really change configuration page in the docs repo before v5 release. This could mislead users.

So I'll change this links to https://cucumber.io/docs/cucumber/configuration/ after v5 release.

mpkorstanje: We also don't have to introduce this in 5.0.0. I'd rather have the docs in order so there is a clear way out before we deprecate it.

mpkorstanje: So. I was planning to do this in in v5.x rather then v5.0 because it would make updating the documentation a bit easier but since we're working with release candidates now I think we can prepare the MR for v5.0 along with the release of v5.0.

PR #1804, opened 2019-10-14
[Core] Extract Gherkin compatibility layer 

aslakhellesoy:
We want to support both Gherkin 5 and Gherkin 8 + Cucumber Messages in the next major release of Cucumber-JVM.


aslakhellesoy:
@mpkorstanje thanks for starting a branch for this. I've created a draft pull request so we have a place to discuss.

aslakhellesoy:
I tried adding this to the cucumber-core pom:

```xml
        <dependency>
            <artifactId>gherkin</artifactId>
            <version>5.2.0</version>
        </dependency>
        <dependency>
            <artifactId>gherkin</artifactId>
            <version>8.0.1-SNAPSHOT</version>
        </dependency>
```

Maven does not allow that, so I think we need to create a separate core module targetting gherkin 8

mpkorstanje:
> Maven does not allow that, so I think we need to create a separate core module targetting gherkin 8

Looks like the group id + artifact combination must be unique so we can only have 1 gherkin on the class path when using maven. Applied an exclusion to v5 in the calculator example and now it just works.

This should be sufficient for a spike-release.

examples/java-calculator/src/test/java/io/cucumber/examples/java/RunCucumberTest.java
```diff
@@ -5,7 +5,7 @@
 import org.junit.runner.RunWith;
 
 @RunWith(Cucumber.class)
-@CucumberOptions(plugin = {"json:target/cucumber-report.json"})
+@CucumberOptions()
```
mpkorstanje: TODO: Json plugin still uses gson.

gherkin8/src/main/java/io/cucumber/core/gherkin5/Gherkin8CucumberFeatureParser.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class Gherkin8CucumberFeatureParser implements CucumberFeatureParser {
+    private static CucumberFeature parseGherkin5(URI path, String source) {
+        try {
+
+            Stream<Messages.Envelope> messages = Gherkin.fromStream(new ByteArrayInputStream(source.getBytes(UTF_8)));
```
mpkorstanje: This conversion back to stream is needed because we read the `#encoding` pragma in [Encoding.java](https://github.com/cucumber/cucumber-jvm/blob/prep-work-gherkin-8/core/src/main/java/io/cucumber/core/feature/Encoding.java#L24). Ideally this is supported by Gherkin. Or we deprecate and drop it.

gherkin8/src/main/java/io/cucumber/core/gherkin5/Gherkin8CucumberFeatureParser.java
```diff
@@ -0,0 +1,62 @@
...
+
+            Parser<Messages.GherkinDocument.Builder> parser = new Parser<>(new GherkinDocumentBuilder());
+            Messages.GherkinDocument gherkinDocument = parser.parse(source).setUri(path.toString()).build();
```
mpkorstanje: Not sure where this URI goes but if it comes out again as string it might be useful. Would be better if this was a string.

gherkin8/src/main/java/io/cucumber/core/gherkin5/Gherkin8CucumberFeatureParser.java
```diff
@@ -0,0 +1,62 @@
...
+    private static List<CucumberPickle> compilePickles(String source, Messages.GherkinDocument document, GherkinDialectProvider dialectProvider, URI path) {
+        if (document.getFeature() == null) {
+            return Collections.emptyList();
+        }
+        String language = document.getFeature().getLanguage();
+        GherkinDialect dialect = dialectProvider.getDialect(language, null);
+        return new PickleCompiler().compile(document, path.toString(), source)
+            .stream()
+            .map(pickle -> new Gherkin8CucumberPickle(pickle, path, document, dialect))
+            .collect(Collectors.toList());
```
mpkorstanje: We don't use the messages here because we need some info from the AST in `cucumber-jvm` this make it hard to use just the pickles.

gherkin8/src/main/java/io/cucumber/core/gherkin5/Gherkin8CucumberPickle.java
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * Returns the line in feature file of the Scenario this pickle was created
+     * from. If this pickle was created from a Scenario Outline this line is the
+     *
+     * @return line in the feature file
+     */
+    @Override
+    public int getScenarioLine() {
+        List<Location> stepLocations = pickle.getLocationsList();
+        return stepLocations.get(stepLocations.size() - 1).getLine();
```
mpkorstanje: Is this still correct? I had to swap this in `Gherkin8CucumberStep`

gherkin8/src/main/java/io/cucumber/core/gherkin5/Gherkin8CucumberStep.java
```diff
@@ -0,0 +1,129 @@
...
+        if (dialect.getThenKeywords().contains(keyWord)) {
+            return StepType.THEN;
+        }
+        if (dialect.getAndKeywords().contains(keyWord)) {
+            return StepType.AND;
+        }
+        if (dialect.getButKeywords().contains(keyWord)) {
+            return StepType.BUT;
+        }
+        throw new IllegalStateException("Keyword " + keyWord + " was neither given, when, then, and, but nor *");
```
mpkorstanje: Ideally this would come from the pickle.

gherkin8/src/main/java/io/cucumber/core/gherkin5/Gherkin8CucumberStep.java
```diff
@@ -0,0 +1,129 @@
...
+                        return ruleChild.getScenario().getStepsList().stream();
+                    }
+                    if (ruleChild.hasBackground()) {
+                        return ruleChild.getBackground().getStepsList().stream();
+                    }
+                    return Stream.empty();
+                });
+        }
+
+        return Stream.empty();
```
mpkorstanje: No inheritance. :cry:

pom.xml
```diff
@@ -266,10 +265,14 @@
...
                         </dependency>
+                        <!--
+                        Gherkin is used during build time to generate code. We add it to the plugin classpath to prevent polluting
+                        the compile/test classpaths.
+                        -->
                         <dependency>
                             <groupId>io.cucumber</groupId>
                             <artifactId>gherkin</artifactId>
-                            <version>${gherkin.version}</version>
+                            <version>8.0.1-SNAPSHOT</version>
```
mpkorstanje: TODO: Make this is a fixed version for easier maintaince.

aslakhellesoy:
The `TestSourcesModel` class is tightly coupled to Gherkin5, and several formatters depend on it. I wonder if we can refactor it to depend on the `cucumber-gherkin` module instead...

aslakhellesoy:
`HTMLFormatter`, `JSONFormatter`, `PrettyFormatter` and `TimelineFormatter` all `import gherkin.*` as well. Aaaaaaargh!

mpkorstanje:
> Aaaaaaargh!

My reaction too. But we can leave those alone and implement the message formatter.

aslakhellesoy:
What do you mean by implementing the message formatter?

Currently, `cucumber-core` depends on `cucumber-gherkin5`. If I understand correctly, the idea is that `cucumber-core` should only depend on `cucumber-gherkin`, and use `ServiceLoader` to load either `cucumber-gherkin5` or `cucumber-gherkin8` at runtime.

So we need to remove the `cucumber-gherkin5` dependcency from `cucumber-core`.

If we're going to leave all those formatters unchanged, perhaps we should move them into `cucumber-gherkin5`?

mpkorstanje:
The idea was to ship with Gherkin 5 and allow people to test drive the protobuf message formatter with Gherkin 8 without getting into woods of fixing the existing formatters.

aslakhellesoy:
I agree with that goal. Do you agree we should move everything that depends on gherkin5 into the new cucumber-gherkin5 module?

mpkorstanje:
No. That would be allot of work all the tests for the formatters are integration tests.

mpkorstanje:
Like this:

https://github.com/cucumber/cucumber-jvm/blob/302b369d00dd66ba9d0532adcc1b743931dfef5d/examples/java-calculator/pom.xml#L43

And tell people not to use any formatters at runtime.

aslakhellesoy:
Aha!

mpkorstanje:
Been thinking about this problem some more. And I think we've been ignoring the elephant in the room.

People who write plugins (us included) will need an easy way to map pickles back to AST nodes. In essence something like `GherkinDocument.getStuffAtLine(int line)`. This API should be provided by Cucumber JVM/Gherkin but neither Gherkin 5 nor Gherkin 8 provide anything like it. A stable representation of the `GherkinDocument` decoupled from any parsing or messaging concerns should also make our lives a bit easier.

You made the suggestion to use `CucumberFeature` in all plugins. I would rather not do that. `CucumberFeature` is very focused on the internals of Cucumber. However we can apply the same trick to `GherkinDocument`. Gherkin 8 is a superset of Gherkin 5 after all.

This approach has some other advantage too. You've mentioned in the past the desire to support other formats for feature files gherkin. A stable representation will allow multiple parser implementations.

So in summary:

 1. Instead of implementing `CucumberFeature` for Gherkin5 and Gherkin8 we implement `GherkinDocument` for Gherkin5 and Gherkin8 and use that everywhere.

 2. We enrich `GherkinDocument` with methods to make mapping pickles to AST nodes easy.

 3. Gherkin becomes a multi-module project with `gherkin`, `gherkin5` and `gherkin8` modules.

aslakhellesoy:
This all sounds good to me. Not sure when I suggested to use `CucumberFeature` in all plugins?

As you know I want to decouple Cucumber even more from Gherkin and "features". I intend to start implementing a Markdown->Pickle compiler soon, and maybe even one for Excel. At this point, the Gherkin and "feature" abstractions break down. It's ok for formatters to depend on the input format, so they can render results on top of the input, which requires custom mapping from pickles back to the input model. Cucumber Core shouldn't know about input formats - it should only use pickles for execution and reporting.

I think of it as this pipeline:

```
           similar                                 +--------+
     +---------------------------------------------+ report |
     |                                             |document|
     |                                             +----^---+
     |                                                  |
     |                                                  |
     |                     all messages                 |
     |                     (source, ast, pickles)  +----+----+
     |                            +---------------->formatter|
     |                            |                +---^-----+
+----+---+      +------+      +---+----+               |
|document+------>parser+------>messages|               | result messages
+--------+      +------+      +---+----+               |
                                  |                    |
                                  |                +---+----+
                                  +---------------->cucumber|
                                pickle messages    +--------+

``` 

If I understand you correctly, you're suggesting implementing a query API for each input AST to make it easier to look up AST nodes from pickles. I think that makes a lot of sense. @vincent-psarga and I actually implemented this logic in Go when we ported the [JSON formatter](https://github.com/cucumber/cucumber/blob/master/json-formatter/go/json.go). This could easily be refactored and decoupled from the JSON parts, and then we could move it to the `cucumber-messages` module and port it to the other languages.

How does this sound?

mpkorstanje:
Sound in theory but I don't think pickles in their current state contain sufficient information to support both execution and reporting.

I also think we'll need a Pickle tree as part of this.

And I think if we do support multiple input formats reporters need to be agnostic of the input as well. Currently I don't see how that is supported.

aslakhellesoy:
I think it will be quite difficult to implement formatters that are agnostic of the input format *and* procude a report that is similar to the input format. I'm not convinced about the hierarchy thing - I have a slightly different vision. Cucumber-React / HTML Formatter is able to display a hierarchy - not because we have hierarchy information in the pickles, but because it traverses the `GherkinDocument` AST.

I don't know much about text-to-speech engines, but I imagine they have different implementations for different spoken languages. You can't feed French text to a Russian engine and expect good results. For the same reason I think a Gherkin pretty formatter (for example) can only pretty-print Gherkin, but not Markdown or Excel. Unless we come up with a montrosity abstraction for representing all possible document formats in the world.

I think it's more useful to group formatters in two categories:

* IDA: Input document agnostic
* IDD: Input document dependent

IDA formatters only need `Pickle` + `TestResult` to produce a report. Examples are `progress`, `junit` and `rerun`.

IDD formatters need AST + `Pickle` + `TestResult` to produce a report. Examples are `pretty`, `html`, `json`.

The AST is different for each input format:

* `GherkinDocument`
* `MarkdownDocument`
* `ExcelDocument`

Each IDD formatter will only work with one kind of input document. The Gherkin pretty formatter will not work with Markdown and Excel for example.

As you've pointed out, looking up AST nodes from pickles is complicated, because the AST has to be traversed first to build a lookup table. This is what `TestSourcesModel` does.

I think that concept is good - we just need to change their name so we don't give the impression that this is a universal model for all kinds input documents:

* `GherkinDocumentModel` 
* `MarkdownDocumentModel`
* `ExcelDocumentModel`

I imagine at runtime we'd have a single instance of e.g. `GherkinDocumentModel` which wraps all the `GherkinDocument` objects and provides a suitable lookup API for IDD Gherkin formatters.

core/src/main/java/io/cucumber/core/plugin/ProtobufFormatterAdapter.java
```diff
@@ -0,0 +1,28 @@
...
+import java.io.OutputStream;
+import java.lang.reflect.InvocationTargetException;
+
+public class ProtobufFormatterAdapter implements ConcurrentEventListener {
+    public static final String PROTOBUF_FORMATTER = "io.cucumber.core.gherkin8.formatter.ProtobufFormatter";
+    private final EventListener delegate;
+
+    public ProtobufFormatterAdapter(OutputStream out) {
+        try {
+            Class<EventListener> delegateClass = (Class<EventListener>) getClass().getClassLoader().loadClass(PROTOBUF_FORMATTER);
```
mpkorstanje: For now this is okay. 

But we should upgrade the plugin system to make registering plugins via SPI possible. Then the protobuf formatter could live in its own module.

aslakhellesoy: Completely agree - I added it to the gherkin8 module out of pure laziness since I think we're still in spike mode. Let's do that before we merge this.

gherkin8/src/main/java/io/cucumber/core/gherkin8/formatter/ProtobufFormatter.java
```diff
@@ -0,0 +1,142 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static java.util.Collections.singletonList;
+
+public class ProtobufFormatter implements EventListener {
```
mpkorstanje: Ideally the `ProtobufFormatter` would be living in the core module. But can't because gherking 5 and 8 have clashing artifact names. Do we want to resolve this?

core/src/main/java/io/cucumber/core/plugin/ProtobufFormatterAdapter.java
mpkorstanje: Cool!

gherkin8/src/main/java/io/cucumber/core/gherkin5/Gherkin8CucumberPickle.java
mpkorstanje: Then I would appreciate a descriptive method to access the last line number. Perhaps we can call it `pickleLine` because it is the line that defines the pickle. It would remove one obstacle that keeps me from using plain pickles in Cucumber.

aslakhellesoy: I like that idea. Just like we've talked about adding a `GherkinDocumentsModel` class to cucumber-messages, we could add a `PickleModel` class to cucumber-messages as well.

In short - some utility classes for getting interesting data out of those dumb protobuf structs.

mpkorstanje: Yes. Though I think utility classes are an anti-pattern when you own/can design the domain objects.

When ever you run into a pattern like:
```
Object dumb = ...
String propertyA = dumb.propertyA();
String propertyB = SmartUtils.propertyB(dumb)
```
It's better to rewrite it to a facade and use the facade everywhere. E.g:

```
Object dumb = ....
Smart  smart = new Smart(dumb);
String propertyA = smart.propertyA()
String propertyB = smart.propertyB();
```

aslakhellesoy: I think we’re after the same thing! This is basically the facade pattern.

aslakhellesoy: We could provide different flavours of facades around protobuf messages. One could process a stream of historical messages (stored in an event store like database, heck could even be Kafka). This way we could generate more interesting data like historical trends for execution time, flakiness, failure rates or even lead times from new to passing. The computed data could be stored in the database (if the computation is time consuming. This could be implemented as streams and pipes, allowing for composition.

mpkorstanje:
I just checked out the contents of `cucumber-messages`. It currently also contains "Cucumber-Engine Messages". Consumers don't need to know about the Cucumber Engine. So I think we should provide split `cucumber-messages` into a part for the engine and a part for the reporters.

mpkorstanje:
Gotta fix this. Protobuf should be shaded and relocated to `io.cucumber.gherkin.internal`.

image

mpkorstanje:
We need to fix a few things here:

1. resources in `google` should be relocated (this is a problem in `cucumber-messages`, it's missing a relocation pattern)
2. `cucumber-messages` should not be shaded into `gherkin` (this is a problem in `gherkin`, it's missing explicit inclusion sets.



image

aslakhellesoy:
I think we can safely exclude the `.proto` files from the jar. I don’t think they are required at runtime.

We use https://github.com/cucumber/cucumber/blob/master/.templates/java/scripts/check-jar.sh to verify jar contents during a release.

I suggest we do 2 things:
- Run this script as part of the build tobdiscover this before release (change `default.mk`)
- Fail if any `io.cucumber.*` classes outside this module are included in the shades jar

aslakhellesoy:
We can use `xmlstarlet` to extract this module’s package, assuming it coincides with its classes’ package (it should).

gherkin8/src/main/java/io/cucumber/core/gherkin8/formatter/ProtobufFormatter.java
aslakhellesoy: I think it’s pragmatic to keep it here for now. We can move it once we get rid of gherkin 5.

aslakhellesoy:
Regarding your suggestion to split up messages in smaller chunks. I want to do that eventually, but not until it's more stable. Some separation of concerns can be achieved in the single file by nesting related messages. The `GherkinDocument` AST for example is organised like that.

There has been a fair amount of flux in the message definitions lately, and I think that will continue until we have some feedback from real users.

When it settles down a bit and has a track record we can clean it up and separate it. For now I think it is more important to make changes more easily.

gherkin8/src/main/java/io/cucumber/core/gherkin8/formatter/ProtobufFormatter.java
mpkorstanje: I've resolved it by moving the protobuf formatter to it's own module. It only has a dependency on `cucumber-plugin` and a shaded dependency on Gherkin 8.

mpkorstanje:
What should be possible now is the following:

```
- Gherkin 5 + Json Formatter + Protobuf - Rule keyword ==> Works 
- Gherkin 8 + Json Formatter + Protobuf - Rule keyword ==> Works
- Gherkin 8 - Json Formatter + Protobuf + Rule keyword ==> Works
```

In my mind this meets the criteria for a smooth migration. People can setup a reporting tool that consumes protobuf. Then drop the json formatter switch to gherkin 8 and start using the rule keyword.

> When it settles down a bit and has a track record we can clean it up and separate it. For now I think it is more important to make changes more easily.

Fair enough.

If possible I'd like to restrict the `cucumber-messges` dependency to `cucumber-protobuf-plugin` for now. Trying to integrate it in Cucumber is another Yak. 

If anything I don't think piggy backing messages onto regular events is the way to go. Ideally we'd create a separate event that contains only the Message. But I would rather not commit to any specific implementation right now.

mpkorstanje:
Fixed a few formatters. Only the PrettyFormatter left to fix.

The JSON and HTML are lost causes as expected.

mpkorstanje:
A think to consider: The protbuf formatter prints either ndjson or protobufs. Perhaps we should rename to the cucumber messages formatter? That we it describe the protocol rather then the serialization format by the protocol.

mpkorstanje:
Notes to self: 
 - Abstract pickle tree to classes for containers and pickles only.
 - JUnit5 testId fragments depend on node type so include that (or use the keyword?). 
 - Remove getPickleAt (should be part of the tree).
 - Rename or remove `Cucumber*` prefix in `io.cucumber.core.gherkin`.
 - Clean up api of `io.cucumber.core.gherkin`. 
 - Rewrite Pretty formatter to use the pickle tree abstraction.
 - Rewrite Pretty formatter to print the tree-path to each pickle (keyword + text)
 - Make pretty formatter smart and only print the path if the previous pickle didn't print it yet.
 - Include embeddings as `[123456 bytes image/png]`
 - Rename keyword+name to keyword+???, something more descriptive anyway.
 -

aslakhellesoy:
On our last call we discussed how to implement new message-based formatters. Did we all agree we'll write them in the monorepo, using a shared test suite similar to gherkin's. We could use `fake-cucumber` to generate messages.

So I assume the new pretty formatter will go into the monorepo? There are a few there already, so should be relatively easy to parrot the design pattern. Essentially a `(InputStream in, OutputStream out)` function that reads messages from `in` and writes *something else* to `out`.

Shall we start with a simple one that essentially just prints pickles? We could use `cucumber-query` to look up the keyword. I think this would be good enough for 90% of our users.

Where do you think the pickle tree abstraction belongs? Inside `cucumber-query` perhaps?

mpkorstanje:
Fixing the TestNG and JUnit formatter came down to changing one small. We'll still have to lift them out but they're no longer in the way of a release now. One less Yak to shave for now.

There are a few assumptions around pretty formatter. Before committing to polyglot project and deal with the input/ouputstream stuff. I'd like to see what the output actually looks like and how it behaves during parallel execution and if we can reasonably work around that.

Currently Cucumber JVM has reasonable system for dealing with messages that occur out of order. We lose that mechanism if we switch to protobufs. If it all does work out it should be easy to lift the pretty formatter. The messages/events we need for the pretty formatter have a 1-1 correspondence.

mpkorstanje:
> Where do you think the pickle tree abstraction belongs? Inside cucumber-query perhaps?

Yeah we'd need something like `cucumberQuery.getAncestors(pickle)` that returns a list of `[Feature, Rule, Scenario, Example Section, Example]`. So we can get the key word and title of each. But we'll see it when we get there.

mpkorstanje:
> We use https://github.com/cucumber/cucumber/blob/master/.templates/java/scripts/check-jar.sh to verify jar contents during a release.

Done. 

https://github.com/cucumber/cucumber/pull/819

mpkorstanje:
I think we have something mergeable now. Only 4 things remain:

 - [ ] Finish message protocol
 - [ ] Release version of Messages
 - [ ] Release version of Gherkin
 - [ ] Integration tests for `message` formatter.

PR #1809, opened 2019-10-17
[Core] Indent write events in PrettyFormatter 

alexandreonterroso:
## Summary

Add indentation to PrettyFormatter

## Motivation and Context

As @mpkorstanje said in #1784 , scenario.write() ignores the steps indentation so it doesn't look so pretty.

### DIFF
--- a/core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
+++ b/core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
@@ -17,6 +17,7 @@ import cucumber.api.event.TestStepStarted;
 import cucumber.api.event.WriteEvent;
 import cucumber.api.formatter.ColorAware;
 import cucumber.api.formatter.NiceAppendable;
+import cucumber.runtime.CucumberException;
 import cucumber.util.FixJava;
 import cucumber.util.Mapper;
 import gherkin.ast.Background;
@@ -28,6 +29,9 @@ import gherkin.ast.Step;
 import gherkin.ast.Tag;
 import gherkin.pickles.PickleTag;
 
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.StringReader;
 import java.net.URI;
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
```diff
@@ -150,7 +151,9 @@ private void handleTestStepFinished(TestStepFinished event) {
     }
 
     private void handleWrite(WriteEvent event) {
-        out.println(event.text);
+        out.println();
+        out.println(STEP_SCENARIO_INDENT + event.text);
+        out.println();
```
mpkorstanje: You may want to test this with multi-line strings.

alexandreonterroso: You are right, I added some changes

core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
```diff
@@ -150,8 +151,18 @@ private void handleTestStepFinished(TestStepFinished event) {
...
-        out.println(event.text);
-    }
+        out.println();
+        try(BufferedReader lines = new BufferedReader(new StringReader(event.text))) {
+		String line = lines.readLine();
+		while(line != null) {
+			out.println(STEP_SCENARIO_INDENT + line);
+			line = lines.readLine();
+		}
+	} catch (IOException e) {
```
mpkorstanje: The canonical way to write this avoids the repetition of `readLine`. It is a little bit less error prone.

```
            String line;
            while ((line = lines.readLine()) != null) {
                out.println(STEP_SCENARIO_INDENT + line);
            }
```

Aside from that looks good. Don't forget to indent your code properly.

core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
```diff
@@ -150,8 +151,17 @@ private void handleTestStepFinished(TestStepFinished event) {
     }
 
     private void handleWrite(WriteEvent event) {
-        out.println(event.text);
-    }
+        out.println();
+        try(BufferedReader lines = new BufferedReader(new StringReader(event.text))) {
+		String line;
```
mpkorstanje: Looks better. But still not quite right. If you are using an IDE there usually is a short cut to do this for you.

core/src/main/java/cucumber/runtime/formatter/PrettyFormatter.java
```diff
@@ -150,8 +151,17 @@ private void handleTestStepFinished(TestStepFinished event) {
     }
 
     private void handleWrite(WriteEvent event) {
-        out.println(event.text);
-    }
+        out.println();
+        try(BufferedReader lines = new BufferedReader(new StringReader(event.text))) {
+			String line;
+			while ((line = lines.readLine()) != null) {
```
mpkorstanje: I don't mean to keep going on about this but please use spaces for indentation rather then tabs.

We've documented the settings to configure your IDE here:

https://github.com/cucumber/cucumber-jvm/blob/master/CONTRIBUTING.md#contributing

And we also provide a .editorconfig file that most IDEs will pick up automatically.

https://github.com/cucumber/cucumber-jvm/blob/master/.editorconfig

mpkorstanje:
Had to manually merge this because your repo was lagging behind Cucumber JVM quite a bit. 

Cheers!

PR #1816, opened 2019-11-05
Fix tests on Windows

janosvitok:
## Summary

The following changes are needed to successfully run `mvn test` on Windows.

## Details

- exception messages have \r\n (CRLF) line separator, so the tests must account for it
- files can't be deleted while they are still open, so we have to `close()` URLOutputStreams

## How Has This Been Tested?

`mvn test` run successfully. Windows 10 latest, JDK 1.8, mvn 3.6.2

## Types of changes

## Checklist:

### DIFF
--- a/core/src/test/java/io/cucumber/core/plugin/URLOutputStreamTest.java
+++ b/core/src/test/java/io/cucumber/core/plugin/URLOutputStreamTest.java
@@ -189,7 +189,8 @@ class URLOutputStreamTest {
                 try {
                     // Every thread should wait command to run
                     countDownLatch.await();
-                    new URLOutputStream(tmp.toURI().toURL());
+                    try (URLOutputStream urlOutputStream = new URLOutputStream(tmp.toURI().toURL())) {
+                    }
                 } catch (IOException e) {
                     threadErrors.add("Thread" + curThreadNo + ": parent dir not created. " + e.getMessage());
                 } catch (InterruptedException e) {
--- a/guice/src/test/java/io/cucumber/guice/GuiceFactoryTest.java
+++ b/guice/src/test/java/io/cucumber/guice/GuiceFactoryTest.java
@@ -19,6 +19,7 @@ import java.util.Arrays;
 import java.util.List;
 
 import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.hamcrest.Matchers.equalToCompressingWhiteSpace;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.core.Is.is;
 import static org.hamcrest.core.IsEqual.equalTo;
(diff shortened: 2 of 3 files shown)


core/src/test/java/io/cucumber/core/plugin/URLOutputStreamTest.java
```diff
@@ -189,7 +189,8 @@ private Reader openUTF8FileReader(final File file) throws IOException {
                 try {
                     // Every thread should wait command to run
                     countDownLatch.await();
-                    new URLOutputStream(tmp.toURI().toURL());
+                    URLOutputStream urlOutputStream = new URLOutputStream(tmp.toURI().toURL());
+                    urlOutputStream.close();
```
mpkorstanje: Nice catch. But let's use try-with-resources here.

guice/src/test/java/io/cucumber/guice/GuiceFactoryTest.java
```diff
@@ -58,9 +58,9 @@ void factoryStartFailsIfScenarioScopeIsNotBound() {
         Executable testMethod = () -> factory.start();
         ConfigurationException actualThrown = assertThrows(ConfigurationException.class, testMethod);
         assertThat("Unexpected exception message", actualThrown.getMessage(), is(equalTo(
-            "Guice configuration errors:\n\n" +
-                "1) No implementation for io.cucumber.guice.ScenarioScope was bound.\n" +
-                "  while locating io.cucumber.guice.ScenarioScope\n\n" +
+            "Guice configuration errors:" + System.lineSeparator() + System.lineSeparator() +
+                "1) No implementation for io.cucumber.guice.ScenarioScope was bound." + System.lineSeparator() +
+                "  while locating io.cucumber.guice.ScenarioScope" + System.lineSeparator() + System.lineSeparator() +
```
mpkorstanje: It might better to use `equalToCompressingWhiteSpace`instead. It is allot more readable. Both here and in `DocStringTypeRegistryDocStringConverterTest`.

core/src/test/java/io/cucumber/core/plugin/URLOutputStreamTest.java
janosvitok: I was not sure about Java level you aim for and if I can use t-w-r.

guice/src/test/java/io/cucumber/guice/GuiceFactoryTest.java
janosvitok: Nice. I've updated code.

core/src/test/java/io/cucumber/core/plugin/URLOutputStreamTest.java
mpkorstanje: Its java 8. It's hidden somewhere in the parent pom.

mpkorstanje:
Cheers!

mpkorstanje:
Btw. We have an open collective to pay and reimburse contributors. If you want to feel free to send an invoice to https://opencollective.com/cucumber

Issue #259, opened 2012-03-14
Running cucumber tests in Eclipse causes exception


remast:
Running cucumber tests in eclipse causes an exception because the *.feature files exist duplicated in the classpath. Eclipse puts all resources including the *.feature files in the classpath. Additionally the classpath contains the directory with the resources.
-> As a result two identical versions of the same *.feature files exist in the classpath.

This causes the following code in the FeatureBuilder to throw a CucumberException:

``` java
public void parse(Resource resource, List<Object> filters) {
    Formatter formatter = this;
    if (!filters.isEmpty()) {
        formatter = new FilterFormatter(this, filters);
    }
    Parser parser = new Parser(formatter);
    String gherkin = read(resource);

    String checksum = checksum(gherkin);
    String path = pathsByChecksum.get(checksum);
    if (path != null) {
        throw new CucumberException(String.format("Found the same source in %s and %s", path, resource.getPath()));
    }
    pathsByChecksum.put(checksum, resource.getPath());
```

Having two identical *.feature files on the claspath does no harm. So I suggest that's tolerated as long as both files have the same checksum. If desired a warning could be logged.

Issue #165, opened 2012-01-22
Warn or fail if duplicate features are detected


jvans:
When I ran my single scenario test from Maven using Main, the generated report.js had my scenario listed twice. I will see if I can isolate this a little further to get a clearer picture of what is going on.

PR #1819, opened 2019-11-14
[Core] Include name in duplicate feature detection

mpkorstanje:
## Summary

Build tools will typically copy the source files to a `build` or
`target` directory duplicating all feature files in the project. When
the root of the project is used as the feature path cucumber will
discover both sets of feature files (https://github.com/cucumber/cucumber-jvm/issues/165).

By including the file name we can distinguish between accidental
duplication and intentional duplication.

We could make this scenario an error but given the complexities of https://github.com/cucumber/cucumber-jvm/issues/259
this is best left as a separate issue.

## Types of changes

## Checklist:


laxersaz:
Thanks a lot @mpkorstanje !

PR #1841, opened 2019-12-13
[Core] Use Gherkin 6+ and Cucumber messages

mpkorstanje:
## Summary

 * Implements the `message` formatter which will output cucumber messages as ndjson as well as an improved `html` formatter which renders a html report based on these messages.

* Switches the default parser to Gherkin 6+. This will enable rule support by default. 

* The `json` formatter has been made optional and has been moved to `gherkin-vintage`. It does not support rules and will throw an exception when rules are used. 

## Types of changes

## Checklist:


mpkorstanje:
Good to keep in mind: File iteration order in a directory is not guaranteed. So we can only do TCK tests with a single feature file at once or we have to specify each feature file individually.

mpkorstanje:
Proposal: Rename message formatter to `protobuf` and `ndjson`. This way we don't have to explain to people who make typos in the file name `out.ndnjson` why they suddenly get a binary blob.

mpkorstanje:
Consideration: Add some maven checks to the build to ensure dependency convergence. We currently have allot of dependencies and transitive dependencies on messages. These should all be in sync. While the protobuf itself is forwards and backwards compatible, the bytecode is decidedly not.

core/src/main/java/io/cucumber/core/plugin/URLOutputStream.java
```diff
@@ -1,6 +1,6 @@
 package io.cucumber.core.plugin;
 
-import gherkin.deps.com.google.gson.Gson;
+import io.cucumber.messages.internal.com.google.gson.Gson;
```
mpkorstanje: We shouldn't need GSON here at all.

PR #1847, opened 2019-12-20
[Core] Include default gherkin version in version.properties

davidjgoss:
## Summary

Expose the gherkin version in `version.properties`, so it can be picked up from the classpath by other code.

## Motivation and Context

A bit more context in https://github.com/cucumber/cucumber/issues/779#issuecomment-561220566, but I am specifically targeting the IDEA autocompletion support for Rule/Example grammar.

I am opening this now for more of a steer, as I can see that with #1804 and others we now have multiple Gherkin versions so this would no longer be valid, and I'm not sure what the best way to achieve this kind of feature detection would be now - or if it's the right thing to do at all.

## Types of changes

## Checklist:

(Sorry, meant to open this as a draft)

### DIFF
--- a/core/src/main/resources/io/cucumber/core/version.properties
+++ b/core/src/main/resources/io/cucumber/core/version.properties
@@ -1,2 +1,2 @@
 cucumber-jvm.version=${parent.version}
-gherkin.version=${gherkin.version}
+gherkin.version=${gherkin-messages.version}
--- a/core/src/main/resources/io/cucumber/core/version.properties
+++ b/core/src/main/resources/io/cucumber/core/version.properties
@@ -1,2 +1,2 @@
 cucumber-jvm.version=${parent.version}
-gherkin.version=${gherkin-messages.version}
+gherkin.version=${gherkin-vintage.version}
--- a/core/src/main/resources/io/cucumber/core/version.properties
+++ b/core/src/main/resources/io/cucumber/core/version.properties
@@ -1,1 +1,2 @@
 cucumber-jvm.version=${parent.version}
+gherkin.version=${gherkin.version}


mpkorstanje:
Mmh. This is an interesting problem.

mpkorstanje:
While the introduction of the compatibility layer may suggest otherwise, the plan for Cucumber is not to support multiple versions of Gherkin indefinitely. Rather we are trying to solve a migration problem.

Cucumbers `json` and `html` plugins create reports on the execution of Cucumber and they do so by creating a facsimile of the original feature file. As a result they are quite complex and hard to maintain. And on top of that they are duplicated in the different implementations of Cucumber. Finally the output produced by the `json` formatter doesn't have a schema and differs per cucumber implementation.

We're looking to solve several problems all at once by introducing a new output format based on protobuf/ndjson. However as there are no consumers of protobuf/ndjson yet we can't simply remove the `json` formatter. We'll have to support it for at least a little while. This means that internally we need parsers for both Gherkin 5 and Gherkin 8+.

core/src/main/resources/io/cucumber/core/version.properties
```diff
@@ -1 +1,2 @@
-cucumber-jvm.version=${parent.version}
\ No newline at end of file
+cucumber-jvm.version=${parent.version}
+gherkin.version=${gherkin-messages.version}
```
mpkorstanje: Currently (RC3) we support Gherkin 5 by default and Gherkin 8+ (rule keyword) as an opt in module. However only JUnit 5 can accurately represent these outputs so I wouldn't enable it by default. Once we've finished the protobuf/ndjson output we'll probably enable Gherkin 8+ by default. 

Which is a really round about way to say that I think this should be: 
```suggestion
gherkin.version=${gherkin-vintage.version}
```

davidjgoss:
@mpkorstanje okay yeah - so we declare the vintage version for now as it's the default. Thanks!

Issue #1846, opened 2019-12-20
Using cucumber-spring does not share Spring context when using parallel (methods) surefire settings


dom54:
## Summary

I have 3 feature files, with a single StepDefinition.   When running in single threaded mode, everything is perfect.   When running with the maven surefire plugin, using parallel as methods,  I get a spring context created per thread, which I cannot have (startup time of the app would mean no benefit to threading)

```java
@WebAppConfiguration
@ContextHierarchy({
        @ContextConfiguration(classes = { TestConfiguration.class }, loader = AnnotationConfigWebContextLoader.class),
        @ContextConfiguration(classes = { ServletConfiguration.class }, loader = AnnotationConfigWebContextLoader.class)
})
public class DefaultSteps {
```

```xml
 <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-surefire-plugin</artifactId>
                <version>2.19.1</version>
                <configuration>
                    <parallel>methods</parallel>
                    <threadCount>8</threadCount>
```
I am working on a test case to share

## Expected Behavior

The problem is the code uses a "TestContext" which binds a context to ThreadLocal.

So I worked around this by putting my own bootstrapper
```java
@BootstrapWith(value = MultiThreadedAwareSpringBootstrapper.class)
public class DefaultSteps {
```

```java
public class MultiThreadedAwareSpringBootstrapper extends org.springframework.test.context.web.WebTestContextBootstrapper {

    @Override
    public TestContext buildTestContext() {
        return new DefaultTestContext(getBootstrapContext().getTestClass(), buildMergedContextConfiguration(),
                new DefaultCacheAwareContextLoaderDelegate()); // dont pass in a context ,which will then use a static default cache of contexts, i.e. share between threads
    }

}
```

## Current Behavior
This works, except once in a while it blows up (SpringFactory seems not to be thread safe)

```java
org.springframework.beans.factory.NoUniqueBeanDefinitionException: No qualifying bean of type 'com.refacted.redacted.acceptance.glue.DefaultSteps' available: expected single matching bean but found 1: com.redacted.redacted.acceptance.glue.DefaultSteps


```
## Possible Solution
Make SpringFactory threadsafe...still investigating.

In particular this line is not thread safe (if the context is shared)
```java
context.getBeanFactory().registerScope(SCOPE_CUCUMBER_GLUE, new GlueCodeScope());
```

## Steps to Reproduce (for bugs)
Working on it...

[...]

Issue #1583, opened 2019-03-13
sharing spring context with plain junit tests


MarcHauptmann:
## Summary

A spring context is not shared between a junit test and cucumber. See #1582 for tests to reproduce this issue.

## Expected Behavior

It would be nice if cucumber was able to share the spring context with other tests.

## Current Behavior

Cucumber manages its own context cache(s) and does not share these. 

## Steps to Reproduce (for bugs)

Run the tests from the pull request mentioned above.

mpkorstanje:
Excellent thanks! I agree. Unfortunately the way `cucumber-spring` is implemented make this impossible to do while also supporting parallel execution. The only way forward is to implement a new spring module that doesn't add step definitions to the spring context.

Have a look at https://github.com/cucumber/cucumber-jvm/issues/1470 for details.

MarcHauptmann:
Ok, thanks for your answer! I will see what I can to do implement this feature.

Issue #1823, opened 2019-11-19
CucumberException: No qualifying bean of type 'Steps' available: expected single matching bean but found 1: Steps


noel-yap:
When using a custom application loader, `expected single matching bean but found 1` is emitted. This behavior is non-deterministic but it seems sticky (I'm guessing there's some caching going on).

We have something like the following code:
```
public class SharedApplicationContextLoader extends SpringBootContextLoader {
  @SpringBootApplication
  static class ApplicationContextSourceFallback {
  }

  private static final AtomicReference<Supplier<ApplicationContext>> applicationContextSupplier = new AtomicReference<>(null);
  private static final AtomicReference<ApplicationContext> applicationContext = new AtomicReference<>(null);

  public static synchronized void setApplicationContext(final ApplicationContext applicationContext) {
    setApplicationContext(() -> applicationContext);
  }

  private static synchronized void setApplicationContext(final Supplier<ApplicationContext> applicationContextSupplier) {
    SharedApplicationContextLoader.applicationContextSupplier.compareAndSet(null, applicationContextSupplier);

    SharedApplicationContextLoader.applicationContext.compareAndSet(
        null,
        SharedApplicationContextLoader.applicationContextSupplier.get().get());
  }

  @Override
  public ApplicationContext loadContext(final MergedContextConfiguration config) {
    SharedApplicationContextLoader.setApplicationContext(() -> {
      try {
        return super.loadContext(config);
      } catch (final Exception e) {
        throw new RuntimeException(e);
      }
    });

    return applicationContext.get();
  }

  @Override
  protected SpringApplication getSpringApplication() {
    return new SpringApplication(ApplicationContextSourceFallback.class);
  }
}
```

```
@SpringBootTest
@ContextConfiguration(
    classes = TestsConfiguration.class,
    loader = SharedApplicationContextLoader.class)
public class Steps {
}
```

```
@SpringBootApplication
@ContextConfiguration(classes = TestsConfiguration.class)
public class Main {
  public static void main(String[] args) {
    final ConfigurableApplicationContext applicationContext = SpringApplication.run(Main.class, args);

    SharedApplicationContextLoader.setApplicationContext(applicationContext);
  }
}
```

**Expected behavior**
No exception thrown.

**Context & Motivation**
During development, the tests are run directly from the IDE. Once committed and pushed, an app is built and deployed housing the tests and will execute them.

**Stacktrace**
```
[...]

PR #1848, opened 2019-12-20
[Spring] Share application context

dom54:
When executing Cucumber in parallel in combination with `cucumber-spring` users would often encounter a cryptic error message: `No qualifying bean of type 'Steps' available: expected single matching bean but found 1` (#1823). 

This was initially fixed by ensuring that the Spring application context was not shared between threads (#1153, #1148). This however has the  downside that the application context is not shared with JUnit tests either (#1583) and comes at significant performance penalty (#1846). 

The root cause however was a race condition in `DefaultListableBeanFactory.registerBeanDefinition(String beanName, BeanDefinition beanDefinition)`. There is a check to see if the definition already exists. If it does not exist, it replaces the bean definition in a `ConcurrentHashMap` (in a synchronized block as well) but then adds the bean name to a list without checking if it already exists.

By synchronizing inside `SpringFactory.start` we prevent concurrent modification of the application context by `cucumber-spring` This in turn makes it possible to share the application context between threads and with JUnit tests.

Fixes: #1846
Fixes: #1583
Closes: #1848
Closes: #1582


dom54:
making this the default - share context between all threads

spring/src/main/java/io/cucumber/spring/SpringFactory.java
```diff
@@ -249,8 +251,10 @@ private ConfigurableApplicationContext getContext() {
 
         private void registerGlueCodeScope(ConfigurableApplicationContext context) {
             do {
-                context.getBeanFactory().registerScope(SCOPE_CUCUMBER_GLUE, new GlueCodeScope());
-                context = (ConfigurableApplicationContext) context.getParent();
+                synchronized (context) {
+                    context.getBeanFactory().registerScope(SCOPE_CUCUMBER_GLUE, new GlueCodeScope());
```
dom54: registerScope is NOT threadsafe.  

In AbstractBeanFactory in Spring:

Scope previous = (Scope)this.scopes.put(scopeName, scope); //scopes is a LinkedHashMap

spring/src/main/java/io/cucumber/spring/FixBootstrapUtils.java
```diff
@@ -57,7 +52,7 @@
 			"org.springframework.test.context.web.WebAppConfiguration";
 
     static BootstrapContext createBootstrapContext(Class<?> testClass) {
-        CacheAwareContextLoaderDelegate contextLoader = new DefaultCacheAwareContextLoaderDelegate(contextCache.get());
+        CacheAwareContextLoaderDelegate contextLoader = new DefaultCacheAwareContextLoaderDelegate();
```
dom54: Not passing in a context will use a static (thread safe) context cache, which will be shared between threads

spring/src/main/java/io/cucumber/spring/SpringFactory.java
```diff
@@ -139,8 +139,10 @@ public void start() {
         notifyContextManagerAboutTestClassStarted();
         if (beanFactory == null || isNewContextCreated()) {
             beanFactory = testContextManager.getBeanFactory();
-            for (Class<?> stepClass : stepClasses) {
-                registerStepClassBeanDefinition(beanFactory, stepClass);
+            synchronized (beanFactory) {
+                for (Class<?> stepClass : stepClasses) {
+                    registerStepClassBeanDefinition(beanFactory, stepClass);
```
dom54: registerStepClassBeanDefinition is not threadsafe. Code in DefaultListableBeanFactory

```java
            if (this.hasBeanCreationStarted()) {
                synchronized(this.beanDefinitionMap) { //its synchronized, but it doesnt check again before adding to the updatedDefinitions list
                    this.beanDefinitionMap.put(beanName, beanDefinition);
                    List<String> updatedDefinitions = new ArrayList(this.beanDefinitionNames.size() + 1);
                    updatedDefinitions.addAll(this.beanDefinitionNames);
                    updatedDefinitions.add(beanName);
                    this.beanDefinitionNames = updatedDefinitions;
```

spring/src/main/java/io/cucumber/spring/SpringFactory.java
```diff
@@ -249,8 +251,10 @@ private ConfigurableApplicationContext getContext() {
 
         private void registerGlueCodeScope(ConfigurableApplicationContext context) {
             do {
-                context.getBeanFactory().registerScope(SCOPE_CUCUMBER_GLUE, new GlueCodeScope());
-                context = (ConfigurableApplicationContext) context.getParent();
+                synchronized (context) {
```
mpkorstanje: I don't think this works. We're swapping out `context` with a different instance.  We should probably use a constant as a monitor.

spring/src/main/java/io/cucumber/spring/SpringFactory.java
```diff
@@ -139,8 +139,10 @@ public void start() {
         notifyContextManagerAboutTestClassStarted();
         if (beanFactory == null || isNewContextCreated()) {
             beanFactory = testContextManager.getBeanFactory();
-            for (Class<?> stepClass : stepClasses) {
-                registerStepClassBeanDefinition(beanFactory, stepClass);
+            synchronized (beanFactory) {
```
mpkorstanje: We also have a synchronization step a few lines up inside `CucumberTestContextManager`. For clarity we could synchronize all operations in the start method.

spring/src/main/java/io/cucumber/spring/SpringFactory.java
dom54: the synchronization will be on the first reference to context passed in, but I agree it looks "unclean"

spring/src/main/java/io/cucumber/spring/SpringFactory.java
dom54: done

dom54: So now - we synchronize the start method statically. So for all instances of SpringFactory created they will start in serial

spring/src/main/java/io/cucumber/spring/SpringFactory.java
```diff
@@ -26,8 +26,6 @@
 import java.util.Set;
 
 import static io.cucumber.spring.CucumberTestContext.SCOPE_CUCUMBER_GLUE;
-import static io.cucumber.spring.FixBootstrapUtils.createBootstrapContext;
-import static io.cucumber.spring.FixBootstrapUtils.resolveTestContextBootstrapper;
```
mpkorstanje: Removed the `FixBootstrapUtils`. Wasn't needed any more.

spring/src/main/java/io/cucumber/spring/SpringFactory.java
```diff
@@ -129,21 +129,24 @@ private static boolean hasAnnotation(Annotation annotation, Collection<Class<? e
 
     @Override
     public void start() {
-        if (stepClassWithSpringContext != null) {
-            testContextManager = new CucumberTestContextManager(stepClassWithSpringContext);
-        } else {
-            if (beanFactory == null) {
-                beanFactory = createFallbackContext();
+        synchronized (monitor) {
```
mpkorstanje: Replaced the `SpringFactory.class` with `monitor`. The class is accessible by any one. And while I don't expect any one to use it to synchronize on it, there also no need to do so.

mpkorstanje: Also explained the why of it all.

spring/src/test/java/io/cucumber/spring/threading/RunParallelCucumberTest.java
```diff
@@ -1,21 +1,23 @@
...
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.junit.jupiter.api.Assertions.assertAll;
 
-class RunParallelCukesTest {
+class RunParallelCucumberTest {
 
-    @Test
-    void test() {
+    @RepeatedTest(100)
```
mpkorstanje: We're running 100 repetitions with 100 concurrent scenarios. Should tone this down a bit for CI but good enough.

mpkorstanje:
```
[INFO] Cucumber-JVM: Spring ............................... SUCCESS [02:12 min]
```
:rofl:

mpkorstanje:
To keep the history readable PRs are squashed when merging. I've drafted a commit message in your original description of the PR. You should be able to edit it.

dom54:
looks good to me

aslakhellesoy:
Hi @dom54,

Aslak Hellesøy
Creator of Cucumber

mpkorstanje:
Aright then we're done!

PR #1851, opened 2020-01-01
[JUnit] Use JUnit 4.13

nhojpatrick:
## Summary

Heard junit v4.13 was released, so tried building cucumber, fixes in separate PR as not 100% sure without testing more projects if we should upgrade to junit v4.13 yet.

## Details

Bump Dependency to junit v4.13

## Motivation and Context

Still Bored

## How Has This Been Tested?

mvn clean install

## Types of changes

## Checklist:


mpkorstanje:
We're currently working on Cucumber 5.0 so now would be the best time do it. 

Going through the release notes I don't see anything obviously breaking.

https://github.com/junit-team/junit4/blob/master/doc/ReleaseNotes4.13.md

So if we forgo using features introduced in 4.13 we should be backwards compatible with 4.12.

mpkorstanje:
> Still Bored

Happy new year. :laughing:

Issue #1617, opened 2019-04-24
datatable-java: Empty cell are not converted to `null`'s for `String` class


gkalnytskyi:
## Summary
The empty cell is not converted to `(Double)null` in Cucumber 4.3.0, as it happened in Cucumber 1.2.5

In Cucumber 1.2.5 empty cell was converted to null of type `Double`, when performing `asMap(String,Double)` table conversion. This no longer happens in Cucumber 4.3.0. I also couldn't find the way to override this behavior because just adding new `DataTableType` definition results in `io.cucumber.datatable.DublicateTypeException`.

    typeRegistry.defineDataTableType(
        new DataTableType(
            Double.class,
            (TableCellTransformer<Double>) value ->
                (!value.isEmpty()) ? (Double.parseDouble(value)) : (null)));

## Expected Behavior
Empty cell is converted to `null` by default for nullable number types (`Double`, `Float`, `Int` etc.). So that the table in the following step 

    Then the marker values are
      | M1   | 1.7988 |
      | M2   |        |
      | M3   | 1.5312 |
      | M4   | 1.7720 |
      | M5   | 1.6840 |
      | M6   | 0.0000 |
      | M9   |        |
      | M12 | 1.7002 |

is transformed to

    {
        "M1": 1.7988,
        "M2": null,
        "M3": 1.5312,
        "M4": 1.7720,
        "M5": 1.6840,
        "M6": 0.0,
        "M9": null,
        "M12": 1.7002,
    }

## Current Behavior
Exception is thrown:
    io.cucumber.datatable.CucumberDataTableException: 'java.util.List<java.util.List<java.lang.Double>>' could not transform

## Possible Solution
Convert empty cells to null values, or allow to override default build-in type transformations.

## Steps to Reproduce (for bugs)

1. The table in the following step:

    Then the marker values are

      | M1   | 1.7988 |
      | M2   |        |
      | M3   | 1.5312 |
      | M4   | 1.7720 |
      | M5   | 1.6840 |
      | M6   | 0.0000 |
      | M9   |        |
      | M12 | 1.7002 |

2. Is trasformed to a `Map`

    @Then("Then the marker values are")
    public void theMarkerValuesAre(DataTable expectedValues) {
        Map<String, Double> expected = expectedValues.asMap(String.class, Double.class);
    }
3. `CucumberDataTableException` is thrown with message: _'java.util.List<java.util.List<java.lang.Double>>' could not transform_

## Context & Motivation
[...]

PR #1857, opened 2020-01-09
[Java] Support empty strings and null values in data tables

mpkorstanje:
## Summary

Empty data table cells can either be considered null or empty strings.
For example table below can converted to a map as
`{name=Aspiring Author, first publication=}` or
`{name=Aspiring Author, first publication=null}`.

```gherkin
 | name            | first publication |
 | Aspiring Author |                   |
```
## Motivation and Context

And as demonstrated #1617 there are good reasons to default the empty
table cell to null. However this does not cover all cases. There are
however good use cases to use the empty string.

Additionally allowing users to replace empty  table cells with a `[blank]` will
reduce the impact of  #1617 and provide a nicer migration pattern.

## Details

By declaring a table transformer with a replacement string it becomes
possible to explicitly disambiguate between the two scenarios. For
example:

```gherkin
Given some authors
   | name            | first publication |
   | Aspiring Author |                   |
   | Ancient Author  | [blank]           |
```

```java
@DataTableType(replaceWithEmptyString = "[blank]")
public Author convert(Map<String, String> entry){
  return new Author(
     entry.get("name"),
     entry.get("first publication")
  );
}

@Given("some authors")
public void given_some_authors(List<Author> authors){
  // authors = [Author(name="Aspiring Author", firstPublication=null), Author(name="Ancient Author", firstPublication=)]
}
```
## Todo
 - [x] Add `DataTable.getDataTableConverter` method so we can recreate data tables.

## Types of changes

## Checklist:


fslev:
Hi,
Is it possible for simillar behavior to apply to the Examples table ?

mpkorstanje:
How would that work?

fslev:
I believe something like this:
```gherkin
Feature: Test examples feature

  Scenario Template: Examples with empty values
    Given param p="<emptyValue>"
    And param x="<nullValue>"
    Examples:
      | emptyValue | nullValue |
      | [_blank]       |                 |
```
and I think 'replaceWithEmptyString' could be made available also for DefaultParameterTransformer:
```java
@DefaultParameterTransformer(replaceWithEmptyString = "[_blank]")
    public Object defaultTransformer(Object fromValue, Type toValueType) {
        return objectMapper.convertValue(fromValue, objectMapper.constructType(toValueType));
    }
```

mpkorstanje:
That's not what Scenario Outlines do.

```
Scenario Outline: I visit <website>
   Given I visit a <website>

   Examples: Social Media
     | website       | 
     | instagram.com |
     | snapchat.com  |

   Examples: Technology
     | website       | 
     | github.com    |
     | gitlab.com     |
```

Is syntactic sugar for:

```
Scenario: I visit  instagram.com 
   Given I visit a instagram.com

Scenario: I visit  snapchat.com 
   Given I visit a snapchat.com

Scenario: I visit  github.com 
   Given I visit a github.com

Scenario: I visit  gitlab.com 
   Given I visit a gitlab.com
```

> and I think 'replaceWithEmptyString' could be made available also for DefaultParameterTransformer:

If I'm not mistaken you can use an optional capture group to do this. 

```
Given param x=true
And param x=false
And param x=

@Given("^param x=(true|false)?$")
public void param(Boolean optional){
   // optional is either true, false or null
}
```

And with a non-capturing group you can 

```
Given param x="string"
And param x=""
And param x=null

@Given("^param x=(?:null)?(?:\"(.*)\")?$")
public void param(String optional){
   // optional is either "string", "", or null
}
```

fslev:
I know, I was referring to the values passed from the examples table to each scenario step parameters.
I was wondering if a generic solution for parsing empty/null strings existed in this case. Without re-defining each of my hundreds of step definitions, accordingly.
But I think the simplest solution is to write a little code inside my default parameter transformer, that treats empty strings received from the gherkin feature, either as empty or nulls

PR #1863, opened 2020-01-20
[TestNG] Run scenarios in customized groups

Dzieciak:
## Summary

By making `io.cucumber.testng.Pickle` implement `io.cucumber.core.gherkin.Pickle` and making `getFeatures()` private method in `io.cucumber.testng.TestNGCucumberRunner` idempotent you can define multiple TestNG data providers and with utilizing `org.testng.IDataProviderInterceptor` you can have more control on how Cucumber scenarios are executed.

For example:

If there is a set of scenarios and some of them cannot be run in parallel with others, there is no need to launch cucumber twice. Custom runner which extends `io.cucumber.testng.AbstractTestNGCucumberTests` can be implemented like:

    @CucumberOptions(
        strict = true,
        features = {
            "src/test/resources/features/"
        },
        plugin = {
            "pretty",
            "html:target/cucumber/report-html",
            "json:target/cucumber/report.json",
            "timeline:target/cucumber/timeline"
        },
        tags = {
            "@TestFeature"
        })
    @Listeners(CucumberDataProviderInterceptor.class)
    public class CucumberRunner extends AbstractTestNGCucumberTests {

        @Override
        @DataProvider(parallel = true)
        public Object[][] scenarios() {
            return super.scenarios();
        }

        @DataProvider
        public Object[][] notParallelScenarios() {
            return super.scenarios();
        }
    
        @Test(groups = "cucumber", description = "Runs @NotParallel scenarios", dataProvider = "notParallelScenarios")
        public void runNotParallelScenarios(PickleWrapper pickleWrapper, FeatureWrapper featureWrapper) throws Throwable {
            super.runScenario(pickleWrapper, featureWrapper);
        }

    }

where in `CucumberDataProviderInterceptor` scenario filtering can be done basing on e.g. tags and data provider name.

## Motivation and Context

Have a possibility to run scenarios in user defined groups where scenarios within one group can be run in parallel or sequentially in one Cucumber run. The result can look like in attached screenshot where scenarios were split into three groups. Two were run in parallel and one sequentially.

timeline

## Types of changes

## Checklist:


testng/src/main/java/io/cucumber/testng/Pickle.java
```diff
@@ -2,11 +2,17 @@
...
+
+import io.cucumber.core.gherkin.Location;
+import io.cucumber.core.gherkin.Step;
+
 /**
  * Wraps CucumberPickle to avoid exposing it as part of the public api.
  */
 @API(status = API.Status.STABLE)
-public final class Pickle {
+public final class Pickle implements io.cucumber.core.gherkin.Pickle {
```
mpkorstanje: By implementing ` io.cucumber.core.gherkin.Pickle` we're exposing internal parts of the API to external users.  I don't think this interface here is needed to add all the methods so we can remove this.

mpkorstanje: Additionally I don't think all information should be exposed. For now I'd like to limit this to `name`, `tags`, `uri` and `scenarioLine`.

testng/src/main/java/io/cucumber/testng/TestNGCucumberRunner.java
```diff
@@ -150,12 +152,11 @@ public void finish() {
     }
 
     private List<Feature> getFeatures() {
```
mpkorstanje: Documenting that this method is idempotent would be nice.

testng/src/main/java/io/cucumber/testng/TestNGCucumberRunner.java
```diff
@@ -62,6 +63,7 @@
     private final RuntimeOptions runtimeOptions;
     private final Plugins plugins;
     private final FeaturePathFeatureSupplier featureSupplier;
+    private final List<Feature> features = new ArrayList<>();
```
mpkorstanje: This should be initialized as null. There is a difference between not initialized and no feature found.

testng/src/main/java/io/cucumber/testng/TestNGCucumberRunner.java
mpkorstanje: A test would be nice too.

testng/src/main/java/io/cucumber/testng/Pickle.java
Dzieciak: I updated PR accordingly.

testng/src/main/java/io/cucumber/testng/TestNGCucumberRunner.java
Dzieciak: I have a hard time with coming up with a test for this particular method since it's a private one. I added a test which demonstrates whole solution (multiple data providers and filtering scenarios)

testng/src/main/java/io/cucumber/testng/TestNGCucumberRunner.java
Dzieciak: Updated accordingly

testng/src/main/java/io/cucumber/testng/TestNGCucumberRunner.java
mpkorstanje: That's good too. It will help document stuff.

mpkorstanje:
Found a clearer way to do this. Doesn't rely on the interceptor but rather filters the pickles provided by the `TestNGCucumberRunner` directly.

```java
 package io.cucumber.testng;

 import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;

 import java.util.Arrays;
 import java.util.function.Predicate;

 @CucumberOptions(features = "classpath:io/cucumber/testng/scenarios_with_tags.feature", plugin = "timeline:target/timeline")
 public class ScenariosInDifferentGroupsTest {

     private static final Predicate<Pickle> isSerial = pickle -> pickle.getTags().contains("@serial");

     private TestNGCucumberRunner testNGCucumberRunner;

     @BeforeClass(alwaysRun = true)
     public void setUpClass() {
         testNGCucumberRunner = new TestNGCucumberRunner(this.getClass());
     }

     @test(groups = "cucumber", description = "Runs Cucumber Scenarios", dataProvider = "parallelScenarios")
     public void runParallelScenario(PickleWrapper pickleWrapper, FeatureWrapper featureWrapper) throws Throwable {
         testNGCucumberRunner.runScenario(pickleWrapper.getPickle());
     }

     @dataProvider(parallel = true)
     public Object[][] parallelScenarios() {
         if (testNGCucumberRunner == null) {
             return new Object[0][0];
         }
         return filter(testNGCucumberRunner.provideScenarios(), isSerial.negate());
     }

     @test(groups = "cucumber", description = "Runs Cucumber Scenarios in the Serial group", dataProvider = "serialScenarios")
     public void runSerialScenario(PickleWrapper pickleWrapper, FeatureWrapper featureWrapper) throws Throwable {
         testNGCucumberRunner.runScenario(pickleWrapper.getPickle());
     }

     @dataProvider(parallel = true)
     public Object[][] serialScenarios() {
         if (testNGCucumberRunner == null) {
             return new Object[0][0];
         }

         return filter(testNGCucumberRunner.provideScenarios(), isSerial);
     }

     @afterclass(alwaysRun = true)
     public void tearDownClass() {
         if (testNGCucumberRunner == null) {
             return;
         }
         testNGCucumberRunner.finish();
     }

     private Object[][] filter(Object[][] scenarios, Predicate<Pickle> accept) {
         return Arrays.stream(scenarios).filter(objects -> {
             PickleWrapper candidate = (PickleWrapper) objects[0];
             return candidate == null || accept.test(candidate.getPickle());
         }).toArray(Object[][]::new);
     }
}
```

PR #1867, opened 2020-01-23
[JUnit] Add module-info for cucumber-junit-platform-engine

mpkorstanje:
## Summary

Adds module-info for Cucumber Junit Platform Engine implementation.
This allows Java to ensure that all dependencies are available at
boot time.

To ensure backwards compatibility down to Java 8 we are using
multi-release jar files which will contain both Java 8 and
Java 9 class files.

To ensure the right JDK (OpenJDK 11) is used to build the project
the Maven Toolchains plugin will verify that this JDK is used.
This does require adding a `toolchains.xml` file to the `~/.m2/`
folder.

Example:

```xml
<?xml version="1.0" encoding="UTF8"?>
<toolchains>
    <toolchain>
        <type>jdk</type>
        <provides>
            <version>11</version>
            <vendor>openjdk</vendor>
        </provides>
        <configuration>
            <jdkHome>/path/to/openjdk11</jdkHome>
        </configuration>
    </toolchain>
</toolchains>
```

### DIFF
--- a/.m2/travis-ci-toolchains.xml
+++ b/.m2/travis-ci-toolchains.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF8"?>
+<toolchains>
+    <toolchain>
+        <type>jdk</type>
+        <provides>
+            <version>11</version>
+            <vendor>openjdk</vendor>
+        </provides>
+        <configuration>
+            <jdkHome>/usr/local/lib/jvm/openjdk11</jdkHome>
+        </configuration>
+    </toolchain>
+</toolchains>
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,33 +3,37 @@ sudo: false
 
 cache:
   directories:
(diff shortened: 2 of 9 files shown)


mpkorstanje:
No `tool-chain.xml` provided by Travis CI. 

https://github.com/travis-ci/travis-ci/issues/2727

mpkorstanje:
https://travis-ci.community/t/add-support-for-maven-toolchain-plugin/5004
> it is extremely difficult to maintain toolchains by project in travis for the simple reason that the locations of the installed jdks are not obvious and they do keep changing.

Looks like it is time to drop Travis for something else.

mpkorstanje:
The first `mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V` doesn't help either because travis is not providing tool chains.

https://stackoverflow.com/questions/31945809/skip-first-mvn-install-in-travis-ci

PR #1868, opened 2020-01-25
bug: on windows machine you can't determine the resource uri of inner packages

PeretzNadav:
when running from windows machine  relativize method return the path of sub packages as the native windows seperator path config \
so when you try to build uri app crash.
 i can give example from my own env:

[Utils] [ERROR] [Error] java.lang.IllegalArgumentException: Illegal character in opaque part at index 23: classpath:features/Casb\ActivitiesIntake/CASB-****-********permission_changesTo_delta_API.feature
	at java.base/java.net.URI.create(URI.java:906)

you can see that the seperator change between Casb folder and ActivitiesIntake folder

can you please fix that issue?(it's actually a bug, on windows machine you cant determine the resource uri of inner packages) 
i suggest also a fix to that problem on line 88 (  .replacAll("\\","/"); ) but you migh want to do something smarter

thanks

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/java/io/cucumber/core/resource/ClasspathSupport.java
+++ b/core/src/main/java/io/cucumber/core/resource/ClasspathSupport.java
@@ -85,7 +85,7 @@ public final class ClasspathSupport {
     }
 
     static URI determineClasspathResourceUri(Path baseDir, String basePackagePath, Path resource) {
-        String subPackageName = baseDir.relativize(resource.getParent()).toString();
+        String subPackageName = baseDir.relativize(resource.getParent()).toString().replacAll("\\","/");
         String resourceName = resource.getFileName().toString();
         String classpathResourcePath = of(basePackagePath, subPackageName, resourceName)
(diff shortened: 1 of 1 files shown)


mpkorstanje:
Thanks! If you want you can checkout the latest snapshot by checking out master and build that using `mvn clean install`.

mpkorstanje:
Or you can use v5.1.0. It should be in Maven central in a few hours.

PeretzNadav:
Thanks :) 

can i ask you question not relate to that issue???

i just finish to build my own plugin (based on JSONFormatter) just for give the ability to show
reports on runtime...
i actually build the cucmber.json after each FinishedStep event and then generate report.. 
so i override the same json and this is what make it publish report on runtime,
instead of waiting couple hours for answer in order to figure what happened  

do you familiar with some other technique in order to achieve that goal???
i prefer to take something embedded and not to maintain my plugin everytime i change cucumber version...

again thanks a lot for the fix .. i need it :)

Issue #1706, opened 2019-07-22
Enhancement request --retry flag is still not available


hemanthsridhar:
Currently the cucumber Java CLI does not provide us the --retry option 
Can you guys please integrate it? 
If there's any issue can we have an alternate solution that we can do so that it's almost the same as --retry

mpkorstanje:
From the README.md

> Please bear in mind that this project is almost entirely developed by volunteers. If you do not provide the implementation yourself (or pay someone to do it for you), the bug might never get fixed. If it is a serious bug, other people than you might care enough to provide a fix.

And if you don't mind using only surefire reports:

https://maven.apache.org/surefire/maven-surefire-plugin/examples/rerun-failing-tests.html

hemanthsridhar:
Is there a way to do the same using gradle

mpkorstanje:
No. Gradles JUnit runner implemetation is quite primitive. I don't think it looks below the class level at all.

Anyway there is a good chance that implementing rerun functionality is more work then making your tests stable. Otherwise there would have been a contribution by now.

If you are using a Webdriver look up articles about writing stable tests with that. Otherwise look up articles about making tests stable in general. It's not hard, just a few patterns to follow and habits to break.

hemanthsridhar:
Thank you @mpkorstanje

hemanthsridhar:
@mpkorstanje 
I tried with your suggestion. Migrated to Maven and used the surefire way of rerunning. However now, I will not be able to run scenarios in parallel.
Is there a way to to use --threads from cucumber cli and use the retry option?

mpkorstanje:
Surefire can execute tests in parallel. Look up the threads option.

hemanthsridhar:
@mpkorstanje 
Yes. But it executes feature files in parallel not the scenarios. Is there a way to execute the scenarios of a single feature in parallel.

I am able to do it using cucumber-testng but when I migrate to cucumber-testng immediately the retry feature of the surefire vanishes.

mpkorstanje:
These are unfortunately the constraints of the tools we're working with. I believe TestNG also has a retry mechanism build in. Have you tried to use that?

hemanthsridhar:
@mpkorstanje The IRetryListener doesn't get executed unfortunately.

camohub:
[...]

PR #1886, opened 2020-02-04
#1706 - WIP - add feature retries

nddipiazza:
## Summary

Add a retry feature.

## Details

Should be able to retry using a --retry parameter, or a Retry annotation.

## Motivation and Context

https://github.com/cucumber/cucumber-jvm/issues/1706

## How Has This Been Tested?

Pending test coverage.

## Types of changes

## Checklist:

- [ ] I've added tests for my code. TODO
- [ ] I have updated the documentation accordingly. TODO

### DIFF
--- a/core/src/main/java/cucumber/api/CucumberOptions.java
+++ b/core/src/main/java/cucumber/api/CucumberOptions.java
@@ -99,5 +99,9 @@ public @interface CucumberOptions {
      * Pass options to the JUnit runner.
      */
     String[] junit() default {};
-
+    
+    /**
+     * Number of times to retry a scenario before giving up.
+     */
+    int retry() default 0;
 }
--- a/core/src/main/java/io/cucumber/core/options/CucumberOptionsAnnotationParser.java
+++ b/core/src/main/java/io/cucumber/core/options/CucumberOptionsAnnotationParser.java
@@ -285,10 +285,10 @@ public final class CucumberOptionsAnnotationParser {
             return annotation.junit();
         }
 
-		@Override
-		public int retry() {
-			return annotation.retry();
-		}
+        @Override
+        public int retry() {
+            return annotation.retry();
+        }
     }
 
     private static class CoreCucumberOptionsProvider implements OptionsProvider {
--- a/core/src/main/java/io/cucumber/core/options/CucumberOptionsAnnotationParser.java
+++ b/core/src/main/java/io/cucumber/core/options/CucumberOptionsAnnotationParser.java
@@ -55,6 +55,7 @@ public final class CucumberOptionsAnnotationParser {
                 addGlue(options, args);
                 addFeatures(options, args);
                 addJunitOptions(options, args);
+                addRetry(options, args);
             }
         }
         addDefaultFeaturePathIfNoFeaturePathIsSpecified(args, clazz);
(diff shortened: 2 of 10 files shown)


mpkorstanje:
I appreciate that you're creating a PR for a new feature. However we request that you open an issues to discuss significant contributions prior to making them.

This would have allowed me to point out that the reason `IRetryListener` doesn't work in TestNG due to https://github.com/cbeust/testng/issues/891. And I would prefer that you work on fixing this issue rather then build it into the TestNG and JUNit integrations.

Additionally Cucumber-JVM currently doesn't support reporting on retried tests at all. So merely adding functionality to rerun tests until passed will not be sufficient to accept a PR. There is functionality coming up to make this possible but that won't be available the new Message protocol has been implemented (#1841).

nddipiazza:
Cool your final comment will be sufficient for my purposes. 
I will work on https://github.com/cbeust/testng/issues/891 as time allows.
I'll also keep an eye on https://github.com/cucumber/cucumber-jvm/pull/1841

nddipiazza:
> Finally it would also have allowed me to point out that if you are using TestNG and aren't interested in fixing cbeust/testng#891 nor willing to wait for #1841 you can also extend AbstractTestNGCucumberTests and invoke runScenario until the tests passes.

@mpkorstanje question for you about this comment... let's say I do what you suggested here and then it passes on the 2nd+ try. Will the cucumber-report have a single successful attempt? or will the cucumber-report get messed up?

mpkorstanje:
I honestly I don't know. It will very much depend on the way a particular reporting plugin was written. However given that currently there is no way to differentiate between the events from the original execution and the events from any subsequent rerun, I suspect it won't go well.

nddipiazza:
@mpkorstanje  I'm currently using https://mvnrepository.com/artifact/net.masterthought/cucumber-reporting/4.11.2

What I can try to do if necessary is run test, some errors occur. we then parse the testng test output and initiate retry loop where appropriate. and if the test passes, inject the cucumber output with the passing test to replace the previously failing test and regenerate the report html. Meh?

mpkorstanje:
Ideally your reporter would understand that a test was rerun and report it as such. Currently only the native TestNG and JUnit reporters can do that. So you'd have to rely on those reports rather then anything generated by Cucumber or based on Cucumbers output.

So if you ask me what you can try: I'd spend my time making tests more stable.

nddipiazza:
> I'd spend my time making tests more stable.

this is the classic response to the flaky test retry situation. and in my case, there is a mainframe system that causes intermittent test failures completely outside of my control.

mpkorstanje:
Been there done that.

https://github.com/cucumber/cucumber-jvm/issues/1120

In hindsight it was the biggest waste of my time. 😁

nddipiazza:
@mpkorstanje this works great. I noticed that it still has a json record of the failures prior to the passing retry however. Is there some way to "erase" the previous failed attempts prior to letting the scenario pass? I am looking into the source code of the cucumber-reporting github project as well to see if I can do this during the generation of the pretty HTML report.

nddipiazza:
@mpkorstanje  never mind my previous question. i went ahead and added this ability into the cucumber-reporting in a PR. 
see 
https://github.com/damianszczepanik/cucumber-reporting/pull/914

mpkorstanje:
Ooh. Well done.

Issue #1870, opened 2020-01-25
Add FeatureParsedEvent


mpkorstanje:
The current events in the `plugin` module only provide information about pickles. However most plugins need some information about the hierarchy pickles are in.

Currently internal plugins (e.g. `TeamCityPlugin` and `JUnitFormatter`) work around this by using the `FeatureParser`. We should however make this information available via an event.

The easiest way would probably to extend the interfaces from the `gherkin` module with identical interfaces in the `plugin` module and then try to minimize the available methods to the bare minimum. I.e:

```java
public final class FeatureParsed extends TimeStampedEvent {
    private final Feature feature;

    public FeatureParsed(Instant timeInstant, Feature feature) {
        super(timeInstant);
        this.feature = Objects.requireNonNull(feature);
    }

    public String getFeature() {
        return feature;
    }

}

public interface Feature extends Node, Container<Node> {
    URI getUri();

    String getSource();

}
```

@rasklaad thoughts?

mpkorstanje:
Consider adding tags to nodes.

rasklaad:
Sry for the late comment.
Should we rewrite internal plugins than to use FeatureParsedEvent? I guess it would be the nice example  for plugin writers.

mpkorstanje:
Yeah. Already did. Not for the HTML and JSON formatter because they reproduce a facimile of the feature file which can only be done with the AST. The purpose of the `TestSourceParsed` event is to provide a abstract tree of the hierarchy in the test source.

 But I'm not happy about the structure yet. There is this really annoying way to unpack the `Container`.

```
   private void handleTestSourceParsed(TestSourceParsed event) {
        TestSourceReadResource source = new TestSourceReadResource(event);
        Node.Container<Node> container = event.getContainer();
        parser.parseResource(source).ifPresent(feature ->
        if (container instanceof Node) {
            featuresNames.put(feature.getUri(), feature.getName())
            Node node = (Node) container;
        );
            featuresNames.put(event.getUri(), node.getName());
        } else {
            featuresNames.put(event.getUri(), event.getUri().toString());
        }
    }
    }
```
I introduced the container because feature files don't have to be the root of the abstract tree. It's possible that we may introduce multiple feature files or scenarios that aren't part of a feature file. 

So perhaps the root container should never be a node itself.

PR #1895, opened 2020-02-14
[Plugin] Add TestSourceParsed event

mpkorstanje:
## Summary

Cucumber scenarios and individual examples in a scenario outline are compiled
into pickles. These pickles are wrapped by a `TestCase`. As such
Cucumbers internal representation lacks any hierarchy. I.e. once compiled
into a a pickle a scenario is no longer associated with a feature file.

However consumers of Cucumbers output generally expect results to be reported
in hierarchical fashion. This event allows test cases to be associated with
 with a `Node` in the hierarchy. The `TestSourceParsed` event provides this information. 

A node is an element in a tree. It has a name, keyword and location. A node can optionally be container of yet more nodes. 

```
└── TestSourceParsed
      ├── Node
      └── Container
      │ ├── Node
      │ └── Node
      └── Container
        └── Container
           ├── Node
           └── Node
```

Note that this information is intentionally abstract. To recreate a a facsimile of the original feature file it is still recommended to use the AST produced by a Gherkin parser. 

Closes: #1870

## Types of changes

## Checklist:


mpkorstanje:
Note: Document that people should shade the gherkin parser if they need the AST.

PR #1898, opened 2020-02-19
javascript write function can be write XML text

sgrillon14:
## Summary

I want write a XML text in report.

## Details

all xml balise are not visible in html report before this PR.

## Motivation and Context
```
import cucumber.api.Scenario;

public class MyStepdefs {
    private Scenario scenario;

    @Before
    public void before(Scenario scenario) {
        this.scenario = scenario;
    }

    @Given("^I print a text in Cucumber report$")
    public void print() {
        scenario.write("<AppHdr xmlns="urn:iso:std:iso:20022:tech:xsd:head.001.001.01">\r\n<Fr>\r\n<FIId>\r\n<FinInstnId>\r\n<BICFI>TRGTXE2SXXX</BICFI>\r\n<Othr>\r\n<Id>RSETFRPPXXX</Id>\r\n</Othr>\r\n</FinInstnId>\r\n</FIId>\r\n</Fr>\r\n</AppHdr>\r\n");
    }
}
```
in `report.js` file:
```
formatter.write("SAVE message4sese23FreeRece\u003d\u003cAppHdr xmlns\u003d\"urn:iso:std:iso:20022:tech:xsd:head.001.001.01\"\u003e\r\n      \u003cFr\u003e\r\n         \u003cFIId\u003e\r\n            \u003cFinInstnId\u003e\r\n               \u003cBICFI\u003eTRGTXE2SXXX\u003c/BICFI\u003e\r\n               \u003cOthr\u003e\r\n                  \u003cId\u003eRSETFRPPXXX\u003c/Id\u003e\r\n               \u003c/Othr\u003e\r\n            \u003c/FinInstnId\u003e\r\n         \u003c/FIId\u003e\r\n      \u003c/Fr\u003e   ....... \u003e");
```

## Screenshots :

image

after this PR:

## Types of changes

## Checklist:

### DIFF
--- a/core/src/main/resources/io/cucumber/core/plugin/html/formatter.js
+++ b/core/src/main/resources/io/cucumber/core/plugin/html/formatter.js
@@ -121,7 +121,7 @@ CucumberHTML.DOMFormatter = function(rootNode) {
     if (currentStepIndex == 1) {
       this.dummyStep();
     }
-    currentStep.append('<pre class="embedded-text">' + text + '</pre>');
+    currentStep.append('<pre class="embedded-text">' + String(text).replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;').replace(/"/g, '&quot;') + '</pre>');
   };
 
   this.before = function(before) {


core/src/main/resources/io/cucumber/core/plugin/html/formatter.js
```diff
@@ -121,7 +121,7 @@ CucumberHTML.DOMFormatter = function(rootNode) {
     if (currentStepIndex == 1) {
       this.dummyStep();
     }
-    currentStep.append('<pre class="embedded-text">' + text + '</pre>');
+    currentStep.append('<pre class="embedded-text">' + String(text).replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;').replace(/"/g, '&quot;') + '</pre>');
```
mpkorstanje: My javascript is a little bit rusty but shouldn't this be done using the `document.createTextNode(text)` instead? I realize the current implementation is using string concatenation to create the embedded text everywhere but that seems like a bad idea all together.

sgrillon14: I try `currentStep.append('<pre class="embedded-text">' + $(text).text() + '</pre>');`

I try `currentStep.append('<pre class="embedded-text">' + document.createTextNode(text) + '</pre>');`

but do not work.

sgrillon14: "we'll eventually deprecate this html formatter."  => there will be no more html report?

mpkorstanje: > "we'll eventually deprecate this html formatter." => there will be no more html report?

I'm not sure yet how that will play out yet. The current html formatter can't handle the rule key word that was introduced into Gherkin a while ago and the current html formatter is decrepit to the point that's not worth maintaining.

We're working on a new html formatter written in node that uses react. Its currently a stand alone process, if we can turn it into single page application we'll replace the current html formatter with that.

Just wanted to let you know that trying to patch this up it may not be the best investment of your time.

mpkorstanje: > I try currentStep.append('<pre class="embedded-text">' + $(text).text() + '</pre>');

In both cases you're trying to concatenate a string and a html node. I believe you may have to get hold of the `pre` node and append the text node to it's children.

mpkorstanje:
We've got a working Html formatter on the `v6.x.x` branch. It is significantly better then the current one. So I'd rather not spend any more time on it.

PR #1910, opened 2020-03-04
[Core] Support limited set of cURL commands for UrlOutputStream

aslakhellesoy:
## Summary

Gives users more fine grained control over how formatter output is sent via http(s). By default the `UrlOutputStream` will make a post request to a given url. By specifying `--plugin "message:https://example.com/gherkin/upload -X PUT -H 'Content-Type: application/json' -H 'X-Access-Token: Y3VjdW1iZXI'"` the following modifications are made

* Sets `content-type: application/json`
* Overrides HTTP method to `PUT`
* Adds additional access token header.

## Motivation and Context

We're building a new hosted service where Cucumber users can send their results, we need to provide users with a way to set security related HTTP headers (e.g. `Authorization: token TOKEN`).

## Types of changes

## Checklist:


mpkorstanje:
Rather then reusing the query parameters I'd like to prepend the additional information. Some ideas: 

```
Variant 1: key=value pairs, delimited by +

 - ndjson:method=PUT+https://example.com/gherkin/upload
 - ndjson:method=PUT+access_token=Y3VjdW1iZXI=+https://example.com/gherkin/upload

Variant 2: key=value pairs, delimited by space
 - ndjson:method=PUT https://example.com/gherkin/upload
 - ndjson:method=PUT access_token=Y3VjdW1iZXI https://example.com/gherkin/upload

Variant 3: key=value pairs, delimited by space with optional quotes

 - ndjson:method=PUT "header with spaces"="hello world!" https://example.com/gherkin/upload
 - ndjson:method=PUT access_token=Y3VjdW1iZXI https://example.com/gherkin/upload

Variant 4: key=value pairs, everything else assumed to be a curl header.

 - ndjson:method=PUT "access_token: Y3VjdW1iZXI" https://example.com/gherkin/upload

Variant 5: Curl CLI commands

 - ndjson:--X PUT -H "access_token: Y3VjdW1iZXI" https://example.com/gherkin/upload

```

I think the last one is most readable.

mpkorstanje:
I've removed the explanation from the change log section. This results in frequent conflicts because every change touches this file.

mpkorstanje:
I've made it work.

I changed the plugin string format from `PLUGIN[:[PATH|[URL]]` to `PLUGIN[:[PATH|[URI [OPTIONS]]]`

Because `PATH` still has to support paths with spaces the curl command must start with either the `URI` or `PATH`. After this the `-X` and `-H` options can be used. We can discover these because unlike URLs URIs may not contain spaces.

Issue #1896, opened 2020-02-14
Remove the neeed for the dummy method when using cucumber-spring


mpkorstanje:
**Is your feature request related to a problem? Please describe.**

Currently a dummy method is needed to ensure that spring configuration is picked up when no step definitions are present. 

```
@SpringBootTest
@AutoConfigureMockMvc
public class CucumberContextConfiguration  {

    @Before
    public void setup_cucumber_spring_context(){
        // Dummy method so cucumber will recognize this class as glue
        // and use its context configuration.
    }
}
```

**Describe the solution you'd like**

```
@CucumberSpringTest
@SpringBootTest
@AutoConfigureMockMvc
public class CucumberContextConfiguration  {

}
```

This can be done by adding a minimal implementation of `Backend` to `cucumber-spring` that scans for `@CucumberSpringTest` and adds it to the glue.

PR #1911, opened 2020-03-04
[Spring] Add @CucumberContextConfiguration annotation

rasklaad:
## Summary

To make Cucumber aware of your test configuration you can annotate a
configuration class with `@CucumberContextConfiguration` and with one of the
following annotations: `@ContextConfiguration`, `@ContextHierarchy` or
`@BootstrapWith`. If you are using SpringBoot, you can annotate configuration
class with `@SpringBootTest(classes = TestConfig.class)`.

## Details

- [x] Spring backend minimum implementation
- [x] Spring backend tests
- [x] Docs/javadocs

## Motivation and Context
Fixes: https://github.com/cucumber/cucumber-jvm/issues/1896

## Types of changes

## Checklist:


mpkorstanje:
Don't worry about `revapi` failing the build right now. It's a semantic version checker. You can ignore it, I'll adjust the version numbers.

spring/src/main/java/io/cucumber/spring/SpringBackend.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public void disposeWorld() {
+
+    }
+
+    @Override
+    public Snippet getSnippet() {
+        return null;
```
mpkorstanje: We should make sure the caller in `Runner.java` filters out the null values.

spring/src/main/java/io/cucumber/spring/CucumberSpringTest.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@API(status = API.Status.STABLE)
+public @interface CucumberSpringTest {
```
mpkorstanje: I suggested `@CucumberSpringTest` but I'm not sure if that is the right name. This annotation allows a user to mark a application context configuration class as the class that Cucumber should use. Maybe different name can convey this better.

rasklaad: What about `@CucumberContextConfiguration` ?

mpkorstanje: That would work.

mpkorstanje:
And we're done. Thanks you very much!

PR #1918, opened 2020-03-10
Html message formatter

aslakhellesoy:
HTML Formatter based on https://github.com/cucumber/cucumber/tree/master/html-formatter


core/src/test/java/io/cucumber/core/plugin/HTMLFormatterTest.java
```diff
@@ -1,795 +1,795 @@
...
-            .withHooks(hooks)
-            .withHookLocations(hookLocations)
-            .withHookActions(hookActions)
-            .withTimeServiceIncrement(stepDuration)
-            .build()
-            .run();
-    }
-
-}
+//package io.cucumber.core.plugin;
```
mpkorstanje: Should probably delete this or replace it with an actual test.

PR #1935, opened 2020-04-03
[All] Use Jupiter Assertions over JUnit4 Assert

timtebeek:
## Summary

Use Jupiter Assertions over JUnit4 Assert. In particular deltaspike seems to have had a few old style Asserts. Jupiter should be preferred where possible, unless we're verifying compatibility such as in the junit module.

## Types of changes

## Checklist:

### DIFF
--- a/spring/src/test/java/io/cucumber/spring/contextcaching/SomeTest.java
+++ b/spring/src/test/java/io/cucumber/spring/contextcaching/SomeTest.java
@@ -1,7 +1,6 @@
 package io.cucumber.spring.contextcaching;
 
-import org.hamcrest.MatcherAssert;
-import org.junit.jupiter.api.Test;
+import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.test.context.ContextConfiguration;
@@ -12,12 +11,12 @@ import static org.hamcrest.MatcherAssert.assertThat;
 
 @RunWith(SpringJUnit4ClassRunner.class)
 @ContextConfiguration(classes = {ContextConfig.class})
-class SomeTest {
+public class SomeTest {
     @Autowired
     ContextCounter contextCounter;
 
     @Test
-    void contextCountIsOne() {
+    public void contextCountIsOne() {
         assertThat(contextCounter.getContextCount(), is(1));
     }
 
--- a/spring/src/test/java/io/cucumber/spring/contextcaching/SomeTest.java
+++ b/spring/src/test/java/io/cucumber/spring/contextcaching/SomeTest.java
@@ -1,7 +1,7 @@
 package io.cucumber.spring.contextcaching;
 
 import org.hamcrest.MatcherAssert;
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
 import org.junit.runner.RunWith;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.test.context.ContextConfiguration;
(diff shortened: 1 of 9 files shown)


timtebeek:
@mpkorstanje wrote myself a script yesterday to convert some work projects from JUnit4 to Jupiter; thought it couldn't hurt to run it on cucumber-jvm as well. Is it worth trying to make this work by changing the archetype dependencies to use jupiter?

spring/src/test/java/io/cucumber/spring/contextcaching/SomeTest.java
```diff
@@ -12,12 +12,12 @@
 
 @RunWith(SpringJUnit4ClassRunner.class)
 @ContextConfiguration(classes = {ContextConfig.class})
-public class SomeTest {
+class SomeTest {
```
mpkorstanje, line 15: `@RunWith` is a JUnit 4 annotation. So I don't think this will work as expected.

mpkorstanje:
>  Is it worth trying to make this work by changing the archetype dependencies to use jupiter?

Not yet. The archetype is used in our documentation as part of the 10-minute tutorial. The tutorial is aimed at extremely new developers. There is little very tolerance for platform and tooling related problems and tooling support for Junit Platform hasn't quite reached the point yet where I'd be comfortable changing that.

Internally Surefire, Gradle assume the old class based execution model. This means that Cucumber works less then stellar with these integrations. So I'm currently waiting for a few issues to be resolved here:
 * https://issues.apache.org/jira/browse/SUREFIRE-1724
 * https://github.com/gradle/gradle/issues/4773

With respect to IDEA, there is no good way to use the JUnit Platform to run individual scenarios and so I'm waiting for:
 * https://youtrack.jetbrains.com/issue/IDEA-227508
 * https://github.com/junit-team/junit5/issues/2146 

You may find some enjoyment in trying solve https://github.com/junit-team/junit5/issues/2146. JUnit 5s code base has amazing quality.

mpkorstanje:
> I wrote myself a script yesterday to convert some work projects from JUnit4 to Jupiter; 

This should also picks up the Spring related `@RunWith` annotations.

https://github.com/sbrannen/junit-converters

timtebeek:
> > I wrote myself a script yesterday to convert some work projects from JUnit4 to Jupiter;
> 
> This should also picks up the Spring related `@RunWith` annotations.
> 
> https://github.com/sbrannen/junit-converters

Yes thanks; I had seen you star that repository a while back ; I adapted a bit from that yesterday, eventually ending up with: https://gist.github.com/timtebeek/51d0988397b9e9deac9a018d297e88d5

Some of it is a bit specific to the job, and it's not the cleanest code, but it gets the job done.

timtebeek:
> > Is it worth trying to make this work by changing the archetype dependencies to use jupiter?
> 
> Not yet. The archetype is used in our documentation as part of the 10-minute tutorial. The tutorial is aimed at extremely new developers. There is little very tolerance for platform and tooling related problems and tooling support for Junit Platform hasn't quite reached the point yet where I'd be comfortable changing that.

Understandable; if I take out the archetype changes we're left with changes in deltaspike, an possibly some in spring/src/test/java/io/cucumber/spring/contextcaching/SomeTest.java which doesn't yet use the new `CucumberContextConfiguration`; should I distill this PR down to just those changes?

Diving into JUnit5 might seem fun, but not sure if I'll find the time easily; also, the package access challenge posed in the issue you raised does not seem to have a clear resolve yet. I'll explore if I find time. :)

mpkorstanje:
Sure every bit helps!

timtebeek:
@mpkorstanje Deltaspike changes ready for review/merge now then. :)

mpkorstanje:
Cheers!

I fixed up spring too. `RunWith` could be replaced with `ExtendsWith` with no issue. I guess this may not have been quite clear given the class was named `SomeTest`.

PR #1942, opened 2020-04-10
[Core] Add parameter type events

mpkorstanje:
* Adds parameter type defined events for the CCK.
 * Adds exception to `TestRunFinished` event.
 * Adds error state json to `JsonFormatter` output.
 * Add exception handling around Cucumber execution. When the Engine fails, Cucumber will still emit the `TestRunFinished` event.

## Types of changes

## Checklist:


mpkorstanje:
Not passing because the resolution mentioned in https://github.com/cucumber/cucumber/pull/890 has not yet been implemented.

mpkorstanje:
Should probably split this into a few PRs. There are some re-factorings and features mixed together.

plugin/src/main/java/io/cucumber/plugin/event/TestRunFinished.java
```diff
@@ -7,7 +7,18 @@
 @API(status = API.Status.STABLE)
 public final class TestRunFinished extends TimeStampedEvent {
 
+    private final Exception exception;
```
mpkorstanje, line 10: Wrap this in a result object. Add state.

Consider doing the same for the messages.

mpkorstanje:
Check if existing reporters can understand the failure json.

pom.xml
```diff
@@ -51,6 +51,14 @@
...
         <tag-expressions.version>2.0.4</tag-expressions.version>
+        <!--
+        IMPORTANT!
+        Whenever messages.version is updated, also update the files from the compatibility kit:
+        cd ../cucumber # the monorepo
+        git checkout messages/vMAJOR.MINOR.PATCH
+        cd ../cucumber-jvm
+        ./compatibility/copy-cck-expectations.sh
+        -->
         <messages.version>12.0.0</messages.version>
```
mpkorstanje, line 62: This includes the new error messages. I'm assuming the parser doesn't know this and we don't have test for it.

aslakhellesoy: The parser does know about it. Fixed in cd2e019716f1f49fe260fe85e53f5d08a953ea66

PR #1945, opened 2020-04-15
Update dependencies

aslakhellesoy:


aslakhellesoy:
This failed the semantic versioning check [in gherkin-messages](https://travis-ci.org/github/cucumber/cucumber-jvm/jobs/675269142#L7468).

I'm not sure how I feel about that. I'm struggling to see how this change affects users in a way that would warrant a major version bump.

WDYT @mpkorstanje?

mpkorstanje:
Right now, it only seems to affect `gherkin-messages` which is an internal module and not yet used. But there also are changes to to messages themselves which would be a breaking change. 

So I think the check is useful. For now lets target the v6.x.x branch.

mpkorstanje:
If I try to resolve the conflicts in v6.x.x there are no changes left.

What are you trying to achieve by updating the dependencies in v5 right now?

PR #1951, opened 2020-04-21
[Java8] Fix dependency name typo

natritmeyer:
The `README.md` for `cucumber-java8` informs the user:

    To use add the `cucumber-java` dependency

...whereas it should say...

    To use add the `cucumber-java8` dependency

That's all this PR fixes :)


mpkorstanje:
Cheers!

Issue #1788, opened 2019-10-04
Make `--strict` mode default


mpkorstanje:
Depending on the outcome of the discussion in https://github.com/cucumber/cucumber/issues/714 we should make `--strict` mode default and possibly remove it all together from Cucumber.

This will make integration with other frameworks and reporting tools easier as they will only need to understand the standard variations of passed, skipped and failed.

Deprecation strategy:

- [X] v5.x.x: Print out a warning when tests are not ran in strict mode https://github.com/cucumber/cucumber-jvm/pull/1835
- [x] v6.0.0: Remove snippet from summary printer (is included in exception - except for CLI) https://github.com/cucumber/cucumber-jvm/pull/1790
- [x] v6.0.0: Default to strict and ignore strict mode input `--strict` option, error on `--no-strict`.
- [ ] v7.0.0: Remove `--strict` and `--no-strict` options.

PR #1960, opened 2020-04-24
[Core] Default to --strict mode

mpkorstanje:
Unlike other implementations Cucumber-JVM defaults to `--non-strict` mode. By defaulting to `--strict` Cucumber follows the same standards as other implementations.

Additionally the option to set strict mode adds complexity for consumers of Cucumbers output. They have to interpret Cucumbers 6 scenario outcomes with this in mind. By removing `--strict/--non-strict` mode we remove this complexity all together.

With PR Cucumber will:
 * default to `--strict` 
 * throw an error on `--non-strict`.  
 * warn when `--strict` is used.
 * only print snippets as part of the exception thrown by JUnit rather then use the summary printer
 * only print snippets as part of the exception thrown by TestNG rather then use  the summary printer

Partial fix for #1788, https://github.com/cucumber/cucumber/issues/714

## Types of changes

## Checklist:


sebrose:
The title of this PR (and the comment in the Changelog) imply only a change of default. It looks like the whole idea of non-strict has been removed. 

Am I missing something?

mpkorstanje:
Internally the implementation is gone. However we're still in the process of phasing out the user facing part.

https://github.com/cucumber/cucumber-jvm/blob/d3f795686afbee0d633baa71b27539ebbc6e15ee/core/src/main/java/io/cucumber/core/options/CucumberPropertiesParser.java#L137-L142

Colin-Heyl:
Hi, wondering if someone can verify something for me?

We're using Cucumber at the moment and a test will **not** fail, if a feature is not found. 

Shouldn't that happen if strict mode is now the default? Or has strict functionality just been stripped out for some reason?

mpkorstanje:
Strict mode only affects translating PENDING, UNDEFINED, AMBIGUOUS steps results into a pass or fail. It does not do anything w.r.t to features being discovered.

If you have a feature request, please create a new issue though.

Issue #1971, opened 2020-05-11
Vulnerable version of jquery packaged in cucumber-core


aikebah:
**Describe the bug**
An OWASP depenendency-check plugin run against an internal test library surfaced a vulnerable version of jquery is packaged within the cucumber-core library.

**To Reproduce**
* Run [OWASP dependency-check](/jeremylong/DependencyCheck) against any version of cucumber-core.
* See it indicating a vulnerable version (3.4.1) of jquery in the analysis results.

**Expected behavior**
Cucumber-core has no vulnerable libraries included.

**Context & Motivation**

While I suspect that the given vulnerability (XSS vulnerability in html.preFilter) is not directly applicable to the library I would expect vulnerable javascript libraries to be fixed in a subsequent cucumber release.

**Your Environment**
 - Version used: 5.7.0; validated in github that the [current master](/cucumber/cucumber-jvm/blob/master/core/src/main/resources/io/cucumber/core/plugin/timeline/) still uses jquery-3.4.1.min.js

**Additional context**
See JQuery's release-notes for version 3.5.0 for details on the vulnerability as well as notes on potentially breaking code that depended on the old behaviour:
https://blog.jquery.com/2020/04/10/jquery-3-5-0-released/

Issue #1791, opened 2019-10-07
Cucumber 4.7.3 Uses jquery 1.8.2 with a security vunerability


foxy82:
**Describe the bug**
Thanks for fixing: https://github.com/cucumber/cucumber-jvm/issues/1786

However the HTML formatter still uses jquery 1.8.2 which again has a high impact security vulnerability. 

https://github.com/cucumber/cucumber-jvm/blob/8e7957fb7d2995bcc155ca5a2db648924653468c/core/src/main/java/cucumber/runtime/formatter/HTMLFormatter.java#L59

Can this one also be updated?

Thanks

aslakhellesoy:
Can you please send a pull request with a fix @foxy82?

Many thanks

clarkead:
PR Submitted: https://github.com/cucumber/cucumber-jvm/pull/1794

mpkorstanje:
Fixed by https://github.com/cucumber/cucumber-jvm/pull/1794

Issue #1786, opened 2019-10-04
Cucumber 4.7.2 uses version of jquery with a high impact security vulnerability.


foxy82:
**Describe the bug**
Cucumber 4.7.2 uses jquery versions that are susceptible to Prototype injection as reported by Nexus IQ server - report from the server below:

Explanation
The jquery package is vulnerable to Prototype Pollution. The jQuery.extend and jQuery.fn.extend functions defined in many files allow an untrusted object to extend Object.prototype. An attacker can modify and add prototype properties to JavaScript objects and can potentially leverage those changes to crash the application or execute remote code.

NOTE: This vulnerability has been assigned [CVE-2019-11358.](https://www.cvedetails.com/cve/CVE-2019-11358/)

Advisory Deviation Notice: The Sonatype security research team discovered that the CVSS score of the vulnerability is 9.8, not 6.1 as the advisory states.

Detection
The application is vulnerable by using this component.

Recommendation
We recommend upgrading to a version of this component that is not vulnerable to this specific issue.

Project: [https://blog.jquery.com/2019/04/10/jquery-3-4-0-released/](https://blog.jquery.com/2019/04/10/jquery-3-4-0-released/)

mpkorstanje:
Fixed with 8e7957fb7d2995bcc155ca5a2db648924653468c. v4.7.3 should be available in a few hours.

PR #1972, opened 2020-05-12
Upgrade JQuery to 3.5.1 through webjars, for easier Maven version bumps

timtebeek:
## Summary

Upgrade JQuery to 3.5.1 through webjars, for easier Maven version bumps

## Details

Replace packaged JQuery-3.4.1.min.js with a webjars dependency, which can be bumped using the Maven version bump script. That way it should be easier to upgrade to newer versions without code, html and test changes.

## Motivation and Context

Fixes https://github.com/cucumber/cucumber-jvm/issues/1971
After similar issues previously:
https://github.com/cucumber/cucumber-jvm/issues/1791
https://github.com/cucumber/cucumber-jvm/issues/1786
https://github.com/cucumber/cucumber-jvm/issues/1759

## How Has This Been Tested?

Unit tests.

## Types of changes

## Checklist:


timtebeek:
@aikebah could you have a look if this resolves your OWASP issues, while still keeping the TimelineFormatter functional?

timtebeek:
Now this does introduce two dependencies into cucumber-core(!) @mpkorstanje , which might be somewhat undesirable. I'll leave that up to decide whether this approach is acceptable.

core/pom.xml
```diff
@@ -63,6 +63,17 @@
...
         </dependency>
 
+        <dependency>
+            <groupId>org.webjars</groupId>
+            <artifactId>jquery</artifactId>
+            <version>3.5.1</version>
+        </dependency>
+        <dependency>
+            <groupId>org.webjars</groupId>
+            <artifactId>webjars-locator-core</artifactId>
```
mpkorstanje, line 73: This might conflict with other applications that use this as a production dependency.

You'll have to shade both.

aikebah:
> @aikebah could you have a look if this resolves your OWASP issues, while still keeping the TimelineFormatter functional?

@timtebeek This resolves the OWASP dependency-check issue (as it would no longer find the JQuery 3.4.1 script in the jar and the dependency for jquery-3.5.1 has a JQuery script that has -for now- no known vulnerabilities associated). 

But I can't say what it would do with respect to the timelineformatter. I just happened to have a maintenance task for an internal test-utility-library that defines some common Gherkin step definitions) and ran into the JQuery vulnerability being flagged in our build, which by default runs OWASP DependencyCheck as part of maven's verify phase.

I did an aggregate scan on all of Cucumber-JVM (source-branch of your PR) and see some other libs (that could use upgrades (as well as some false positives). I'll sift out the FPs (spotted some at first read, need further checks on others) and will file my findings as separate issue(s).

@mpkorstanje @timtebeek Would you prefer an issue per vulnerable lib, or an overall list of the vulnerable library versions that the Dependency-Check maven plugin was able to spot for the entire cucumber-jvm project tree?

As a side-note, do you know the [dependabot preview](https://github.com/marketplace/dependabot-preview)? It would assist you keeping libraries up-to-date by filing per-lib PRs when it spots updates - for example [this jackson databind upgrade PR in the DependencyCheck project](https://github.com/jeremylong/DependencyCheck/pull/2615).

core/src/main/java/io/cucumber/core/plugin/TimelineFormatter.java
```diff
@@ -38,7 +39,6 @@
             "/io/cucumber/core/plugin/timeline/index.html",
             "/io/cucumber/core/plugin/timeline/formatter.js",
             "/io/cucumber/core/plugin/timeline/report.css",
-            "/io/cucumber/core/plugin/timeline/jquery-3.4.1.min.js",
```
mpkorstanje, line 41: If I'm not mistaken, `chosen.jquery` is a jquery plugin. That probably have to be updated as well.

mpkorstanje:
@aikebah much appreciated but to avoid shuffling reports around I ran the check myself.

```
mvn org.owasp:dependency-check-maven:aggregate -P-examples
```

After d840161516e4ba1067539f1e5cf64d5f8c8561a9 there weren't any actionable items left - jQuery aside. Either there is no new non-major version of the non-transitive dependency and/or the dependency is in the provided scope.

But please feel free to out any items you do find actionable. I'll take them as a single list.

> As a side-note, do you know the dependabot preview? It would assist you keeping libraries up-to-date by filing per-lib PRs when it spots updates 

I'm aware of it but not convinced it benefits Cucumber too much.

Upgrading dependencies may have implications for semantic versioning and needs to be paired up with the right release. This does not combine well with a strategy where we immediately upgrade dependencies. Instead upgrading dependencies is included as part of the release process. Being mostly automated through a make command it is not too much of a burden.

core/src/main/java/io/cucumber/core/plugin/TimelineFormatter.java
aikebah: chosen.jquery is indeed a jquery plugin, but there's no update for it. It's at its latest release (which was a securityfix) released in June 2018. Nowadays the project [marks itself as deprecated](https://github.com/harvesthq/chosen#deprecated) awaiting decisions on their future direction.

mpkorstanje:
@timtebeek the `webjars-locator-core` dependency is a bit heavy. Just jackson alone will result in frequent owasp findings. And without the locator I think it will be easier to stick with just using the files for now. 

```
$ mvn dependency:tree
[INFO] +- org.webjars:jquery:jar:3.5.1:compile
[INFO] +- org.webjars:webjars-locator-core:jar:0.41:compile
[INFO] |  +- org.slf4j:slf4j-api:jar:1.7.7:compile
[INFO] |  +- io.github.classgraph:classgraph:jar:4.8.44:compile
[INFO] |  +- com.fasterxml.jackson.core:jackson-core:jar:2.9.8:compile
[INFO] |  \- org.webjars.npm:angular__http:jar:2.4.10:compile
```

timtebeek:
Agreed & understandable.. was just a nice experiment to try out this in a non-web setting. :)

PR #1979, opened 2020-05-25
[Core] Add location to tag expression expcetion

### LINKED ISSUE #1976: TagExpressionException doesn't include location
`TagExpressionException` are either thrown when parsing command line arguments, `cucumber.properties`, environment properties, or conditional hooks. Currently very little information is included. This could be improved by including the location of the tag expression that failed to be parsed and the full tag expression.

For example:

```
Tag expression '@foo not@bar' at 'com.example.MyStepDefinitions.beforeHook()' could not be parsed because <reason>
``` 

* This may require catching the `TagExpressionException` and rethrowing it.
* This may require moving the parsing of tag expressions from `TagPredicate` to the `CucumberPropertiesParser`, `CommandlineOptionsParser` and `CucumberOptionsAnnotationParser`.
* This may require making `CucumberPropertiesParser` aware of the properties it is parsing (either system, environment or file).

cyocum:
## Summary

This integrates changes to the way that TagExpressionParser works in Cucumber-JVM to fix #1976.

## Details

Updated the TagPredicate and the CoreHookDefinition to catch TagExpressionException and add information from the relevant classes to the exception message which is then wrapped in a RuntimeException and rethrown.  The rethrow as a RuntimeException is due to the fact that the constructor for TagExpressionParser is package scope only and is not available in other packages.

## Motivation and Context

This is to fix issue #1976.

## How Has This Been Tested?

Added tests to both TagPredicateTest and CoreHookDefinition test to see if the proper exception was rethrown and if the information expected was included in the expected exception.

## Types of changes

## Checklist:


core/src/main/java/io/cucumber/core/filter/TagPredicate.java
```diff
@@ -2,40 +2,28 @@
...
 final class TagPredicate implements Predicate<Pickle> {
 
     private final List<Expression> expressions;
 
-    TagPredicate(String tagExpression) {
-        this(tagExpression.isEmpty() ? emptyList() : singletonList(tagExpression));
+    TagPredicate(Expression tagExpression) {
+        this(Optional.ofNullable(tagExpression)
+                .map(Collections::singletonList)
+                .orElse(emptyList()));
```
mpkorstanje, line 20: Generally it is preferable to avoid nullable parameters.  The original `tagExpression` was already non-null because invoking  `tagExpression.isEmpty()` so you may be able to reduce this to

 ```
this(singletonList(requireNonNull(tagExpression))`;
```

File: core/src/main/java/io/cucumber/core/filter/TagPredicate.java [hunk 2]
```diff
@@ -2,40 +2,28 @@
...
     }
 
-    TagPredicate(List<String> tagExpressions) {
-        expressions = Optional.ofNullable(tagExpressions)
-                .map(this::tryParseTagExpression)
-                .orElse(new ArrayList<>());
-    }
+    TagPredicate(List<Expression> exprs) {
+        expressions = Optional.ofNullable(exprs)
+                .orElse(emptyList());
```
mpkorstanje, line 25: Same here: `exprs` is not nullable.

core/src/main/java/io/cucumber/core/filter/TagPredicate.java
mpkorstanje: Convention wise, using `Optional` to do a null check is not prefered. A ternary is usually more readable.

cyocum: Ok, no problem.  I tend to avoid ternary as I find them harder to read myself.

core/src/main/java/io/cucumber/core/options/CucumberOptionsAnnotationParser.java
```diff
@@ -66,10 +69,14 @@ private void addMonochrome(CucumberOptions options, RuntimeOptionsBuilder args)
...
 
-    private void addTags(CucumberOptions options, RuntimeOptionsBuilder args) {
+    private void addTags(Class<?> clazz, CucumberOptions options, RuntimeOptionsBuilder args) {
         String tagExpression = options.tags();
         if (!tagExpression.isEmpty()) {
-            args.addTagFilter(tagExpression);
+            try {
+                args.addTagFilter(TagExpressionParser.parse(tagExpression));
+            } catch (TagExpressionException tee) {
+                throw new IllegalArgumentException(String.format("%s at '%s'", tee.getMessage(), clazz.getName()), tee);
```
mpkorstanje, line 78: It's generally speaking best to capture exceptions close to the source. So I move the try-catch from the `parse` method here.

core/src/main/java/io/cucumber/core/options/CucumberPropertiesParser.java
```diff
@@ -77,7 +79,7 @@ public RuntimeOptionsBuilder parse(Map<String, String> properties) {
 
         parse(properties,
             FILTER_TAGS_PROPERTY_NAME,
-            Function.identity(),
+            TagExpressionParser::parse,
```
mpkorstanje, line 82: We'll have to look at this.

core/src/main/java/io/cucumber/core/runner/CoreHookDefinition.java
```diff
@@ -20,7 +21,13 @@
     private CoreHookDefinition(UUID id, HookDefinition delegate) {
         this.id = requireNonNull(id);
         this.delegate = delegate;
-        this.tagExpression = new TagExpressionParser().parse(delegate.getTagExpression());
+
+        try {
+            this.tagExpression = TagExpressionParser.parse(delegate.getTagExpression());
+        } catch (TagExpressionException tee) {
+            throw new IllegalArgumentException(String.format("%s at '%s'", tee.getMessage(), delegate.getLocation()),
```
mpkorstanje, line 28: You originally pushed this down into the different hook implementations. I think because I suggested something about being as close to the source as possible. However this means that we'll have to implement the correct location reporting in all different hook implementations. Using the `getLocation` method we can get the location of the hook. So this is close enough.

core/src/test/java/io/cucumber/core/filter/TagPredicateTest.java
```diff
@@ -3,19 +3,22 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 class TagPredicateTest {
 
     @Test
     void empty_tag_predicate_matches_pickle_with_any_tags() {
         Pickle pickle = createPickleWithTags("@FOO");
-        TagPredicate predicate = new TagPredicate("");
+        TagPredicate predicate = new TagPredicate(TagExpressionParser.parse(""));
```
mpkorstanje, line 21: Might be good to extract a little helper method here that takes a vararg of strings and returns a lsit of tag expressions. Avoids some repetition.

core/src/main/java/io/cucumber/core/options/CucumberPropertiesParser.java
mpkorstanje: Currently the problem is that it doesn't report back where the problem came from.

core/src/main/java/io/cucumber/core/options/CommandlineOptionsParser.java
```diff
@@ -89,7 +91,11 @@ private RuntimeOptionsBuilder parse(List<String> args) {
                 URI parse = GluePath.parse(gluePath);
                 parsedOptions.addGlue(parse);
             } else if (arg.equals("--tags") || arg.equals("-t")) {
-                parsedOptions.addTagFilter(removeArgFor(arg, args));
+                try {
+                    parsedOptions.addTagFilter(TagExpressionParser.parse(removeArgFor(arg, args)));
+                } catch (TagExpressionException tee) {
+                    throw new IllegalArgumentException(tee);
```
mpkorstanje, line 97: We'll have to look at this. Currently the problem is that it doesn't report back where the problem came from. But this is true for all other CLI arguments as well.

mpkorstanje:
I've reverted the changes to `cucumber-java` and `cucumber-java8`. I also cleaned up some non-idiomatic java stuff. You can use `Optional` to avoid null checks but only when a method already returns an optional type. Using `Optional.offNullable(thing).ifPresent(...)` doesn't help.

core/src/main/java/io/cucumber/core/filter/TagPredicate.java
```diff
@@ -2,31 +2,24 @@
...
 import static java.util.Collections.singletonList;
 
 final class TagPredicate implements Predicate<Pickle> {
 
-    private final List<Expression> expressions = new ArrayList<>();
+    private final List<Expression> expressions;
 
-    TagPredicate(String tagExpression) {
-        this(tagExpression.isEmpty() ? emptyList() : singletonList(tagExpression));
+    TagPredicate(Expression tagExpression) {
```
mpkorstanje, line 16: This constructor isn't used in production code. We might be able to get rid of it.

core/src/test/java/io/cucumber/core/filter/TagPredicateTest.java
cyocum: Cool.  I will have a look at this to see if I can do it.

core/src/main/java/io/cucumber/core/options/CommandlineOptionsParser.java
cyocum: It may be easier if we have a shared session to discuss the options.

core/src/main/java/io/cucumber/core/filter/TagPredicate.java
cyocum: Yeah, always a good idea to get rid of code that is not used anywhere.

cyocum:
Thanks for the changes. I have not had a chance to look over them yet.  I know you had suggested to revert the change that I made to HookDefinition.  I have done that locally but I am having serious difficulties with a test that I cannot see why it is failing so I am a bit stuck at the moment.

PR #2003, opened 2020-06-08
Renovate Config

nhojpatrick:

Renovate update, don't limit pr's being raised, also automatically label pr for quick filtering.

Rate Limit 0

Label dependencies

**Additional context**

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,5 +1,7 @@
 {
   "extends": [
     "config:base"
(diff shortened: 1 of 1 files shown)


mpkorstanje:
Not sure about the rate limit here. We are using free resources to build. They're protected by a leaky bucket. Otoh updates don't happen that often.

PR #2010, opened 2020-06-09
ensure user locale does not make DefaultSummaryPrinterTest failling

rmannibucau:
I'm in french locale by default so decimal separator is "," and not "." so DefaultSummaryPrinterTest fails because 0.00 != 0,00 whereas the test actually functionally worked.

this pr enforces the locale for the test to ensure the assert will pass.


mpkorstanje:
Ah thanks! This doesn't fix the root cause. I'll have a stab at it, there are some other complexities surrounding this.

rmannibucau:
@mpkorstanje root cause is there https://github.com/cucumber/cucumber-jvm/blob/afa3a99089f351cde9af53767f13e84c4e892eb7/core/src/main/java/io/cucumber/core/plugin/DefaultSummaryPrinter.java#L24 ;)

PR #2030, opened 2020-06-21
Rename the UndefinedStepException class to MissingStepDefinitionExcep…

### LINKED ISSUE #2026: Consider renaming UndefinedStepException -> MissingStepDefinitionException
When a step definition is missing cucumber fails with a undefined step exception. This seems kinda misleading because its the step definition that was missing, not the step that was undefined.

s4ik4t:
This pull request fixes #2026 i.e. when a step definition is missing cucumber fails with a undefined step exception. This seems kinda misleading because its the step definition that was missing, not the step that was undefined.


mpkorstanje:
Cheers! 

(Some how forgot to put that in!)

PR #2045, opened 2020-07-08
[JUnit] Make duplicate pickle names unique

mpkorstanje:
Individual examples do not have a unique name. And while a unique name is not
required by Junit, various integrations such as sbt-junit assume this the case.

By appending ` [#n]` to a scenario name it becomes unique again. So JUnit 4s
output becomes:

```
A feature with scenario outlines
  A scenario outline #1
  A scenario outline #2
  A scenario outline #3
  A scenario outline #4
```

The `#n` was taken from JUnit 5 which renders examples as:

```
A feature with scenario outlines
  A scenario outline
    With some other text
      Example #1
      Example #2
    With some text
     Example #1
     Example #2
```

Fixes: https://github.com/cucumber/cucumber-jvm-scala/issues/102

### DIFF
--- a/junit/src/main/java/io/cucumber/junit/Cucumber.java
+++ b/junit/src/main/java/io/cucumber/junit/Cucumber.java
@@ -5,8 +5,6 @@ import io.cucumber.core.feature.FeatureParser;
 import io.cucumber.core.filter.Filters;
 import io.cucumber.core.gherkin.Feature;
 import io.cucumber.core.gherkin.Pickle;
-import io.cucumber.core.logging.Logger;
-import io.cucumber.core.logging.LoggerFactory;
 import io.cucumber.core.options.Constants;
 import io.cucumber.core.options.CucumberOptionsAnnotationParser;
 import io.cucumber.core.options.CucumberProperties;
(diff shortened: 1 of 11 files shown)


gaeljw:
I tested it in the context of sbt junit, it seems to work but the error is not very self explanatory:

```[error] Test bdd.tests.XxxTest.initializationError failed: java.lang.NullPointerException: null, took 0.017 sec
[error]     at io.cucumber.junit.FileNameCompatibleNames.uniqueSuffix(FileNameCompatibleNames.java:29)
[error]     at io.cucumber.junit.FeatureRunner.lambda$new$0(FeatureRunner.java:52)
```

mpkorstanje:
It's not supposed to throw an error. :D

mpkorstanje:
@gaeljw found the problem. I moved the filter into the wrong stream.

rhande-mdsol:
Hello @mpkorstanje Thanks for looking into this. Are you planning to release this change ?

mpkorstanje:
After you tell me that it fixes the problem. :smile:

rhande-mdsol:
Sorry @mpkorstanje I am still trying to publish this project / PR and use it in my application, as I have not worked on Maven before.  Apologies

gaeljw:
Works perfectly for me in the two cases:
- "Regular" scenarios with same name
- Scenario Outlines

Great work @mpkorstanje

PR #2070, opened 2020-07-26
[Core] Add --publish option

aslakhellesoy:
This pull request introduced `--publish` - a simplified interface for publishing reports, similar to https://github.com/cucumber/cucumber-ruby/pull/1452

The report is printed to STDERR, because surefire suppresses STDOUT. The banner is only printed when the URL matches the message store URL. The base URL is hard-coded to `https://messages.cucumber.io/api/reports`, but can be overridden by defining the `CUCUMBER_PLUGIN_PUBLISH_URL` environment variable.

TODO:
- [x] Do not append `/api/reports` to the base url, and let users define the full URL with `CUCUMBER_PLUGIN_PUBLISH_URL`
- [x] Print a banner suggesting the use of `--publish` so users can more easily discover this feature.
- [x] Introduce a mechanism allowing users to suppress the banner (a config file)
- [x] Send additional HTTP authentication headers if an environment variable is defined (`CUCUMBER_PLUGIN_PUBLISH_TOKEN`)

Screenshots:

Screenshot 2020-07-26 at 10 26 57

Screenshot 2020-07-26 at 10 26 22


core/src/main/java/io/cucumber/core/plugin/UrlReporter.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public UrlReporter(Writer out) {
+        this.out = out;
+    }
+
+    public void report(URL url) {
+        String path = url.getPath();
+        int pathLength = path.length();
+        try {
+            out.append(CYAN).append("┌─────────────────────────────").append(times('─', pathLength)).append("┐")
```
mpkorstanje, line 22: Plugins that use ansi colors should implement the `ColorAware` interface so as to avoid printing ansi symbols when this is unwanted.

mpkorstanje:
For a bit of future proving. Consider using `cucumber.plugin.publish.*` as a prefix for all publish related properties. Currently it is not possible to inject properties into a plugin but we may make this possible in the future. The prefix for this would be `cucumber.plugin.<plugin-name>`.

.travis.yml
```diff
@@ -20,13 +20,19 @@ jobs:
     - stage: test
       jdk: openjdk11
       script: mvn verify -P-spotless-apply --toolchains .travis-toolchains.xml
-      env: VERIFY=true
+      env:
+      - VERIFY=true
+      # Remove this when the messages.cucumber.io domain is available
+      - CUCUMBER_MESSAGE_STORE_URL=https://b1f2yk58d8.execute-api.eu-west-3.amazonaws.com
```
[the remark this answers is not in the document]
aslakhellesoy: The presence of this environment variable is not what triggers the publishing. This just overrides the hard-coded `https://messages.cucumber.io` because that subdomain has not yet been set up.

Publishing is triggered by setting `publish=true` in the JUnit annotation, or by passing `--publish` on the CLI.

We have changed Cucumber so that it can publish results/reports to an external service. We're going to encourage users to do this. If we're encouraging users to do this I see no problem with doing it ourselves, eating our own dog food.

Can you elaborate on why you don't think we should publish results?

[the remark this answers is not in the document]
mpkorstanje: There are are too many downsides to publishing reports as part of the build. I'd roughly group them into technical and legal issues.

1. If I were to run the build locally I now must have an internet connection. I've developed much of Cucumber in public transport without internet or on a metered connection. So this is generally undesirable.

2. The test itself doesn't assert anything but the fact that uploading resulted in a 200 status code. We've already got this more or less covered by the unit tests for the `UrlOutputStream` so the only unique thing that tests add is testing the connection the reports API which isn't part of Cucumber.

3. A dependency on an external service is a possibility for failure. Since this service isn't strictly necessary to build and or publish Cucumber JVM it should be eliminated from the build. This removes a possible cause of flakes.

4. If you do design an integration tests it might be worth doing so properly. This means using multiple versions of Cucumber JVM and using the CCK rather then a fairly minimal example.

Then from a legal and risk perspective:

1. It is unlikely that random contributors will have agreed to the T&Cs of the reports API.

2. Software should be safe by default. We should prevent a scenario where someone by copying an example unwillingly and/or unwittingly uploads proprietary information to a third party service. The liabilities and contractual penalties involved can be quite significant.

core/src/main/java/io/cucumber/core/options/CucumberMessageStoreUrl.java
```diff
@@ -0,0 +1,28 @@
...
+            String apiUrl = new URL(new URL(url), "/api/reports").toExternalForm();
+            return String.format("message:%s", apiUrl);
+        } catch (MalformedURLException e) {
+            throw new CucumberException("Bad URL: " + url);
+        }
+    }
+
+    public static String getUrl() {
+        Map<String, String> props = CucumberProperties.create();
+        String url = props.get("CUCUMBER_MESSAGE_STORE_URL");
```
mpkorstanje, line 22: Please also register this property in the `Constants` class.

mpkorstanje: Comes to mind that this should probably also be written property style. The `CucumberProperties` class will translate this to upper case underscores.

.travis.yml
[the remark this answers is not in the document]
aslakhellesoy: Excellent points. @vincent-psarga and I had a chat about this today, and propose:

The `PublishFormatter` is enabled by either specifying `--publish` OR by defining `CUCUMBER_REPORTS_TOKEN`
We'll make it clear to users that `--publish` is only meant for "testing the waters" of Cucumber Reports, and should not be committed to git. Reports published this way will be open to anyone who knows the URL. Reports will be automatically deleted after a certain delay (1h, 6h, 24h, not sure) unless they are *claimed*.

When someone views a report published with `--publish` they will be able to *claim* that report by linking it to a GitHub repo. They'll authenticate with OAuth and then select the corresponding GitHub repo. After doing this they will see instructions to define an environment variable with a token in their CI environment: `CUCUMBER_REPORTS_TOKEN=xxxxxxx`. The presence of this environment variable will trigger the `PublishFormatter` and send the token as the `CUCUMBER_REPORTS_TOKEN` HTTP header. This will cause the report to be automatically claimed against the corresponding git repo.

Claimed reports can only be accessed by users who are authenticated via GitHub and have read permission to the corresponding repo.

Now onto your comments:

Technical:

1. This should be solved by the above. People hopefully won't commit their `--publish` / `publish = true` config, and if `CUCUMBER_REPORTS_TOKEN` is not defined, no publishing will happen

2. We'll later set `CUCUMBER_REPORTS_TOKEN` as an environment in CI (encrypted env var on Travis, or we can migrate to CircleCI or GitHub Actions where it can be set outside git.

3. The Cucumber Reports service is going to be a central part of the Cucumber ecosystem. It's our best shot at converting Cucumber Open users to paid products, and SmartBear's investment in Cucumber Open is contingent on this. We understand that some users may not want to publish reports, which is why we're trying to make it easy to opt out. It is however important that we gain confidence that reports publishing works as expected, which is why we want to verify this in the Cucumber-JVM build.

4. I disagree. Publishing reports from the CI environment seems unproblematic to me. Since the reports service is hosted by AWS we don't expect much downtime.

Legal:

1. The reports API is not collecting any [PII](https://en.wikipedia.org/wiki/Personal_data) from people contributing to Cucumber Open. The reports will only be published on the CI server using a `CUCUMBER_REPORTS_TOKEN` associated with the `cukebot@cucumber.io` GitHub user.

2. This is mitigated by deleting reports that are not claimed within a timeframe.

core/src/main/java/io/cucumber/core/options/CucumberMessageStoreUrl.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Map;
+
+public class CucumberMessageStoreUrl {
+    public static String getPluginString() {
+        String url = getUrl();
+        try {
+            String apiUrl = new URL(new URL(url), "/api/reports").toExternalForm();
```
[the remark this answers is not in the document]
aslakhellesoy: Done!

junit/src/main/java/io/cucumber/junit/JUnitCucumberOptionsProvider.java
```diff
@@ -72,7 +77,11 @@ public String tags() {
 
         @Override
         public String[] plugin() {
-            return annotation.plugin();
+            List<String> plugins = new ArrayList<>(asList(annotation.plugin()));
```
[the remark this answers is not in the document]
aslakhellesoy: Yep, will do those later when I'm happy with the basics.

mpkorstanje:
Looks like IDEA is mismatching the output to the wrong element in the hierarchy. This looks like an IDEA problem though.

image

mpkorstanje:
@aslakhellesoy because Junit 5 doesn't use an annotation to pass options I've added `cucumber.plugin.publish=true|false`.  For consistency sake we may as wel introduce it everywhere.

aslakhellesoy:
@mpkorstanje I've made some more changes to the banners, and also improved the junit platform integration.

The only thing I can think of that still needs doing is the banner suggestion to add `cucumber.plugin.publish.quiet=true` to `src/test/resources/junit-platform.properties`. That seems to have no effect. OTOH, adding it to `src/test/resources/cucumber.properties` works.

Do you know how to fix this? And even if you do, is there any reason we need to give users two choices here when `src/test/resources/cucumber.properties` works both with and without junit platform?

mpkorstanje:
> Do you know how to fix this? And even if you do, is there any reason we need to give users two choices here when src/test/resources/cucumber.properties works both with and without junit platform?

I don't think that is the case. Can you show me tomorrow?

mpkorstanje:
I've refactored the plugin logic in such a way that we don't add the `NoPublishFormatter` when `cucumber.plugin.publish.quiet=true` is  used.

core/src/main/java/io/cucumber/core/options/Constants.java
```diff
@@ -152,6 +152,29 @@
      */
     public static final String PLUGIN_PROPERTY_NAME = "cucumber.plugin";
 
+    /**
+     * Setting this to true will enable publishing.
+     */
+    public static final String PLUGIN_PUBLISH_ENABLED_PROPERTY_NAME = "cucumber.plugin.publish.enabled";
```
mpkorstanje, line 158: TODO: This should be consistent with the other constant documentation.

core/src/main/java/io/cucumber/core/options/CurlOption.java
```diff
@@ -13,7 +13,7 @@
     private final HttpMethod method;
     private final List<Entry<String, String>> headers;
 
-    CurlOption(URI uri, HttpMethod method, List<Entry<String, String>> headers) {
+    public CurlOption(URI uri, HttpMethod method, List<Entry<String, String>> headers) {
```
mpkorstanje, line 16: TODO: Nicer method here.

core/src/main/java/io/cucumber/core/plugin/PublishFormatter.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    public void setMonochrome(boolean monochrome) {
+        urlReporter.setMonochrome(monochrome);
+    }
+
+    private OutputStream makeUrlOutputStream(String token) throws IOException {
+        Map<String, String> properties = CucumberProperties.create();
+        // TODO: Move to properties parsing
+        String url = properties.getOrDefault(PLUGIN_PUBLISH_URL_PROPERTY_NAME, DEFAULT_CUCUMBER_MESSAGE_STORE_URL);
```
mpkorstanje, line 51: TODO: Move to properties parsing

core/src/main/resources/io/cucumber/core/options/USAGE.txt
```diff
@@ -153,3 +153,6 @@ cucumber.snippet-type=          # underscore or camelcase.
 Each property also has an `UPPER_CASE` and `snake_case` variant. For example
 `cucumber.ansi-colors.disabled` would also be understood as
 `CUCUMBER_ANSI_COLORS_DISABLED` and `cucumber_ansi_colors_disabled`.
+
+
+TODO: Document --publish and new properties here
```
mpkorstanje, line 158: TODO: Document --publish and new properties here

junit-platform-engine/src/main/java/io/cucumber/junit/platform/engine/Constants.java
```diff
@@ -92,6 +92,28 @@
      */
     public static final String PLUGIN_PROPERTY_NAME = io.cucumber.core.options.Constants.PLUGIN_PROPERTY_NAME;
 
+    /**
+     * Setting this to true will enable publishing.
```
mpkorstanje, line 96: This should use the same format as the other options

junit/src/main/java/io/cucumber/junit/CucumberOptions.java
```diff
@@ -93,6 +93,14 @@
      */
     String[] plugin() default {};
 
+    /**
+     * Publish report to https://reports.cucumber.io.
```
mpkorstanje, line 97: Should be consistent with the documentation for the options with the same name

testng/src/main/java/io/cucumber/testng/CucumberOptions.java
```diff
@@ -93,6 +93,14 @@
      */
     String[] plugin() default {};
 
+    /**
+     * Publish report to https://reports.cucumber.io.
```
mpkorstanje, line 97: Should be consistent with the documentation for the options with the same name

core/src/main/resources/io/cucumber/core/options/USAGE.txt
mpkorstanje: Annoyingly this documentation is also duplicated in `cucumber-junit-platform-engine`.

core/src/main/java/io/cucumber/core/plugin/PublishFormatter.java
mpkorstanje: This is kinda tricky because plugins are instantiated in a factory and we don't have a way to pass in the properties.

mpkorstanje:
Documentation aside I think we've got everything important now. I'd like to get rid of `CucumberProperties.create();` but I don't quite see how yet so I'm willing to leave it. It doesn't seem that important yet.

mpkorstanje:
@aslakhellesoy 

we have a several sources of documentation.

1. The change log
2. The release notes
3. The git log/merge request

And when making a new release I often distil the release notes from the git log and merge requests. However currently neither the git log, nor the merge request describe how this feature works in full. This leaves a gap in the documentation making it harder for future contributors to work out how we intended something to work. This will in turn either result in having to answer questions or people failing to contribute.

image

Cucumber JVM is configured to squash merge all commits. So when merging in Cucumber-JVM you'll get the chance to edit the commit message. I usually use this as an opportunity to update the PR to describe the why and copy this into the commit message.

See also [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/)

aslakhellesoy:
👍 I'll deal with this tomorrow before the release.

PR #2085, opened 2020-08-08
Update io.cucumber dependencies

aslakhellesoy:
This updates `io.cucumber` dependencies to use semver ranges - similar to https://github.com/cucumber/cucumber/pull/1140

This will allow end users to pull in minor and patch releases of `io.cucumber` libraries we depend on without the core team having to make a new release of `cucumber-jvm`.

As long as we're careful about respecting semver rules, this will make it easier to roll out bugfixes and minor additions in underlying libraries.


aslakhellesoy:
I seem to have broken a semver check: https://travis-ci.org/github/cucumber/cucumber-jvm/builds/716189841

Tracing this back, I wonder if this is because of this change:
https://github.com/cucumber/cucumber/commit/8668bdc5acf0194adf9d9dc94cf6dab2451315b0#diff-d695d6bdb4f931333a5c682cf999ab80

I moved the `JavaMethod` and `JavaStackTraceElement` messages as children of `SourceReference` - this is a convention we've used elsewhere in `messages.proto`. Messages that are only used inside one other message are defined inside that message.

What would happen if we ignored the semver recommendation and still released this as a minor bump? /cc @mpkorstanje

mpkorstanje:
I would recommend against using ranges in Cucumber-JVM

It will be harder to work out which combination of dependencies caused certain bugs and it may break the facade that Cucumber JVM provides around its dependencies.

We also aren't running integration tests against multiple versions. This means that while we may try not to break anything there are no guarantees.

I believe we discussed this before.

mpkorstanje:
In this instance you can supress the semver recommendation. Messages is an internal library.

pom.xml
```diff
@@ -22,15 +22,18 @@
...
+        <cucumber-expressions.version>[10.2.2,11.0.0)</cucumber-expressions.version>
+        <datatable.version>[3.4.0,4.0.0)</datatable.version>
+        <create-meta.version>[2.0.0,3.0.0)</create-meta.version>
+        <tag-expressions.version>[3.0.0,4.0.0)</tag-expressions.version>
+        <!--
+        Whenever messages is updated run `MSG_VERSION=... make update-compatibility`,
+        specifying the latest released version number for messages. 
+        -->
+        <messages.version>[13.0.1,14.0.0)</messages.version>
+        <gherkin.version>[15.0.0,16.0.0)</gherkin.version>
```
mpkorstanje, line 35: This won't work. We shade Gherkin in to avoid conflicts with tooling that uses older versions of Gherkin.

aslakhellesoy:
> So I see a reason to do this in the first place

Did you mean "I see *no* reason to do this in the first place?"

The way I see it, having floating dependencies allows us to roll out bugfixes without end users needing to do anything. People are generally slow upgrading their dependencies, and many teams don't even have a routine for it.

I can undo the ranges in this `pom.xml`, but what will happen with transitive dependencies that aren't shaded?

For example, if we release `cucumber-jvm` with a fixed dependency on `io.cucumber:html-formatter:9.0.0` (not shaded AFAICT), users will transitively get `io.cucumber:messages:13.0.1`. However, when we release a new patch or minor release of `io.cucumber:messages`, they will get the latest one.

So by using version ranges in monorepo libraries, there is still a chance that the ground will change underneath our users.

I remember we discussed this, and my understanding/recollection was that we're OK with this, since they are libraries we control, and we now have a [policy](https://github.com/cucumber/cucumber/blob/master/RELEASE_PROCESS.md#decide-what-the-next-version-should-be) for when version numbers are bumped.

We should go all-in with ranges, or not at all.

Didn't you agree that it would be OK to use ranges in monorepo libs? If we're shading them in the final product, what purpose would it serve?

mpkorstanje:
I do not believe this is an either-or situation.

When using version ranges, depending on when Maven last checked for updates different executions may result in different results. While the use of versions ranges is common in the Javascript eco system, Javascript also has utilities such as [npm-package-lock.json](https://docs.npmjs.com/files/package-lock.json) that ensure dependencies do not change between builds. These utilities are simply not present in the Java eco-system.

Version ranges do solve a specific problem in the mono repo and I do not object to using version ranges to solve it. However it also introduces a set of problems of its own. So we should not expose users of Cucumber-JVM to floating dependencies. Instead Cucumber-JVM should act like a facade around its dependencies. This means (amongst other things) pinning the floating dependencies to a specific version.

This does mean that a release in the mono-repo will result in a release of Cucumber-JVM. This does mean that our updates still cascade however unlike the mono-repo this is a very different trade-off. 
By pinning the versions we'll ensure builds are reproducible and follow the expected practices in the Java eco system. As part of the build we'll also run several integration tests to verify that the dependencies are truly compatible. This is important as while we currently have an intention not to break anything but when using floating versions we would not verify this prior to any release.
On the other side of the balance, because Cucumber-JVM acts like a façade around its dependencies it is rather rare that a dependency update does not include any changes in Cucumber-JVM. And at the same time the release process for Cucumber JVM is more streamlined then the monorepo.
So all in all the problems that floating dependencies would introduce are quite severe while the pain that it removes is rather minimal.

aslakhellesoy:
I've deleted version ranges.

@mpkorstanje we still have an unshaded dependency on `html-formatter` which has a transitive dependency on `messages` using a version range:

https://github.com/cucumber/cucumber/blob/bbd5c13b0dbf0842496ce51414c680e909b20b86/html-formatter/java/pom.xml#L29-L33

From what I can tell, this will cause end-users to get a different version of `messages` depending on when they build, even if `html-formatter` is pinned to a specific version.

Is this OK with you? From what you've said above I'm guessing no, so it makes me wonder why you would be ok with version ranges in the monorepo.

mpkorstanje:
That should be okay. `core/pom.xml` declares a dependency on both `messages` and `html-formatter`.  And because a fixed version is specified in the `pom.xml` there will be no floating versions. You can verify this by running `mvn dependency:tree -Dverbose=true` against `core/pom.xml`.

The shading shouldn't be relevant. We only shade `gherkin`, we don't shade `messages`.

aslakhellesoy:
Gotcha.

What's the use to have version ranges in the monorepo when it will have no effect in cucumber-jvm?

mpkorstanje:
So when you make a minor change in `messages` and `create-meta`, then a release of `gherkin` and `html-formatter` aren't needed. This avoids some work.

Normally this would be a problem because the version `messages` as a dependency of `create-meta` would conflict with the one from the `html-formatter` and our depencies wouldn't converge onto a single version. There would be a conflict between the two. And this conflict wouldn't be resolved by pinning.

With version ranges this conflict is resolved and we can also pin a version.

aslakhellesoy:
> So when you make a minor change in messages and create-meta, then a release of gherkin and html-formatter aren't needed. This avoids some work.

Aha! I finally understand. Thanks for explaining :-)

PR #2089, opened 2020-08-14
Remove 'plugin' from publish-related env vars/system props.

aslakhellesoy:
This changes the publish-related environment variables (and corresponding JVM system properties) from:

* `CUCUMBER_PLUGIN_PUBLISH_ENABLED`
* `CUCUMBER_PLUGIN_PUBLISH_TOKEN`
* `CUCUMBER_PLUGIN_PUBLISH_URL`
* `CUCUMBER_PLUGIN_PUBLISH_QUIET`

to:

* `CUCUMBER_PUBLISH_ENABLED`
* `CUCUMBER_PUBLISH_TOKEN`
* `CUCUMBER_PUBLISH_URL`
* `CUCUMBER_PUBLISH_QUIET`

The reasons for doing this are:

* The environment variables should be identical across Cucumber implementations
* Cucumber-JVM is the only implementation that has renamed "formatter" to "plugin", and "plugin" is confusing on Cucumber.js and Cucumber-Ruby
* The names are shorter

This is related to https://github.com/cucumber/cucumber-ruby/pull/1452#discussion_r470004804

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ jobs:
       script: mvn verify -P-spotless-apply --toolchains .travis-toolchains.xml
       env:
       - VERIFY=true
-      - CUCUMBER_PLUGIN_PUBLISH_TOKEN=trigger/build/with/publish/plugin
+      - CUCUMBER_PUBLISH_TOKEN=trigger/build/with/publish/plugin
 
       # 1.3 Coverage
     - stage: test
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
     Aslak Hellesøy, M.P. Korstanje).
    There are several ways to enable this:
    * `@CucumberOptions(publish = true)`
-   * `CUCUMBER_PLUGIN_PUBLISH_ENABLED=true` (Environment variable)
-   * `-Dcucumber.plugin.publish.enabled=true` (System property)
+   * `CUCUMBER_PUBLISH_ENABLED=true` (Environment variable)
+   * `-Dcucumber.publish.enabled=true` (System property)
 
 ### Changed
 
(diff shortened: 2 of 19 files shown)


mpkorstanje:
From: https://github.com/cucumber/cucumber-ruby/pull/1452#discussion_r470004804

> Cucumber-JVM is the only implementation that has renamed "formatter" to "plugin", and "plugin" is confusing on Cucumber.js and Cucumber-Ruby

I'd suggest keeping `PLUGIN` everywhere and planning to allocate some resources to migrate ruby and js to use plugin. Now it seems like we're piling technical debt of Cucumber Ruby onto Cucumber JVM. On the other hand, it doesn't matter that much. Being a plugin is an implementation detail of the `--publish` option.

aslakhellesoy:
> I'd suggest keeping PLUGIN everywhere and planning to allocate some resources...

Do you mean people? That could take months, and we can't hold up the release of this feature to wait for that to happen.

> ...to migrate ruby and js to use plugin.

Even if all Cucumber implementations had already adopted the *plugin* terminology I would have argued for shorter environment variable names to configure publishing. The variables are configuring *cucumber*'s *publish* feature, and naming related variables `CUCUMBER_PUBLISH_*` is more consistent with the exposed naming (`--publish` option, `publish` annotation property).

As you correctly point out, being a plugin is an implementation detail. It isn't even visible to end users.

I'd like to merge this on Monday (tomorrow) and make a release as soon as we've agreed whether or not to roll back https://github.com/cucumber/cucumber/pull/1145 and https://github.com/cucumber/cucumber-jvm/blob/eba65e956e53279b72b1907f21a2e8abe23bd5ad/pom.xml#L635-L646

mpkorstanje:
> I'd like to merge this on Monday (tomorrow) and make a release as soon as we've agreed whether or not to roll back cucumber/cucumber#1145 and 

:+1:

mpkorstanje:
> as soon as we've agreed whether or not to roll back cucumber/cucumber#1145 and  cucumber-jvm/pom.xml 

As mentioned in cucumber/cucumber#1145 I think we can keep the changes to cucumber-jvm/pom.xml, roll back cucumber/cucumber#1145, and then make a plan to properly deprecate/remove the case sensitives in the long.

PR #2096, opened 2020-08-20
Improve publish instructions

aslakhellesoy:
The old instructions for activating publishing allowed too many things to go wrong:

* Many users don't know how to set environment variables. They'll set them in Windows system settings not realising a reboot is required to make it take effect
* Adding `publish = true` to `@CucumberOptions` will have no effect if Cucumber is run from IDEA's native runner (green triangles in the gutter)

For this reason the top recommendation is now to set `cucumber.publish.enabled=true` in `src/test/resources/cucumber.properties`. This should be picked up regardless of how people run Cucumber.

I've also removed the recommendation to use `src/test/resources/junit-platform.properties` since it will be ignored when using the IDEA runner.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+* Publish instructions now recommend using `src/test/resources/cucumber.properties`.
+ ([#2096](https://github.com/cucumber/cucumber-jvm/pull/2096)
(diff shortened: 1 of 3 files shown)


mpkorstanje:
> I've also removed the recommendation to use src/test/resources/junit-platform.properties since it will be ignored when using the IDEA runner.

It depends how you launch the tests. If you launch tests in IDEA via JUnit 5 and you are using `cucumber-junit-platform-engine` then it will use the `junit-platform.properties`.

PR #2099, opened 2020-08-21
Issue a PUT request after a GET request responding with 202 and a Location header

aslakhellesoy:
The `https://messages.cucumber.io` service responds with [413 Payload Too Large](https://httpstatuses.com/413) for message files larger than [6 Mb](https://jun711.github.io/aws/handling-aws-api-gateway-and-lambda-413-error/). This was discovered during a [chat with one of our users](https://cucumberbdd.slack.com/archives/C60TKS3SL/p1597955993030800).

We've adopted the solution outlined in [Option 2 in this blog post](https://theburningmonk.com/2020/04/hit-the-6mb-lambda-payload-limit-heres-what-you-can-do/). A `GET https://messages.cucumber.io/api/reports` request will now respond with `202 Accepted` and a `Location` header pointing to S3.

This is not a proper redirect, so we're handling this in the code by issuing a new `PUT` request if the status is `202` and the `Location` header is present.

This change preserves backwards compatibility for all HTTP verbs except `GET`. However, this change in behaviour will only manifest itself for users issuing `GET` request to servers expecting a body in the request, which is [legal, but not useful](https://stackoverflow.com/questions/978061/http-get-with-request-body). We don't expect anyone to be affected.

For this reason we consider this change to be an addition, which should trigger a minor release.

The same change has been [applied to Ruby](https://github.com/cucumber/cucumber-ruby/pull/1464).

### DIFF
--- a/core/src/main/java/io/cucumber/core/plugin/PublishFormatter.java
+++ b/core/src/main/java/io/cucumber/core/plugin/PublishFormatter.java
@@ -7,19 +7,16 @@ import io.cucumber.plugin.ConcurrentEventListener;
 import io.cucumber.plugin.event.EventPublisher;
 
 import java.io.IOException;
-import java.net.URI;
-import java.util.AbstractMap.SimpleEntry;
 import java.util.Map;
 
 import static io.cucumber.core.options.Constants.PLUGIN_PUBLISH_URL_PROPERTY_NAME;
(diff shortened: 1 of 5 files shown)


mpkorstanje:
For future consideration, it might be worth removing the `UrlOutputstream` from all formatters except the publish formatter. It's becoming quite protocol specific.

PR #2103, opened 2020-08-26
Improve release instructions

cbliard:

While releasing Cucumber-JVM 6.6.0, we ran into some issues:
- We (wrongly) assumed that setting the version is done by editing it in the root `pom.xml`, which made the release process because versions in root pom and submodules pom do not match
- We manually reverted some files and got issues with the `update-changelog.sh` script
- We were not sure that credentials for pushing artifacts to sonatype would automatically work

Overall the overall release process went smoothly. Anyway we think we can improve it even further.

- [x] Make `update-changelog.sh` more robust and runnable on osx
- [x] Add information about proper way to set version
- [x] Clarify what credentials information is needed
- [x] Remove empty sections automatically

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -18,18 +18,49 @@ Also check if you can upgrade any dependencies:
 make update-dependency-versions
 ```
 
-## Make the release ##
+## Decide what the next version should be ##
 
-In order to make releases secrets are required. Members of the core team can
-install keybase and join the `cucumberbdd` team to access these secrets.
+This depends on what's changed (see `CHANGELOG.md`):
 
-Remove the empty sections in the changelog. Don't commit these but run: 
+* Bump `MAJOR` if:
+  * There are `Changed` or `Removed` entries
+  * A cucumber library dependency upgrade was major
+* Bump `MINOR` if:
+  * There are `Added` entries
+* Bump `PATCH` if:
+  * There are `Fixed` or `Deprecated` entries
+
+Display future version by running:
 
 ```
 make version
 ```
 
-Check if branch name and version are as expected.
+Check if branch name and version are as expected. To change version run:
+
+```
+mvn versions:set -DnewVersion=X.Y.Z-SNAPSHOT
+```
+
+## Secrets ##
+
+Secrets are required to make releases. Members of the core team can install
+keybase and join the `cucumberbdd` team to access these secrets.
+
+During the release process, secrets are fetched from keybase and used to sign
+and upload the maven artifacts.
+
+## Make the release ##
+
+Remove the empty sections in the changelog. Don't commit anything yet.
+
+Check if branch name and version are as expected:
+
+```
+make version
+```
+
+Do the release:
 
 ```
 make release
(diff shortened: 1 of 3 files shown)


cbliard:
I really feel like we could add some unit tests on `scripts/update-changelog.sh. Some kind of golden testing ones where we have input files and expected outputs for many different changelog files would be easy to maintain.

RELEASING.md
```diff
@@ -18,74 +18,58 @@ Also check if you can upgrade any dependencies:
...
 
-In order to make releases secrets are required. Members of the core team can
-install keybase and join the `cucumberbdd` team to access these secrets.
+* Bump `MAJOR` if:
+  * There are `Changed` or `Removed` entries
+  * A cucumber library dependency upgrade was major
+* Bump `MINOR` if:
+  * There are `Added` entries
+* Bump `PATCH` if:
+  * There are `Fixed` or `Deprecated` entries
```
mpkorstanje, line 31: Deprecations can only happen on a minor version.

> https://semver.org/
> 7. Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.

RELEASING.md
```diff
@@ -18,74 +18,58 @@ Also check if you can upgrade any dependencies:
...
 
-## Make the release ##
+## Decide what the next version should be ##
+
+This depends on what's changed (see `CHANGELOG.md`):
 
-In order to make releases secrets are required. Members of the core team can
-install keybase and join the `cucumberbdd` team to access these secrets.
+* Bump `MAJOR` if:
+  * There are `Changed` or `Removed` entries
```
mpkorstanje, line 26: This depends on the change. Not all changes have to be breaking changes.

mpkorstanje (commented):
Looking good!

Don't forget to copy this to:
- https://github.com/cucumber/cucumber-jvm-scala
- https://github.com/cucumber/cucumber-parent

> I really feel like we could add some unit tests on `scripts/update-changelog.sh. Some kind of golden testing ones where we have input files and expected outputs for many different changelog files would be easy to maintain.

This script is executed in the docker container so it could be added to the docker image. And as part of building the container some tests could be added. This would also remove the duplication I mentioned above.

The docker image is kept in:

https://github.com/cucumber/cucumber-build

RELEASING.md
cbliard: I copied the block from https://github.com/cucumber/cucumber/blob/master/RELEASE_PROCESS.md#decide-what-the-next-version-should-be

mpkorstanje: Then that's wrong too. 😄

cbliard: Indeed 🙂
Once we agree on this pull request, I can make the same changes in other repositories.
I changed it to reference semver. Sounds better?

RELEASING.md
cbliard: Yep, updated

RELEASING.md
```diff
@@ -18,74 +18,59 @@ Also check if you can upgrade any dependencies:
...
+  * There are `Added` entries, `Changed` entries preserving compatibility, or
+  `Deprecated` entries
+* Bump `PATCH` when you make backwards compatible bug fixes:
+  * There are `Fixed` entries
 
-Remove the empty sections in the changelog. Don't commit these but run: 
+Display future version by running:
 
 ```
 make version
```
mpkorstanje, line 37: We're running `make versions` twice now.

mpkorstanje (approved):

mpkorstanje:
I think we're good to go now.

PR #2110, opened 2020-09-04
Update creta-meta.version to v2.0.2

vincent-psarga:
This follows  [cucumber#1168](https://github.com/cucumber/cucumber/pull/1168)
There is a security issue involved which was fixed by `cucumber-ruby-meta` 2.0.2, this change ensures we use the fixed version.


mpkorstanje (commented):
Looks good. But for future changes, might be good to limit the change strictly to the version numbers and put the boyscouting in a separate commit and MR.

vincent-psarga:
Yes my editor removes leading spaces automatically, I did not really mean to commit that but it seemed to long to put them back :D

mpkorstanje:
The trick you can apply there is to roll back the additions you made. And only commit the stuff your editor did. Then add your additions on the next commit and PR.

PR #2111, opened 2020-09-04
[Spring] Notify TestContextManager about test method

mpkorstanje:
Fixes: https://stackoverflow.com/questions/63373893/


mpkorstanje:
To use `TextContextManager.afterTestExecution(instance, method, exception)` we need access to the scenario result. We can't access this because the object factory and backend are is isolated from the the scenarios execution. So we'll have to redesign the  redesign the backend module with call backs in mind.

PR #2121, opened 2020-09-13
[JUnit Platform] Support discovery selectors with FilePosition

mpkorstanje:
There was no good way to discover and execute a single test when using a file
based test system like Cucumber. Recently JUnit added support for file position
in the `FileSelector` and `ClasspathResourceSelector`.

See: https://github.com/junit-team/junit5/issues/2146

This implements a filter that prunes all scenarios that do not match the
`FilePosition` from the discovery selector. This feature itself is not new. It
was already implemented for the `UriSelector`.

This feature will benefit the implementation of [IDEA-227508](https://youtrack.jetbrains.com/issue/IDEA-227508). If you would like to use this feature in the future please up up-vote it.

TODO:
 - [ ] Create demo project for the benefit of the IDEA/Eclipse people.
 - [ ] Create issue to track feature request with IDEA/Eclipse/VsCode/Gradle/Maven ect

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased] (In Git)
 
 ### Added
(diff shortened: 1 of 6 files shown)


aslakhellesoy:
Wow amazing!

Issue #2122, opened 2020-09-15
Remove `CUCUMBER_PUBLISH_TOKEN` validation


aslakhellesoy:
The `PublishTokenParser` only allows tokens that look like a base64-encoded string. Token validation should only be the server's responsibility. We don't actually want it to be a base64-encoded string, just a UUID.

@mpkorstanje do you remember why we added this?

mpkorstanje:
Garbage in protection. Caching errors early is better in general. Rather then removing the validation you can expand the validation regex to include UUID like values too.

aslakhellesoy:
We don't want to commit to any particular format for authentication tokens. We want to have the flexibility to change this on the server without needing to make a new release of Cucumber-JVM.

The server needs to validate the token anyway, so the only benefit of client validation is early errors for users.

The price of that benefit is too high: Not being able to change the token validation on the server without also having to make a new release of Cucumber-JVM.

mpkorstanje:
Fair enough. Removing it won't break anything.

PR #2123, opened 2020-09-15
[Core] Remove publish token validation

aslakhellesoy:
This is a fix for #2122


mpkorstanje (commented):
LGTM. Don't forget to add the changelog entry.

aslakhellesoy:
The revapi checks failed again. I wonder if it's possible to have it ignore *anything* that isn't part of the public API as described in [SEMVER.md](https://github.com/cucumber/cucumber-jvm/blob/main/SEMVER.md)

mpkorstanje:
Unfortunately not.

PR #2134, opened 2020-09-27
[Build] Explicit docker pull before docker run

gaeljw:
Although `docker run` should also pull automatically, we have had several people mentioning the fact they had to pull by themselves, otherwise the release script was failing.

This PR adds an explicit `docker pull` command before the `docker run`.

This was tested succesfully when releasing Cucumber-Scala.


mpkorstanje:
Cheers!

PR #2139, opened 2020-10-03
[JUnit Platform] Map Cucumber events to JUnit report entries

mpkorstanje:


mpkorstanje:
This would significantly simply things:

https://github.com/junit-team/junit5/issues/2296

Issue #1805, opened 2019-10-15
Add Step information in @BeforeStep and @AfterStep hook


darsh9292:
Currently, we can only access `Scenario` level information in `@BeforeStep` & `@AfterStep`, it would be really great having step level information like, step name, line no, etc. 

This will help to add implementation like, common step level log.

mpkorstanje:
A common cause for this request is the need to create a custom report. Hooks however are designed to do things before and after a step and not suitable for generating reports. Instead you may want to examine the `Plugin` system. You can find a [recent blog post here](https://blog.jdriven.com/2019/01/cucumber-jvm-plugin/) and you can find examples of existing plugin [here](https://github.com/cucumber/cucumber-jvm/tree/master/core/src/main/java/io/cucumber/core/plugin) 

However should you still need this information in a before step hook please feel free to send a PR. For the end user this should look something like the example below. 

```java
@BeforeStep
public void before_step(Scenario scenario, Step step){

}
```

laeubi:
Maybe Scenario could contain a `getCurrentStep()` method that way it won't be needed to enhance the existing annotations. See also #1713 acess to the Glue Classes from a before/after step would be also very usefull.

laeubi:
> This will help to add implementation like, common step level log.

as @darsh9292 mentioned logging here, one might want to add information from the current step to the [MDC](http://www.slf4j.org/api/org/slf4j/MDC.html), beside that, it sounds obvious that a before/after hook can access the object it is a hook for (here a step) .

mpkorstanje:
>  one might want to add information from the current step to the MDC,

Again, you may want to examine the Plugin system. Step hooks hooks are only invoked around regular steps. Not around hooks so your context would be incomplete if you tried to use step hooks for this. 

> Maybe Scenario could contain a getCurrentStep() method that way it won't be needed to enhance the existing annotations.

Such a method would have to return null when accessed from a `@Before` or `@After` hook as there is no current step in that context. So while more convenient to implement, it would be harder to document and use.

laeubi:
Just wanted to give a hint why @darsh9292 might has use for this. Returning null if there is no actual current step seems appropriate to me. A plug-in might be too broad and does not has access to the current glue, but I can only guess whats the real use-case is.

bhreinb:
Hi there,

[...]

PR #2141, opened 2020-10-05
Add Step information in @BeforeStep and @AfterStep hook #1805

BenVercammen:
Related to [https://github.com/cucumber/cucumber-jvm/issues/1805]. 

Added "related step" to HookTestStep, referring to the (PickleStep)TestStep the BeforeStep/AfterStep annotated HookTestStep relates to. This way the actual PickleStepTestStep can be added to the signature of the BeforeStep/AfterStep annotated method.

### DIFF
--- a/core/src/main/java/io/cucumber/core/runner/HookTestStep.java
+++ b/core/src/main/java/io/cucumber/core/runner/HookTestStep.java
@@ -8,6 +8,7 @@ final class HookTestStep extends TestStep implements io.cucumber.plugin.event.Ho
 
     private final HookType hookType;
     private final HookDefinitionMatch definitionMatch;
+    private io.cucumber.plugin.event.TestStep relatedTestStep;
 
     HookTestStep(UUID id, HookType hookType, HookDefinitionMatch definitionMatch) {
         super(id, definitionMatch);
@@ -18,6 +19,16 @@ final class HookTestStep extends TestStep implements io.cucumber.plugin.event.Ho
     @Override
     public HookType getHookType() {
         return hookType;
+    }
+
+    @Override
+    public void setRelatedTestStep(io.cucumber.plugin.event.TestStep step) {
+        this.relatedTestStep = step;
+    }
+
+    @Override
+    public io.cucumber.plugin.event.TestStep getRelatedTestStep() {
+        return relatedTestStep;
     }
 
     HookDefinitionMatch getDefinitionMatch() {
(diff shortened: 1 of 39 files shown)


BenVercammen:
> Method was added to an interface. (breaks semantic versioning)
Should I raise from 6.8.1-SNAPSHOT to 6.9.0-SNAPSHOT?

mpkorstanje:
> Should I raise from 6.8.1-SNAPSHOT to 6.9.0-SNAPSHOT?

I'd leave it alone for now. Make the MR rather hard to review.

core/src/main/java/io/cucumber/core/runner/HookTestStep.java
```diff
@@ -8,6 +8,7 @@
 
     private final HookType hookType;
     private final HookDefinitionMatch definitionMatch;
+    private io.cucumber.plugin.event.TestStep relatedTestStep;
```
mpkorstanje, line 11: Instead of making test steps mutable consider passing the current step along with the `TestCaseState` and wrapping it in a `Step` object in the same way the `Scenario` object wraps the `TestCaseState`. This ensures the information passed to hooks is consistently.  It also ensures that an illegal state exception can be thrown when `Step` object is used outside of an active step hook.

java/src/main/java/io/cucumber/java/JavaHookDefinition.java
```diff
@@ -22,35 +27,74 @@
 
     private static Method requireValidMethod(Method method) {
         Class<?>[] parameterTypes = method.getParameterTypes();
-        if (parameterTypes.length > 1) {
-            throw createInvalidSignatureException(method);
+        Class<?>[] acceptedTypes = new Class<?>[] { io.cucumber.java.Scenario.class };
+
+        if (method.isAnnotationPresent(BeforeStep.class) || method.isAnnotationPresent(AfterStep.class)) {
+            acceptedTypes = new Class<?>[] { io.cucumber.java.Scenario.class, io.cucumber.core.gherkin.Step.class };
```
mpkorstanje, line 33: The arguments types can be in any order and neither is required.

mpkorstanje: Also the `io.cucumber.core.gherkin.Step` should not be exposed to the end user. The user should only have to use imports from `io.cucumber.java`.

core/src/main/java/io/cucumber/core/runner/HookTestStep.java
BenVercammen: Does this mean I need to create a new class **io.cucumber.java.Step** for this purpose?

mpkorstanje: Yes please.

BenVercammen:
Created new pull request (not using the main branch, my bad) at https://github.com/cucumber/cucumber-jvm/pull/2143

PR #2143, opened 2020-10-06
Add Step information in @BeforeStep and @AfterStep hook #1805

BenVercammen:
Related to https://github.com/cucumber/cucumber-jvm/issues/1805 

Added "related step" to HookTestStep, referring to the (PickleStep)TestStep the BeforeStep/AfterStep annotated HookTestStep relates to. This way the actual PickleStepTestStep can be added to the signature of the BeforeStep/AfterStep annotated method.

### DIFF
--- a/core/src/main/java/io/cucumber/core/backend/TestCaseState.java
+++ b/core/src/main/java/io/cucumber/core/backend/TestCaseState.java
@@ -1,9 +1,11 @@
 package io.cucumber.core.backend;
 
+import io.cucumber.plugin.event.TestStep;
 import org.apiguardian.api.API;
 
 import java.net.URI;
 import java.util.Collection;
+import java.util.Optional;
 
 @API(status = API.Status.STABLE)
 public interface TestCaseState {
@@ -91,4 +93,8 @@ public interface TestCaseState {
      */
     Integer getLine();
 
+    /**
+     * @return the current TestStep being executed in the Scenario.
+     */
+    Optional<TestStep> getCurrentTestStep();
 }
--- a/core/src/main/java/io/cucumber/core/backend/TestCaseState.java
+++ b/core/src/main/java/io/cucumber/core/backend/TestCaseState.java
@@ -1,6 +1,5 @@
 package io.cucumber.core.backend;
 
-import io.cucumber.plugin.event.TestStep;
 import org.apiguardian.api.API;
 
 import java.net.URI;
@@ -96,5 +95,5 @@ public interface TestCaseState {
     /**
      * @return the current TestStep being executed in the Scenario.
      */
-    Optional<TestStep> getCurrentTestStep();
+    Optional<Step> getCurrentTestStep();
 }
--- a/core/src/main/java/io/cucumber/core/runner/TestCaseState.java
+++ b/core/src/main/java/io/cucumber/core/runner/TestCaseState.java
@@ -1,6 +1,7 @@
 package io.cucumber.core.runner;
 
 import io.cucumber.core.backend.Status;
+import io.cucumber.core.backend.Step;
 import io.cucumber.core.eventbus.EventBus;
 import io.cucumber.messages.Messages;
 import io.cucumber.messages.Messages.Attachment;
@@ -145,8 +146,19 @@ class TestCaseState implements io.cucumber.core.backend.TestCaseState {
     }
 
     @Override
-    public Optional<TestStep> getCurrentTestStep() {
-        return testCase.getTestSteps().stream().filter(t -> t.getId().equals(currentTestStepId)).findFirst();
+    public Optional<Step> getCurrentTestStep() {
+        return testCase.getTestSteps().stream()
+                .filter(t -> t.getId().equals(currentTestStepId))
+                .findFirst()
+                .map(step -> {
+                    if (testStep instanceof HookTestStep) {
+                        return new HookStep((HookTestStep) testStep);
+                    }
+                    if (testStep instanceof PickleStepTestStep) {
+                        return new PickleStep((PickleStepTestStep) testStep);
+                    }
+                    return null;
+                });
     }
 
     Throwable getError() {
--- a/core/src/main/java/io/cucumber/core/runner/TestCaseState.java
+++ b/core/src/main/java/io/cucumber/core/runner/TestCaseState.java
@@ -8,6 +8,7 @@ import io.cucumber.messages.Messages.Attachment.ContentEncoding;
 import io.cucumber.plugin.event.EmbedEvent;
 import io.cucumber.plugin.event.Result;
 import io.cucumber.plugin.event.TestCase;
+import io.cucumber.plugin.event.TestStep;
 import io.cucumber.plugin.event.WriteEvent;
 
 import java.net.URI;
@@ -15,6 +16,7 @@ import java.util.ArrayList;
 import java.util.Base64;
 import java.util.Collection;
 import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
@@ -140,6 +142,11 @@ class TestCaseState implements io.cucumber.core.backend.TestCaseState {
     @Override
     public Integer getLine() {
         return testCase.getLocation().getLine();
+    }
+
+    @Override
+    public Optional<TestStep> getCurrentTestStep() {
+        return testCase.getTestSteps().stream().filter(t -> t.getId().equals(currentTestStepId)).findFirst();
     }
 
     Throwable getError() {
--- a/core/src/main/java/io/cucumber/core/runner/TestCaseState.java
+++ b/core/src/main/java/io/cucumber/core/runner/TestCaseState.java
@@ -9,7 +9,6 @@ import io.cucumber.messages.Messages.Attachment.ContentEncoding;
 import io.cucumber.plugin.event.EmbedEvent;
 import io.cucumber.plugin.event.Result;
 import io.cucumber.plugin.event.TestCase;
-import io.cucumber.plugin.event.TestStep;
 import io.cucumber.plugin.event.WriteEvent;
 
 import java.net.URI;
@@ -151,11 +150,11 @@ class TestCaseState implements io.cucumber.core.backend.TestCaseState {
                 .filter(t -> t.getId().equals(currentTestStepId))
                 .findFirst()
                 .map(step -> {
-                    if (testStep instanceof HookTestStep) {
-                        return new HookStep((HookTestStep) testStep);
+                    if (step instanceof HookTestStep) {
+                        return new HookStep((HookTestStep) step);
                     }
-                    if (testStep instanceof PickleStepTestStep) {
-                        return new PickleStep((PickleStepTestStep) testStep);
+                    if (step instanceof PickleStepTestStep) {
+                        return new PickleStep((PickleStepTestStep) step);
                     }
                     return null;
                 });
--- a/java/src/main/java/io/cucumber/java/JavaHookDefinition.java
+++ b/java/src/main/java/io/cucumber/java/JavaHookDefinition.java
@@ -1,20 +1,10 @@
 package io.cucumber.java;
 
 import io.cucumber.core.backend.HookDefinition;
-import io.cucumber.core.backend.HookStep;
 import io.cucumber.core.backend.Lookup;
-import io.cucumber.core.backend.PickleStep;
 import io.cucumber.core.backend.TestCaseState;
 
-import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
 
 import static io.cucumber.java.InvalidMethodSignatureException.builder;
 import static java.util.Objects.requireNonNull;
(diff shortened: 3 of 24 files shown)


BenVercammen:
Haven't updated the version yet, didn't help in the previous pull request either. Otherwise I've implemented the suggested changes.

mpkorstanje:
Cheers! I'm going to have to find some time to take a good look at this.

mpkorstanje:
I've tagged this with `hacktoberfest-accepted` just in case you were participating and I don't get around to reviewing this before the end of it.

hjh188:
@mpkorstanje @BenVercammen I'm using the cucumber for our test automation framework. I have a requirement on reusing the step parameters in step with data table input. here is a simple example for this feature requirement:
```
Given a book "<bookName>"
Check book from db:
| bookName       |  status   |
| ${bookName}  |    1          |

Examples:
|bookName|
|bookA|
|bookB|
```
Currently, I'm thinking on using this pr for a solution: save step info(parameters) afterStep and format the data table input with already saved step parameters in beforeStep.
My question is that what's plan for this pr, when it can be released?

mpkorstanje:
> when it can be released

When I can find enough time to give it the review it deserves.

core/src/main/java/io/cucumber/core/backend/TestCaseState.java
```diff
@@ -1,9 +1,11 @@
 package io.cucumber.core.backend;
 
+import io.cucumber.plugin.event.TestStep;
```
mpkorstanje, line 3: The `backend` api and the `plugin` and `java` api can evolve at different speeds. The `plugin` API is implemented by third parties that are slow to change while the `backend` almost entirely implemented by projects that are part of Cucumber. As such neither `backend` nor `java` should not depend or use any classes from `plugin`.

java/src/main/java/io/cucumber/java/Step.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+        if (!(currentTestStep.get() instanceof HookTestStep)) {
+            throw new IllegalStateException("Current TestStep is not a HookTestStep");
+        }
+        TestStep relatedTestStep = ((HookTestStep) currentTestStep.get()).getRelatedTestStep();
+        if (null == relatedTestStep) {
+            throw new IllegalStateException("No related TestStep for current HookTestStep was found");
+        }
+        if (!(relatedTestStep instanceof PickleStepTestStep)) {
+            throw new IllegalStateException("Related TestStep is not a PickleStepTestStep");
```
mpkorstanje, line 38: Given that most of these conditions this method checks for can only happen in case of a bug in Cucumber, consider using `Optional.filter`, and friends to make this more readable.

java/src/main/java/io/cucumber/java/Step.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Optional;
+
+/**
+ * BeforeStep or AfterStep Hooks that declare a parameter of this type will
+ * receive an instance of this class.
+ */
+@API(status = API.Status.EXPERIMENTAL)
+public class Step implements io.cucumber.plugin.event.Step {
+
+    final TestCaseState testCaseState;
```
mpkorstanje, line 20: This variable appears to go unused.

java/src/main/java/io/cucumber/java/Step.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apiguardian.api.API;
+
+import java.util.Optional;
+
+/**
+ * BeforeStep or AfterStep Hooks that declare a parameter of this type will
+ * receive an instance of this class.
+ */
+@API(status = API.Status.EXPERIMENTAL)
+public class Step implements io.cucumber.plugin.event.Step {
```
mpkorstanje, line 18: All classes that are part of the public API should be final.

java/src/main/java/io/cucumber/java/JavaHookDefinition.java
```diff
@@ -22,41 +29,88 @@
...
+        }
+    }
+
+    private static InvalidMethodSignatureException createInvalidSignatureException(Method method,
+                                                                                   Class<?>[] methodAnnotations,
+                                                                                   Set<Class<?>> acceptedTypes) {
+        InvalidMethodSignatureException.InvalidMethodSignatureExceptionBuilder builder = builder(method);
+        if (methodAnnotations.length == 0) {
+            throw new IllegalArgumentException(
+                    "Method should be annotated with one of: @Before, @After, @BeforeStep, @AfterStep");
```
mpkorstanje, line 87: This can't happen. If a method is not annotated there will be no instance of `JavaHookDefinition` created.

java/src/main/java/io/cucumber/java/JavaHookDefinition.java
```diff
@@ -22,41 +29,88 @@
...
+            }
+            if (BeforeStep.class.equals(annotationType) || AfterStep.class.equals(annotationType)) {
+                relevant = new Class<?>[] {BeforeStep.class, AfterStep.class};
             }
         }
+        return relevant;
+    }
 
-        return method;
+    private static Set<Class<?>> getAcceptedTypes(Class<?> annotationType) {
```
mpkorstanje, line 58: This method changes the behaviour of the `JavaHookDefinition` based on what it is annotated with. This indicates that `After` and `Before` annotated steps are different from `BeforeStep` and `AfterStep` annotated classes. When there is a difference in behaviour it makes sense to create a separate class for each. This should reduce the complexity of this class significantly.

java/src/main/java/io/cucumber/java/Step.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    public String getText() {
+        return delegate.getText();
+    }
+
+    @Override
+    public int getLine() {
+        return delegate.getLine();
```
mpkorstanje, line 60: The current line is part of the location. A dedicated method for line can be omitted.

java/src/main/java/io/cucumber/java/Step.java
```diff
@@ -0,0 +1,68 @@
+package io.cucumber.java;
+
+import io.cucumber.core.backend.TestCaseState;
+import io.cucumber.plugin.event.HookTestStep;
+import io.cucumber.plugin.event.Location;
+import io.cucumber.plugin.event.PickleStepTestStep;
+import io.cucumber.plugin.event.StepArgument;
+import io.cucumber.plugin.event.TestStep;
```
mpkorstanje, line 8: The `java` module and the `plugin` module evolve at different speeds. Users of the java module should only ever have to import classes from `io.cucumber.java`. At a glance `Location` and `StepArgument` are currently exposed.

mpkorstanje (commented):
I've finally had the time to review this. Functionally it works fine! However the implementation should be cleaned up. Solving this may result in some duplicate code, but this is usually okay when aimed at different consumers (e.g. backend implementations vs. plugins vs. users of Cucumber).

java/src/main/java/io/cucumber/java/Step.java
mpkorstanje: Note: Currently this is not enforced but will/may be in the future once we support `module-info.java` in all modules.

krisztianzagonyi-mintestio:
@BenVercammen Can you review it? I am so glad that you worked with it, and cant wait for using it it my code. :)
Thank you for help guys :)

mpkorstanje:
@krisztianzagonyi-mintestio if you can't wait, perhaps you could improve on this merge request @BenVercammen and submit those improvements as PR to @BenVercammen.

BenVercammen:
@krisztianzagonyi-mintestio I'll try to clean up the PR as per the comments of @mpkorstanje , but I'll need to wrap my head around the existing architecture so it might take a while...

BenVercammen:
@mpkorstanje I've pushed some new commits that should resolve most of your previous concerns. Can you already have a look at those? 

In the meantime, I'll see if I can split up the JavaHookDefinition to account for the differences between Before/After and BeforeStep/AfterStep.

core/src/main/java/io/cucumber/core/runner/TestCaseState.java
```diff
@@ -142,6 +145,23 @@ public Integer getLine() {
...
+            return Optional.empty();
+        }
+        TestStep testStep = first.get();
+        if (testStep instanceof HookTestStep) {
+            return Optional.of(new HookStep((HookTestStep) testStep));
+        }
+        if (testStep instanceof PickleStepTestStep) {
+            return Optional.of(new PickleStep((PickleStepTestStep) testStep));
+        }
+        return Optional.empty();
```
vtheuer, line 162:
```suggestion
        return testCase.getTestSteps().stream()
                .filter(t -> t.getId().equals(currentTestStepId))
                .findFirst()
                .map(step -> {
                    if (testStep instanceof HookTestStep) {
                        return new HookStep((HookTestStep) testStep);
                    }
                    if (testStep instanceof PickleStepTestStep) {
                        return new PickleStep((PickleStepTestStep) testStep);
                    }
                    return null;
                });
```

java/src/main/java/io/cucumber/java/JavaHookDefinition.java
```diff
@@ -22,41 +32,108 @@
...
+        if (currentTestStep.get() instanceof HookStep) {
+            PickleStep currentStep = ((HookStep)currentTestStep.get()).getRelatedStep();
+            if (null == currentStep) {
+                throw new IllegalStateException("Current HookStep has no related PickleStep");
+            }
+            return currentStep;
+        }
+        throw new IllegalArgumentException(
+                String.format("Current TestStep should be a %s instead of a %s",
+                        HookStep.class.getSimpleName(), currentTestStep.get().getClass().getSimpleName()));
```
vtheuer, line 136:
```suggestion
        io.cucumber.core.backend.Step step = state.getCurrentTestStep()
                .orElseThrow(() -> new IllegalStateException("No current TestStep was found in TestCaseState"));

        HookStep hookStep = Optional.of(step)
                .filter(HookStep.class::isInstance)
                .orElseThrow(() -> new IllegalArgumentException(String.format("Current TestStep should be a %s instead of a %s",
                        HookStep.class.getSimpleName(), currentTestStep.get().getClass().getSimpleName())));
                                
        return Optional.of(hookStep)
                .map(HookStep::getRelatedStep)
                .orElseThrow(() -> new IllegalStateException("Current HookStep has no related PickleStep"));

```

vtheuer (commented):
Here are two more idiomatic ways of using Optionals *(I'm not very familiar with the Github etiquette, so if these unsolicited suggestions seem out of place, feel free to ignore)*.

BenVercammen:
@mpkorstanje I think I've taken all your feedback into account, so if you have time to have another look at the code...

mpkorstanje:
I'll try and have a look soon!

mpkorstanje:
> The java module and the plugin module evolve at different speeds. Users of the java module should only ever have to import classes from io.cucumber.java. At a glance Location and StepArgument are currently exposed.

At a glance this has not been addressed. Currently users of the java module will still have to use imports from plugin or core. Simply providing step arguments as an `Object` hides this.

Additionally the additon of `PickleStep.Keyword` seems rather odd. Could you provide an explanation of how you intend to use the methods of `Step`?

BenVercammen:
> Additionally the additon of `PickleStep.Keyword` seems rather odd. Could you provide an explanation of how you intend to use the methods of `Step`?

The use case we have in mind is the following:

```java
    @AfterStep
    public void doSomethingAfterEveryGiven(Step step) {
        if (Step.Keyword.GIVEN.equals(step.getKeyword())) {
            ...
        }
    }
```

I added the `PickleStep.Keyword` enum (in **cucumber-core**) in order not to expose the `GwtStepType` (from **cucumber-gherkin**). I'm not sure if that's really necessary, but there aren't any `io.cucumber.core.gherkin.*` imports in **cucumber-java** (except for a few Test classes).

BenVercammen:
> > The java module and the plugin module evolve at different speeds. Users of the java module should only ever have to import classes from io.cucumber.java. At a glance Location and StepArgument are currently exposed.
> 
> At a glance this has not been addressed. Currently users of the java module will still have to use imports from plugin or core. Simply providing step arguments as an `Object` hides this.

I don't understand. The java module is heavily depending on core but I don't see any links with the plugin module? I guess I'm missing something, so could you please elaborate a bit?

mpkorstanje:
> I added the PickleStep.Keyword enum (in cucumber-core) in order not to expose the GwtStepType (from cucumber-gherkin). 

I was looking for a use case. Though I think I may have missed your explanation of it in https://github.com/cucumber/cucumber-jvm/issues/1805#issuecomment-703457441

> I'm looking for something similar, however I don't need it for logging, but to do something after each @given step. I'll try to come up with a pull request...

Generally speaking we try to discourage scoping steps to keywords. It tends to lead to ambiguous language. This seems to veer into the same direction.

> I don't understand. The java module is heavily depending on core but I don't see any links with the plugin module? I guess I'm missing something, so could you please elaborate a bit?

Cucumber is organized such that a user implementing step definitions with Cucumber Java never has to interact with the core directly. This makes change and estimating the impact of it easier. So the api for a user writing step definitions is `cucumber-java`, not `cucumber-core` nor `cucumber-plugin`. 

Currently by having `io.cucumber.java.step` implement `io.cucumber.plugin.event.Step` the `io.cucumber.plugin.event.StepArgument` and `io.cucumber.plugin.event.Location` as well as `io.cucumber.plugin.event.Step` itself are exposed to the user.

For now I would suggest exposing only the following:

```
public final class Step {

    String getKeyword();

    String getText();

    URI getUri();

    Location getLocation();

}
```

BenVercammen:
> Generally speaking we try to discourage scoping steps to keywords. It tends to lead to ambiguous language. This seems to veer into the same direction. Without further explanation I don't think I'd consider this acceptable.

Our specific use case is as follows: 
 - we have a revisioning system (envers) in place that increments the revision number of an entity on each change
 - depending on the changes in the revision, we do some extra post processing (eg: sending out events)
 - we have scenario's in which we want to test whether or not the extra processing occurs
 - unfortunately, changes made during execution of our "given" statements also trigger the revisioning system
     - grouping all relevant properties into a single "given" statement would render the scenario unreadable
     - therefore we have multiple "given" steps to set up the entity under test
 - in order for us to be able to check if the post processing was triggered during a"given" or the "when" statement, we need to know the last revision number before the actual "when" statement
 - therefore, we'd like to keep track of the last revision number after each "given" statement
 - at the moment, we're making use of a "dummy" statement that retrieves and stores the current revision number in our state before the relevant "when" statement is executed
    - this however makes the scenario less readable 
    - and it exposes the fact that we're working with revisions, which business should not be concerned with

So we basically want to do some "implicit state housekeeping" in between steps, and therefore need to distinguish between the actual keywords used for that step.

Another thing to note is that we use Cucumber on an "integration test" level in which we start our full Spring application, while stubbing/mocking external interfaces. We are not able to stub/disable the revisioning system in our tests.

BenVercammen:
> Cucumber is organized such that a user implementing step definitions with Cucumber Java never has to interact with the core directly. This makes change and estimating the impact of it easier. So the api for a user writing step definitions is `cucumber-java`, not `cucumber-core` nor `cucumber-plugin`.

Am I correct to say that importing and/or exposing the `io.cucumber.plugin.event` classes in `JavaHookDefinition` is to be avoided? So the main issue is in my change in the JavaHookDefinition class?

> Currently by having `io.cucumber.java.step` implement `io.cucumber.plugin.event.Step` the `io.cucumber.plugin.event.StepArgument` and `io.cucumber.plugin.event.Location` as well as `io.cucumber.plugin.event.Step` itself are exposed to the user.

This part is confusing, as I don't have anything implementing that `io.cucumber.plugin.event.Step` class. Are you sure you've looked at my latest commit? I did squash all commits, maybe a bit prematurely?

danic-wd:
Hi,

I have a use case along the lines of: https://github.com/cucumber/cucumber-jvm/issues/1805#issuecomment-778688360
A fair amount of work seems to have been going into this PR. Is the plan to get it merged soon ?
Thanks

Issue #2104, opened 2020-08-26
Test failures due to CRLF and Symlink in cucumber-jvm installation tests when running in windows


jdpennell:
**Describe the bug**
When building in Windows using jdk 11.0.2, several tests in BannerTest, CommandLineOptionsParserTest, NoPublishFormatterTest, ResourceScannerTest, and UrlReporterTest fail due to CRLF (carriage return line feed) issues and symlink issues. Maven was launched from git bash.

[BannerTest.results.txt](https://github.com/cucumber/cucumber-jvm/files/5131377/BannerTest.results.txt)
[CommandLineOptionsParserTest.results.txt](https://github.com/cucumber/cucumber-jvm/files/5131378/CommandLineOptionsParserTest.results.txt)
[NoPublishFormatterTest.results.txt](https://github.com/cucumber/cucumber-jvm/files/5131379/NoPublishFormatterTest.results.txt)
[ResourceScannerTest.results.txt](https://github.com/cucumber/cucumber-jvm/files/5131380/ResourceScannerTest.results.txt)
[UrlReporterTest.results.txt](https://github.com/cucumber/cucumber-jvm/files/5131381/UrlReporterTest.results.txt)

**To Reproduce**
1. Add the JDK 11 toolchain to Maven
2. Clone cucumber-jvm to a local disk with no UNC paths
3. In a Git Bash terminal, cd to the cucumber-jvm repository
4. Run either GnuWin32 Make or mvn clean install

**Expected behavior**
The Core project builds without errors, or with errors unrelated to CRLF/LF differences between operating environments
The Core project builds without errors, or with errors unrelated to symlinks not being reported, unless symlink availability is the behavior under test. If performance with symlinks is under test, this test should be skipped if the OS does not support symlinks.

**Context & Motivation**

As an automated test engineer, I would like to be able to build and deploy cucumber to a managed workstation with an unpredictable environment, and I would like test result failures to not exhibit false positives so that important issues can be revealed by included tests.

**Your Environment**
 - commit 26d1ad7fa0a68136461e8d339ce82ff9a7b44f5f
 - Operating system : Windows 10
 - Build tool : Apache Maven 3.6.0

mpkorstanje:
> As an automated test engineer, I would like to be able to build and deploy cucumber to a managed workstation with an unpredictable environment, and I would like test result failures to not exhibit false positives so that important issues can be revealed by included tests.

[...]

PR #2145, opened 2020-10-06
[Core] Fix platform dependent unit tests

kgeilmann:

Fixing the last test of #2104 and the PluginFactoryTest, which also fails on windows.

Ignore ResourceScannerTest#scanForResourcesPathSymlink on windows, because to make it run windows needs to be in developer mode, and the repo has to be cloned with actived symlinks.
The PluginFactory Test fails on windows, because the plugins (the real ones, not the test plugins) which get a file paramter, open the file directly during instantiation. On windows the test failed, because junit cannot delete the tmpDir due to open files. The change uses the EventPublisher to send events to the plugin, especially the TestRunFinished Event. During handling of TestRunFinished the plugins close the file and junit can successfully clean up afterwards.

**Additional context**
Add any other context or screenshots about the pull request here.

### DIFF
--- a/core/src/test/java/io/cucumber/core/plugin/PluginFactoryTest.java
+++ b/core/src/test/java/io/cucumber/core/plugin/PluginFactoryTest.java
@@ -6,7 +6,15 @@ import io.cucumber.core.runner.ClockStub;
 import io.cucumber.core.runtime.TimeServiceEventBus;
 import io.cucumber.plugin.ConcurrentEventListener;
 import io.cucumber.plugin.EventListener;
-import io.cucumber.plugin.event.*;
+import io.cucumber.plugin.event.EventHandler;
+import io.cucumber.plugin.event.EventPublisher;
+import io.cucumber.plugin.event.PickleStepTestStep;
+import io.cucumber.plugin.event.Result;
+import io.cucumber.plugin.event.Status;
+import io.cucumber.plugin.event.TestCase;
+import io.cucumber.plugin.event.TestRunFinished;
+import io.cucumber.plugin.event.TestRunStarted;
+import io.cucumber.plugin.event.TestStepFinished;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.function.Executable;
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/cucumber/core/plugin/PluginFactoryTest.java
```diff
@@ -49,12 +49,16 @@
     void instantiates_junit_plugin_with_file_arg() {
         Object plugin = fc.create(parse("junit:" + tmp.resolve("cucumber.xml")));
         assertThat(plugin.getClass(), is(equalTo(JUnitFormatter.class)));
+
+        release_resources(plugin);
```
aslakhellesoy, line 53: Please use camelCase - we only use snake_case for test methods (because it reads better in test output).

core/src/test/java/io/cucumber/core/plugin/PluginFactoryTest.java
```diff
@@ -4,11 +4,10 @@
...
 import io.cucumber.core.runner.ClockStub;
 import io.cucumber.core.runtime.TimeServiceEventBus;
-import io.cucumber.plugin.event.PickleStepTestStep;
-import io.cucumber.plugin.event.Result;
-import io.cucumber.plugin.event.Status;
-import io.cucumber.plugin.event.TestCase;
-import io.cucumber.plugin.event.TestStepFinished;
+import io.cucumber.plugin.ConcurrentEventListener;
+import io.cucumber.plugin.EventListener;
+import io.cucumber.plugin.event.*;
```
aslakhellesoy, line 9: No wildcard imports please

aslakhellesoy (changes requested):
LGTM!

core/src/test/java/io/cucumber/core/plugin/PluginFactoryTest.java
```diff
@@ -49,12 +49,16 @@
...
+
+        release_resources(plugin);
     }
 
     @Test
     void instantiates_rerun_plugin_with_file_arg() {
         Object plugin = fc.create(parse("rerun:" + tmp.resolve("rerun.txt")));
         assertThat(plugin.getClass(), is(equalTo(RerunFormatter.class)));
+
+        release_resources(plugin);
```
mpkorstanje, line 61: This is quite repetitive. Consider wrapping `fc.create` and releasing the resources before returning the created plugin.

kgeilmann: fc is the instance under test and personally I prefer to see the tested call in the test directly. But I moved it into an afterEach hook, looks less cluttered

PR #2148, opened 2020-10-09
[Compatibility] Fix platform dependent tests

kgeilmann:

This should fix all remaining problems with building under windows #2104

- Disabled scanForResourcesDirectorySymlink test on windows, because it suffers from the same problem as the already disabled scanForResourcesPathSymlink test

To make build work out of the box on windows the following two changes were done:
- Set maven configuration, such that maven is always called with the file encoding set to UTF-8. This is needed, because in the generated I18N sourcefiles use UTF-8 based filenames, which are otherwise not correctly passed to the compiler. 
- Ensure, that git does not do end-of-line conversion on the resource files in the compatibility module. If it does, the feature file gets converted and comparing it bytewise to the provided expected values fails.

### DIFF
--- a/.mvn/jvm.config
+++ b/.mvn/jvm.config
@@ -1,0 +1,1 @@
+-Dfile.encoding=UTF-8
--- a/compatibility/src/test/resources/.gitattributes
+++ b/compatibility/src/test/resources/.gitattributes
@@ -1,0 +1,3 @@
+# Ensure that LF line endings are kept on checkout (even on windows). Tests will fail otherwise
+*.ndjson    text    eol=lf
+*.feature   text    eol=lf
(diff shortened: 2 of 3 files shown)


.mvn/jvm.config
```diff
@@ -0,0 +1 @@
+-Dfile.encoding=UTF-8
```
mpkorstanje, line 1: This is pretty neat!

compatibility/src/test/resources/.gitattributes
```diff
@@ -0,0 +1,3 @@
+# Ensure that LF line endings are kept on checkout (even on windows). Tests will fail otherwise
+*.ndjson    text    eol=lf
+*.feature   text    eol=lf
```
mpkorstanje, line 3: Very nice!

mpkorstanje:
Thanks again. Much appreciated!

Issue #2147, opened 2020-10-08
Replace Travis CI with Github Actions


mpkorstanje:
Currently we are using [Travis CI to build Cucumber JVM](https://github.com/cucumber/cucumber-jvm/blob/main/.travis.yml). 

Currently Travis only supports Java on Unix. While Unix and OSx are quite similar, it does lead to a large number of issues involving Windows: https://github.com/cucumber/cucumber-jvm/pull/2145, https://github.com/cucumber/cucumber-jvm/issues/2108 , https://github.com/cucumber/cucumber-jvm/issues/1849, https://github.com/cucumber/cucumber-jvm/issues/1551, https://github.com/cucumber/cucumber-jvm/issues/1547, https://github.com/cucumber/cucumber-jvm/pull/1541, https://github.com/cucumber/cucumber-jvm/issues/1540, https://github.com/cucumber/cucumber-jvm/pull/981, https://github.com/cucumber/cucumber-jvm/issues/963, https://github.com/cucumber/cucumber-jvm/issues/906, https://github.com/cucumber/cucumber-jvm/pull/801, https://github.com/cucumber/cucumber-jvm/issues/787, https://github.com/cucumber/cucumber-jvm/pull/555, https://github.com/cucumber/cucumber-jvm/issues/559, https://github.com/cucumber/cucumber-jvm/issues/126, https://github.com/cucumber/cucumber-jvm/pull/100, https://github.com/cucumber/cucumber-jvm/issues/99, https://github.com/cucumber/cucumber-jvm/pull/34, https://github.com/cucumber/cucumber-jvm/issues/17.

Switching to Github Actions should allow us to [support Windows](https://github.com/junit-team/junit5/blob/main/.github/workflows/main.yml#L48) as well as [OSx](https://github.com/junit-team/junit5/blob/main/.github/workflows/main.yml#L84). Additionally on Unix we can use the [cucumber-build](https://github.com/cucumber/cucumber-build) image.

cunhazera:
Do you guys plan to keep travis toolchains with Github Actions?

mpkorstanje:
We won't need `.travis-toolchains.xml` but we will need a tool chain configuration.

cunhazera:
@mpkorstanje I can not build the project locally. Do you have a guide listing what is required to build the project locally?

cunhazera:
What command should I run to build the project locally?

mpkorstanje:
```
mvn clean install
```

[...]

PR #2162, opened 2020-10-26
Migrate from coverall to codecov

RDBreed:

Coverall combined with the maven plugin is not (yet?) supported with Github Actions. 

https://github.com/cucumber/cucumber-jvm/issues/2147#issuecomment-713903688

Migrating from coverall to codecov would help because it supports Github Actions.

Note this PR has a base/dependency on PR 2154...


RDBreed:
Oh before I forget... We need to have an account for codecov.

- [ ] Configure account for cucumber account on codecov

mpkorstanje:
I can facilitate that.

mpkorstanje:
All done!

Build is currently failing because Maven central hasn't synced our upstream dependencies yet, that should resolve itself soonish. All in all much appreciated!

PR #2180, opened 2020-12-09
Revert "Configure WhiteSource for GitHub.com"

mpkorstanje:
Reverts cucumber/cucumber-jvm#2177


aslakhellesoy:
Hi @mpkorstanje I think we'll actually want this. We're actually being obliged to use it. /cc @vincent-psarga

mpkorstanje:
@mloewinger74 as request in https://github.com/cucumber/cucumber-js/pull/1505  it would be appreciated if you could discuss any sweeping changes in the #commiters slack first.

Issue #2190, opened 2020-12-14
Cucumber command line options not overriding


a1dutch:
**Describe the bug**
The `cucumber.plugin` command line options are not overriding the ones set on the CucumberOptions.

with the following runner

```
@RunWith(Cucumber.class)
@CucumberOptions(
    plugin = {
        "pretty",
        "summary",
        "html:target/cucumber-html/index.html",
        "json:target/cucumber-report.json",
        "junit:target/cucumber.xml"
    },
    features = "classpath:features",
    tags = "not @manual and not @ignored")
public class TestRunner {
    // not required
}
```

when the following maven command is ran

```
./mvnw --no-transfer-progress clean test -Dcucumber.plugin=pretty
```

gives the following error

```
io.cucumber.core.exception.CucumberException: Only one plugin can use STDOUT, now both pretty and pretty use it. If you use more than one plugin you must specify output path with pretty:DIR|FILE|URL
```

Removing `pretty` from the CucumberOptions makes it work from the command line, but does not add the pretty formatter when run from eclipse.

**Expected behavior**
Command line options should override the CucumberOptions annotation

**Your Environment**
 - Cucumber 6.4.0 +
 - MacOS
 - Maven 3.6

mpkorstanje:
The `cucumber.plugin` options are merged rather then overridden like most others.

I don't quite understand the practical problem you are facing though. Are you intentionally trying to disable the plugins specified in `CucumberOptions` by providing only `pretty` in `cucumber.plugin`?

a1dutch:
We have a common pipeline to run all our test jobs in jenkins that tries to enable the `pretty` plugin and some others.  

The pretty plugin is also specified in the CucumberOptions.

If these are merged shouldnt the check be done after merging them together?

mpkorstanje:
The merge is pretty naive. It doesn't filter out duplicates, but that is something that could be implemented.

a1dutch:
Wow @mpkorstanje that was quick! Thankyou very much

PR #2191, opened 2020-12-14
[Core] Merge duplicate plugin options

mpkorstanje:
It is possible to declare duplicate plugins. Either through `@CucumberOptions`,
or `cucumber.plugin`, ect or a combination thereof. This can be used to ensure
certain plugins are always active during a build in CI.

Merging duplicate plugin options should ensure that plugins are not
instantiated twice with the same output.

Fixes: https://github.com/cucumber/cucumber-jvm/issues/2190


mpkorstanje:
Generated code. Thanks codecov.

Issue #2054, opened 2020-07-17
Deprecate and remove `SummaryPrinter`


mpkorstanje:
Prior to using events the `SummaryPrinter` interface provided a hook that could be used to print a summary. Now it only serves as a marker interface to indicate the plugin has something to print after a test run is done. This marker interface is used to determine if the default summary plugin should be replaced with a custom plugin. Now that only the CLI adds a default summary and progress reporter plugins the benefits of this system are marginal. 

For example:

```
java io.cucumber.core.cli.Main --plugin com.example.CustomSummary --plugin com.example.CustomFormatter
```

Will ensure that neither the default `summary` printer nor the default default `progress` formatter are used by:

```java
        RuntimeOptions runtimeOptions = commandlineOptionsParser
                .parse(argv)
                .addDefaultGlueIfAbsent()
                .addDefaultFeaturePathIfAbsent()
                .addDefaultFormatterIfAbsent()
                .addDefaultSummaryPrinterIfAbsent()
                .build(systemOptions);
```

The internal logic of Cucumber could be simplified by never enabling the default `progress` plugin and always enabling the default `summary` plugin`. The summary plugin would then have to be explicitly turned off via `--no-summary`.

This always prints a summary:

```
java io.cucumber.core.cli.Main
```

This prints a summary and progress dots:

```
java io.cucumber.core.cli.Main --plugin progress
```

This prints only progress dots

```
java io.cucumber.core.cli.Main --no-summary --plugin progress
```

This prints nothing

```
java io.cucumber.core.cli.Main --no-summary
```

This also prints nothing:

```
java io.cucumber.core.cli.Main --no-summary --plugin json:out.json
```

TODO:
 -  [ ] v6.x.x Deprecate `SummaryPrinter` interface
 -  [ ] v6.x.x  Add --[no-]summary comand line option
 -  [ ] v7.0.0 Remove `SummaryPrinter` and clean up plugin related parts of `RuntimeOptions*`.

michiboo:
Hi Could I have a go at this issue?

mpkorstanje:
@michiboo not really. This isn't scheduled until `v7.0.0` and we currently don't have a reason to make the next major version. Until that time this feature can hang around.

PR #2197, opened 2020-12-30
Deprecate the SummaryPrinter, add the '--no-summary' option

brasmusson:
This pull request implements the v7.0.0 part of #2054 

Implemented solution:
* Deprecate the SummaryPrinter interface.
* Remove the NullSummaryPrinter.
* Add the DefaultSummaryPrinter by default when using the cli runner (the core.Main class).
* Add the '--no-summary' option to be able to inhibit the adding of the DefaultSummaryPrinter when using the cli runner.
* Never add the ProgressFormatter when no formatter is explicitly specified.
* Warn when using '-p/--plugin null_summary' is used and treat it like if the '--no-summary' option was used.
* Warn (and ignore) when using '-p/--plugin default_summary' ~~or '-p/--plugin summary'~~.

Additional comments:
~~This solution assumes that there is no need to be able to add the DefaultSummaryPrinter (without using full qualified name) from the JUnit and TestNG runners.~~


core/src/main/java/io/cucumber/core/options/RuntimeOptionsBuilder.java
```diff
@@ -32,8 +31,7 @@
     private PickleOrder parsedPickleOrder = null;
     private Integer parsedCount = null;
     private Class<? extends ObjectFactory> parsedObjectFactoryClass = null;
-    private boolean addDefaultSummaryPrinterIfAbsent;
-    private boolean addDefaultFormatterIfAbsent;
+    private boolean addSummary = true;
```
mpkorstanje, line 34: I would consider printing a summary a sensible default only for the CLI.  So not enabling the summary printer by default and enabling it only in the CLI would be better behaviour. It would also avoid the need to explicitly disable this plugin in various tests again which simplifies things further.

core/src/main/java/io/cucumber/core/options/PluginOption.java
```diff
@@ -44,16 +42,13 @@
...
-        plugins.put("default_summary", DefaultSummaryPrinter.class);
         plugins.put("html", HtmlFormatter.class);
         plugins.put("json", JsonFormatter.class);
         plugins.put("junit", JUnitFormatter.class);
-        plugins.put("null_summary", NullSummaryPrinter.class);
         plugins.put("pretty", PrettyFormatter.class);
         plugins.put("progress", ProgressFormatter.class);
         plugins.put("message", MessageFormatter.class);
         plugins.put("rerun", RerunFormatter.class);
-        plugins.put("summary", DefaultSummaryPrinter.class);
```
mpkorstanje, line 56: >  This solution assumes that there is no need to be able to add the DefaultSummaryPrinter (without using full qualified name) from the JUnit and TestNG runners.

I made that same assumption in the past but people have requested a summary along with a summary of their JUnit tests. 

I don't think `--plugin summary` will be problematic though. Plugins are stored in a set. This avoids duplicate plugins.  I also don't think `--plugin summary --no-summary` is a problem either. If a bit redundant.

mpkorstanje (commented):
Cheers. I've started a v7.x.x branch to merge this into.

core/src/main/java/io/cucumber/core/options/RuntimeOptionsBuilder.java
brasmusson: Fixed in https://github.com/cucumber/cucumber-jvm/pull/2197/commits/d0ba0bbe774b2b6eee7c44156d4f5827806b4fdd

core/src/main/java/io/cucumber/core/options/PluginOption.java
brasmusson: `--plugin summary` added back in https://github.com/cucumber/cucumber-jvm/pull/2197/commits/d0ba0bbe774b2b6eee7c44156d4f5827806b4fdd

mpkorstanje:
Cheers! And congrats on being #1000.

image

PR #2204, opened 2021-01-12
Update EventPublisher javadoc

alb-i986:
Update EventPublisher javadoc to reflect changes in Scenario API.


mpkorstanje:
Cheers!

Issue #2241, opened 2021-02-25
cucumber-cdi2 does not work with weld


dcendents:
**Describe the bug**
cucumber-cdi2 does not work with weld.

**To Reproduce**
1. Edit the file cucumber-jvm/cdi2/pom.xml
2. Remove the 2 dependencies for OpenWebBeans
3. Add a dependency to Weld 3.1.6.Final
4. Build the project
5. Get the error

```
WELD-001335: Ambiguous dependencies for type CdiBellyStepDefinitions with qualifiers 
 Possible dependencies: 
  - Managed Bean [class io.cucumber.cdi2.CdiBellyStepDefinitions] with qualifiers [@Any @Default],
  - Managed Bean [class io.cucumber.cdi2.CdiBellyStepDefinitions] with qualifiers [@Any @Default]
```

**Expected behavior**
cucumber-cdi2 unit tests to pass when using weld.

**Context & Motivation**

I chose to use weld and couldn't make it work until I decided to debug the provided unit tests to understand what I was doing wrong, it seems the problem is global to cucumber-cdi2.

**Your Environment**
 - Versions used: 6.10.0 (but also main git branch)

**Additional context**
It looks like cucumber-cdi2 is adding an explicit bean for every glue class, but then the default behavior for weld is to scan for (all or annotated classes) and will complain about ambiguous dependencies when both are the same class.

I tried to exclude the class from the scan but then I get a ClassCastException on the second scenario (will open another issue for that problem).

dcendents:
Looking at the code I believe this problem is because glue classes are explicitely added:

```java
    @Override
    public boolean addClass(final Class<?> clazz) {
        getInitializer().addBeanClasses(clazz);
        return true;
    }
```

While in the deltaspike implementation it is not:

```java
    @Override
    public boolean addClass(final Class<?> clazz) {
        return true;
    }
```

I think if we want to use CDI, it is up to the user to configure the project to make sure step classes will be injected by the CDI provider, however we decide to do it (annotations, include all classes, etc).

rmannibucau:
[...]

PR #2248, opened 2021-02-27
[Cdi2,Jakarta Cdi] Add step definitions as beans when not discovered

mpkorstanje:
Depending on the CDI implementation used adding bean classes through
`getInitializer().addBeanClasses(clazz);` while also using a `beans.xml` file
to mark all classes in the module as beans results in duplicate bean
definitions. By defining step definitions only as beans when not already defined
we avoid this problem.

Fixes: #2241


mpkorstanje:
@dcendents could you do suggestion to update the `cdi2/README.md`? I'm not a frequent user of CDI so I'm not sure what the best way to explain the need for a `beans.xml` would be.

cdi2/src/main/java/io/cucumber/cdi2/Cdi2Factory.java
```diff
@@ -16,48 +16,41 @@
...
-    private SeContainerInitializer getInitializer() {
-        if (initializer == null) {
-            initializer = SeContainerInitializer.newInstance();
-        }
-        return initializer;
-    }
-
     @Override
     public boolean addClass(final Class<?> clazz) {
-        getInitializer().addBeanClasses(clazz);
```
rmannibucau, line 50: dropping this line can break some step definition (anyone used as injection in other steps or using cdi some extensions for their injections) so not sure it is desired

mpkorstanje: I've added a test for this. I could only break it when I don't include `beans.xml`. Do you know other ways to break it?

https://github.com/cucumber/cucumber-jvm/pull/2248/files#diff-5fd981ab52864d56c2d8468af367ef1eb48cd22c368525a10bb403db7d10d741R104

dcendents: The steps inject  the local test bean Belly so a beans.xml will always be required.
To properly test it we would need a project that injects beans from the main code or from libraries only.

rmannibucau: A way to ensure it is as expected is to write a CDI extension which veto all undesired managed types and add the desired ones (bm.addAnnotatedType(bm.createAnnotatedType(MyType.class))), this way whatever the scanning is, the test is "as expected".

dcendents: Here is a quick test project without a test beans.xml: https://github.com/dcendents/cucumber-cdi2-test
It works fine, getInstance creates UnmanagedInstances, beans are injected.

cdi2/README.md
```diff
@@ -27,7 +27,7 @@ And for Weld it is:
 <dependency>
   <groupId>org.jboss.weld.se</groupId>
   <artifactId>weld-se-core</artifactId>
-  <version>3.1.1.Final</version>
+  <version>4.0.0.Final</version>
```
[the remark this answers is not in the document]
mpkorstanje: Ah good point. We'll have to repeat this for the `jakarta-cdi` implementation as well.

cdi2/src/main/java/io/cucumber/cdi2/Cdi2Factory.java
dcendents: I just thought of something, what about calling `clazz.getClassLoader().getResourceAsStream("META-INF/beans.xml");` and add the class only if the file is not there?

I think this way we have the best of both worlds. (And the classes need to be cached and added to every cdi container).

rmannibucau: Right but bypassing add class and usong a "try to find cdi bean else fallback on unmanaged instance trick" can work better so not that crazy to think that way ;)

dcendents: I have pushed a version that does that, feel free to improve it but the logic works.
It checks every class to see if they are managed, for that I need to start a CDI container.

1. Every class need to be checked, step classes could come from multiple jars on the classpath.
2. Initialization takes a bit longer though

It would be more efficient to keep the container running while we check all classes, I see a few options:

1. Keep the first container running and use it for the first scenario, we need to add unmanaged classes directly on it as we get them. Is it safe to do that (adding classes after initialize method has been called)?
2. Keep the container running while we receive classes and stop it before the first scenario. Is there a lifecycle hook we can use to do that?
3. Use a different variable (e.g.: `checkClassesContainer`), but we need to stop it at some point, either when stop is called or on the first getInstance.

dcendents: And we need a unit test to veto some beans as @rmannibucau mentionned

mpkorstanje: > I have pushed a version that does that, feel free to improve it but the logic works.

That's pretty clever but stopping/starting the application a couple of hundred 100times may kill performance somewhat. 

I'd rather make `beans.xml` a requirement.

rmannibucau: Will make it harder to reuse the same test suite accross env/apps (typically for contract validation).
I think we can make it start once for all too if that it a concern.

dcendents: Personally I lean toward @mpkorstanje, I'd say we have to make `beans.xml` a requirement, after all if you want to use CDI this is not a foreign concept. There are other alternatives that can be used like `PicoContainer` as recommended:

>PicoContainer (The recommended one if your application doesn’t use another DI module)

I also like to be the devil's advocate: 

Let's say I use CDI in my project, therefore I **need** to use CDI for my tests as well. I also want to include a third-party test-jar with shared Step classes in it, but it is not discoverable (no `beans.xml` in the jar). Step classes will still be loaded (as `UnmanagedInstances`). 

Now the only advantage of calling addClasses on the CDI initializer is to be able to inject step classes as beans. But if we _need_ to do that, we can always create our own `Producer`.

I though I'd find a way to do both but in my opinion it adds a lot of complexity for no rewards. I would vote to revert my modifications and improve the project documentation so people are aware they need to include a `beans.xml` file.

rmannibucau: @dcendents where I kind of disagree is on the "write your own producer". Technically it is true but a lot of work in practise when you reuse "stack" steps.
That said the perf issue is likely trivial to solve so I think it makes sense to fix this one rather than working around cucumber contract.

dcendents: > Technically it is true but a lot of work in practise when you reuse "stack" steps.

```java
	@javax.enterprise.inject.Produces
	public BeanInterface produceBellyStepDefinitions() {
		return new BeanImpl();
	}
```

It is one annotation above a factory method, return type is the type of bean to create, this is the standard way to inject third-party POJO.

rmannibucau: @dcendents it is one annotation so to make steps working you must

1. write as much producer as steps (N)
2. do the injections yourself through the producer parameter + setters/reflection/.... (M) most of the time since beans are not supposed to be reinjected anytime later

so complexity of this "trivial" solution is N.M which is why I don't see it as a solution but a workaround.

dcendents: > @dcendents it is one annotation so to make steps working you must
> 
> 1. write as much producer as steps (N)
> 2. do the injections yourself through the producer parameter + setters/reflection/.... (M) most of the time since beans are not supposed to be reinjected anytime later
> 
> so complexity of this "trivial" solution is N.M which is why I don't see it as a solution but a workaround.

Not completely accurate.

Once you do the producer the bean is handled by the CDI container as any other managed bean, including its internal injections. But really if theses steps use CDI injection they should have a `beans.xml` file. If they use a different DI then it will not be compatible with CDI anyway.

But yes you need N producer methods to **inject** N third-party unmanaged steps.

But I still can't understand the use case, **why would I want to inject steps at all?**
Cucumber will load the step and execute it, why would I inject a step and manually call it?

dcendents: @rmannibucau @mpkorstanje Actually I just found the solution and pushed a commit.

When the first container is created and only if it reports ambiguous beans, I test every single addedClasses and remove them if they are ambiguous.

The only performance drawback is that the container might be started twice on the first scenario, and **only** if there are ambiguous beans.

Let me know if you are both happy with that solution, it still needs a unit test though.

dcendents: I'm not allowed to update the workflow but I believe this command would be better to test weld (from the root folder):

`mvn verify -pl cdi2,jakarta-cdi -Pcdi2-weld,jakarta-cdi-weld`

dcendents: Sorry to play ping pong with the code, I spoke too fast yesterday and clearly didn't test properly with weld. I thought I had it but it's not working. When steps are injected, Weld tries to resolve them when the container is initialized and throws an exception for duplicates.

But the first idea is often the best, so I figured out how to properly find all beans.xml files from the classpath. 
- I get the lib path (jar or directory) of all the beans.xml files. 
- In `addClasses` I get the class lib path (jar or directory)
- If the lib is not managed, then it is added to the container synthetic beans (`addBeanClasses`)

I tested different scenarios locally, with injected steps coming from a dependent lib. Both from a folder classpath and from a jar file. This is easy enough to do in eclipse but I have no idea how we can test it completely with unit tests. I think we need some more complex integration tests to test all the permutations.

rmannibucau: Hmm, this is what the container does (+other rules) when scanning is enabled. Issue redoing it is you must ensure to *read* the beans.xml (scan=none or exclude cases), scan jars without beans.xml (annotated mode), respect container scanning rules (owb Filter for ex). This is a lot of work IMHO whereas adding steps is easier. Duplicates issue can be something else since this just impacts the scanning and should be unified.

dcendents: @rmannibucau can we agree that we should not add synthetic beans if the lib (dependency or current code) has a `beans.xml` file, therefore it has been configured already by someone that know better?

This is what the code does:
1. `beans.xml` found: Do not interfere by adding synthetic beans, they will be discovered (or ignored as configured explicitely by the user in `beans.xml`).
2. no `beans.xml` present: Add steps as synthetic beans so the module can inject them.

Otherwise I give up, IMHO this module helps integrate CDI, this is not a full CDI implementation. 

This issue was created because the code was blindly adding synthetic beans without caring about the user configuration. This implementation is already much better and respects the user configuration and only adds synthetic beans when the user did not explicitely configure the CDI container by adding a `beans.xml` file.

mpkorstanje: I'm not happy with the complexity. I would much rather have users add a `bean.xml` file and be done with it. 

I'm afraid I won't have much time this week to help out.

dcendents: I agree with you @mpkorstanje but it seemed important to @rmannibucau and was trying to please you both.

Can cucumber-cdi2 load POJO steps classes without a `beans.xml` file and adding synthetic beans? YES

But in my opinion as soon as someone writes `@Inject` in their code it's normal that they also include a `beans.xml`, this is part of the CDI spec after all.

I'll revert my changes then. It wasn't all for nothing, I found a way to execute tests without the `beans.xml` file so I'll keep that at least it is convenient to test loading POJO step classes.

rmannibucau: > But in my opinion as soon as someone writes @Inject in their code it's normal that they also include a beans.xml, this is part of the CDI spec after all.

No, it is part of @inject spec which is used by CDI but also most of other IoC and this is how some compatibilty test suites are written.

To summarize:

1. Step classes can be added or not -> yes
2. Step classes can be injected without being added -> yes
3. The "@inject" case can be managed with an "integration CDI extension" -> I agree

So here is what can maybe be a compromise:

1. The addClass is controlled/replaced by a package filtering (default to nothing)/initializer.addPackage(true). It means steps will not be managed by default but can become managed - this is the role of the code handling the container so cucumber there.
2. the lookup does not rely 100% on unmanaged but on "find cdi bean, if not use unmanaged"

This implementation drops a lot the complexity and will make all cases working I guess.

Another alternative is to add a SeContainerInitializerConfigurator { void configure(SeContainerInitializer, Collection<Class<?>> steps); } as a SPI, would move the responsability to the test suite coder but solve it too.

dcendents: > No, it is part of @Inject spec which is used by CDI but also most of other IoC and this is how some compatibilty test suites are written.

But it is required by the CDI spec: http://www.cdi-spec.org/faq/
Look at: What is beans.xml and why do I need it?

Also: https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html

What you call `@Inject` spec does not exist, it might be called that by some but CDI is a Java spec.
`@Inject` is just a java annotation that CDI recognises, and other frameworks can use the same annotation but that doesn't make them a CDI implementation.

https://weld.cdi-spec.org/
"Weld is the reference implementation of CDI: Contexts and Dependency Injection for the Java EE Platform - a JCP standard for dependency injection and contextual lifecycle management and one of the most important and popular parts of the Java EE"

https://openwebbeans.apache.org/
"Apache OpenWebBeans delivers an implementation of the Contexts and Dependency injection for Java EE (CDI) 2.0 Specification (JSR-365)."

As a user, if I want to use CDI, at a minimum I need:
- `beans.xml` file in every lib part of the CDI
- write `@Inject` on beans I want injected

If I deploy code in a web or application container that might be it. Nothing else is required.
e.g.: Deploy JAX-WS service on Wildfly and inject beans in it.

If I run in a plain JavaSE environment then I also need to:
- include a CDI implementation on my classpath
- Initialize the container and get beans from it

So as a user, there is no way around having a `beans.xml` file if I want to use CDI.

At this point I hope we can agree CDI is a spec and `beans.xml` is required.


Now let's talk about the tests:

I've used cdi-unit in the past and it is true I did not need a `beans.xml` file in my test resources. Why? Because with plain junit tests I only need to inject production code into my class. I also could easily declare mocks as alternative beans and get them injected into the code to test. All in all, everything about my tests was contained in a single class as is usual with junit tests.

With cucumber it makes sense to split step classes around features and break the 1 test class for 1 production class paradigm. Then we have to inject a shared state object between those steps.

Imagine the following very simple test structure:

- io.cucumber.StepClass1
- io.cucumber.StepClass2
- io.cucumber.PojoSharedState (injected into the 2 teps classes)

With the current cucumber-cdi2 version 6.10.0, is it possible currently to run those tests without having a `beans.xml` file: NO it is not possible.

`StepClass1` and `StepClass2` are added as synthetic beans, but not `PojoSharedState`, cucumber does not know anything about that class, it does not have cucumber annotations in it.

So great if cucumber-cdi2 says, "You can start using CDI in your tests and we'll do some automatic wiring for you and it will work even if you don't follow the spec. As long as you don't inject test POJO though."

But where I don't agree (and is why I opened #2241), is that if you want to use Weld (the reference CDI implementation) it will not work because our automatic wiring broke it. You will have to start editing that `beans.xml` file, figure out what cucumber-cdi2 does under the hood and explicitely ignore these classes from the discovery process. And everytime you add a step class don't forget to go and ignore it in `beans.xml`.

Now to recap different `@Inject`ion scenarios:

- Should a `beans.xml` file be required in `src/test/resources`?
    - Step classes that `@Inject` classes from other libs (and transitive injection):
        - currently (6.10.0): NO (but it is required in `src/main/resources` and every jar part of the injection)
        - proposed (7.0.0): NO (but it is required in `src/main/resources` and every jar part of the injection)
    - Step classes that `@Inject` local pojo test classes (and transitive injection):
        - currently (6.10.0): YES
        - proposed (7.0.0): YES
    - Step classes that `@Inject` other step classes (and transitive injection):
        - currently (6.10.0): NO
        - proposed (7.0.0): **? This is what all this is all about**

I've asked the question a few times before: **Why would I ever need to inject a cucumber step class into another cucumber step class?** In my opinion this is a very bad design and should be refactored, test state should not be saved in step classes.

Now you both @mpkorstanje and @rmannibucau disagree, I'm very new to cucumber so I don't know how this should be resolved.

My vote is to keep the code lean and stop messing with synthetic beans.
It is already great that step classes are loaded as `UnmanagedInstance`s and allow injection of classes outside of the local test scope.

Is it reasonable to spend so much time and effort to allow injection of step classes into other step classes (bad design), decide to use CDI (and not PicoContainer or other alternatives) but don't want to add the required `beans.xml` as per the CDI spec?

P.S.: Should the discussion moved to Slack?

rmannibucau: @Inject is NOT CDI but CDI uses @Inject, don't reverse or merge it. This is also why some steps are written portably and tested against CDI, Spring, Guice and a few other IoC. You have CDI so you have @inject but you can miss all other parts of CDI like its activation (beans.xml).

> What you call @Inject spec does not exist, it might be called that by some but CDI is a Java spec.

It does, it is the JSR330: https://javax-inject.github.io/javax-inject/.

> As a user, if I want to use CDI, at a minimum I need:
> 
> beans.xml file in every lib part of the CDI
> write @Inject on beans I want injected

and

> So as a user, there is no way around having a beans.xml file if I want to use CDI.

No again, CDI 2 enables to no more write/create beans.xml and you don't have to use @inject as this cucumber integration shows - steps are looked up without @inject generally.

AFAIK with the addClass the case I described works as of today and is broken by removing addClass.

> Why would I ever need to inject a cucumber step class into another cucumber step class?

This is a fair question and it can happen to reuse some steps (aliases/rephrasing for a better human writing experience) but you also have the case where you have a transversal bean checking data or resetting in steps with a scope (or not) which is purely technical but requires the step to be made a bean.

> My vote is to keep the code lean and stop messing with synthetic beans.

Agree, this is why I proposed to enable archives with "just" replacing the addClass by a addPackage(true, xxx) so adding a config to the integration instead of trying to be clever.

> It is already great that step classes are loaded as UnmanagedInstances and allow injection of classes outside of the local test scope.

This is true but on the other side, if you do what you describe, ie make them actual bean, it is not what you want, you want to lookup the bean properly, this is why I proposed to use a try to lookup and if not possible use unmanaged instance.

> Is it reasonable to spend so much time and effort to allow injection of step classes into other step classes (bad design),  decide to use CDI (and not PicoContainer or other alternatives) but don't want to add the required beans.xml as per the CDI spec?

Well it is not only about steps into steps - even if this case makes sense from time to time I could live without it - but also being able to use step as beans which is what the addClass enables (but not yet the lookup which I lacked in a few projects already).
And once again CDI does NOT need a beans.xml as per spec but has some funny scanning rules. That said the point is to be able to use steps "as this", don't assume you own all the code you run, it only happens when you are lucky but it seems I'm not as lucky as you on this point ;).

dcendents: then my bad to think that cucumber-cdi2 was a CDI 2.0 injection framework for cucumber.
I suggest the module be renamed to cucumber-javax-inject. ;-)

Seriously I'm not an expert on the matter, but javax.inject is JSR-330, CDI 2.0 is JSR-365 which includes (and respects) JSR-330: https://javaee.github.io/tutorial/cdi-basic002.html

Now unless I'm mistaken `javax.enterprise.inject.se.SeContainerInitializer.SeContainerInitializer.newInstance()` is part of the CDI and not the javax.inject spec.

`javax.enterprise.inject.se.SeContainerInitializer` is from cdi-api-2.0.SP1.jar
`@javax.inject.Inject` comes from javax.inject-1.jar

So by using `SeContainerInitializer` this is a CDI 2.0 module and therefore must respect the spec, at least not break it (again issue #2241)

I think part of the problem is that openwebbeans seems very lax about the CDI spec compared to Weld. Weld is the reference implementation and is used in most application containers so it makes sense that we test properly against it. If enabling a feature breaks Weld then in my opinion this is not acceptable.

Weld 1.0 was following javax.inject spec (I never used Weld 1 so wasn't even aware of that spec)
Weld 2.0 is CDI 1.1
Weld 3.0 is CDI 2.0
Weld 4.0 is CDI 3.0

cucumber-cdi2 project builds with Weld 3.X, jakarta-cdi (CDI 3.0) builds with Weld 4.X

rmannibucau: > then my bad to think that cucumber-cdi2 was a CDI 2.0 injection framework for cucumber.
> I suggest the module be renamed to cucumber-javax-inject. ;-)

I can't understand what you mean there. cucumber-cdi2 is about running in a CDI container with the CDI 2 SE API configuration support. This means it must also support @inject spec since CDI includes it (and once again not the opposite).

> So by using SeContainerInitializer this is a CDI 2.0 module and therefore must respect the spec, at least not break it (again issue #2241)

I see where you come from but to be a CDI container you must pass CDI validations (TCK) but also @inject validation. This means you can reuse @inject modules in CDI features without having to have these steps CDI steps. Hope it is clearer written this way.

> I think part of the problem is that openwebbeans seems very lax about the CDI spec compared to Weld. 

Hmm, it is not, both pass the same test kit but weld tend to be broken in several environments and more restrictive (but it is not in the spec). I agree we must run with weld - and we were at some point, didn't check recently. We can refine the error you get if you want but in current form it should be portable.

dcendents: @rmannibucau well I suppose we disagree on what cucumber-cdi2 should do because I have no idea what is the contract between these extensions and cucumber-core. I'm very new to the project.

But I think I nailed the solution this time. Not only does my latest commit fix #2241, it doesn't break any functionality of 6.1.0, it also makes it possible to inject most POJO objects without a `beans.xml` file.

I register a CDI Extension and test if all the step classes can be resolved, if not I add them as a bean on the BeanManager. So this doesn't break Weld and it means `addClasses` does what it should. This is also a supported way to work with CDI  so no voodoo magic here.

Now the cherry on top: for each class added manually, I introspect the injection points and test them as well, so now we can `@Inject` POJO classes without a `beans.xml` file, as long as the injected types are valid.
i.e.: if it works with a `beans.xml` file, it should also work without. if it is broken with a `beans.xml` file, then it will be broken without it, I'm not doing any magic.

jakarta-cdi/src/main/java/io/cucumber/jakarta/cdi/CdiJakartaFactory.java
```diff
@@ -69,4 +76,51 @@ public boolean addClass(final Class<?> clazz) {
...
+    ) {
+        if (!unmanagedClasses.contains(clazz) && bm.getBeans(clazz).isEmpty()) {
+            unmanagedClasses.add(clazz);
+
+            final InjectionTarget injectionTarget = addBean(afterBeanDiscovery, bm, clazz);
+
+            Set<InjectionPoint> ips = injectionTarget.getInjectionPoints();
+            for (InjectionPoint ip : ips) {
+                Type type = ip.getType();
+                if (type instanceof Class) {
```
mpkorstanje, line 101: What happens if I try to inject an unmanaged parametrized type e.g. `KafkaProducer<String, Pojo>`?

mpkorstanje: I've added a warning if/when this happens and the parameterized bean isn't managed by having a beans.xml. Though that makes me think it should be possible to do so. Not necessary to fix right now though. Some one else can do that.

rmannibucau: This is supported by cdi so no additional warning is needed

mpkorstanje (commented):
I'm quite happy with the current solution. Looks like it doesn't even break anything so we can release this in v6. The only thing left would be a small update to the `README.md` of both projects to explain what we're doing.

jakarta-cdi/src/main/java/io/cucumber/jakarta/cdi/CdiJakartaFactory.java
mpkorstanje: Currently the implementation uses:

```
AnnotatedType clazzAnnotatedType = beanManager.createAnnotatedType(clazz);
```

This only accepts class types and does not support parametrized types. Could you propose a change that makes it work?

dcendents: Good catch with parameterized types, I did not think about it.
Actually a bean can satisfies multiple types and there's a whole set of rules about which bean can satisfies the injection point.

I've modified the code to support parameterized types and adjusted the tests, but I think we need more tests to make sure it works correctly, what I have in mind is having multiple injection points (in multiple beans) but with different parameterized types and also without any types at all.

With the current code I think we will inject a bean for each and I want to make sure we will not get a AmbiguousDependency exception.

I'll try to add those tests tonight or tomorrow.

rmannibucau: addBean() is the entry point, you set beanClass(rawType).types(paramType, Object.class).scope(Default.Literal.INSTANCE, Any.Literal.INSTANCE).scope(Dependent.class).id(whatyoulike).createWith(c -> ...).
Types is what is used, bean class is mainly for proxies when needed - generally not here.

dcendents: Also it is quite safe to add unmanaged beans for step classes but as soon as we resolve new types to satisfy injection points we are bound to get some unthought situation. 

I think it would be safer to log a message and let the user know we recommend adding a `beans.xml` file (or enable bean discovery, there might be a `beans.xml` already).

rmannibucau: the standard trick is to add a qualifier on the injection point (@CucumberInternal for ex) you handle and add it as well when registering the bean, avoids ambiguous situations. To be concrete any warning in a CDI env = a deployment error so logging it is pointless normally.

mpkorstanje: Good points. I'd like to prevent is a situation where we create a passing test but a failing application by creating beans for all unsatisfied dependencies.

We're currently recursively going through the step definitions and their dependencies. I'm thinking that perhaps instead of recursively adding all dependencies, we only add the step definitions and check if the dependencies can be injected. If not, then we can throw an informative error.

dcendents: I think we should log a warning as soon as we create a bean definition to let the user know it would be better to add a `beans.xml` (and if necessary create Producers) to take full control.

One scenario I was thinking was a user (let's call him John) including a shared test-jar with a simple Step class that `@Inject`s a POJO. Let's say they were using guice to do the injection so they did not include a `beans.xml` file in the test-jar. 

Now John cannot simply add a `beans.xml` because the step class comes from a test-jar dependency. He also needs to create a `@Producer`. So even though I'm sure there are some situations that we won't be able to configure correctly, I believe that for all the Johns in the world that are not so familiar with CDI it can still be useful to them if we try to create beans for all the unmanaged injection points.

For the more eccentric cases, they will get a warning from us and an error from the CDI container.

rmannibucau: I wouldnt do something i consider as a warning here. CDI gives you all data to know it will be an error or not so there is no warn case I think. But logging an info or even debug message that a custom bean was registered sounds ok to me.

mpkorstanje:
We may have some irreconcilable design goals:

1.  As testing framework it is important that we don't interfere with the application under test. This isn't entirely possible to avoid since we are using the same dependency injection container but we should be quite conservative in what we do.

     So while I'm quite comfortable registering step definitions as beans when not picked up by the framework already, it gets more complicated when dealing with any other classes. Now I don't quite know enough about CDI to make a working example but I can imagine a situation where the application dynamically creating or looking up beans might be confused by us manually creating additional beans that wouldn't exists if it weren't for us creating them.
     
     And while it might be reasonable to say that this would be very poor application design or bad use of CDI, a testing framework is still not expected to interfere.

2. It is important to inform the user about what the framework is doing. But given the importance this non-interference principle I don't think a log message is sufficient in the proposed implementation. Log messages get overlooked. Especially amidst all the other perfectly ordinary log messages that the container emits. It is only when something breaks that people will look at the log. However we are currently creating a situation where things _just work_ and people will not look at the log.

3. As a framework in general developer ergonomics are important and we would like dependency injection into step definitions to _just work_. This poses a problem when the dependencies of step definitions are not automatically picked up by the CDI implementation (for example due to a lack of beans.xml).

4. As contributors we don't want to break existing implementations on a minor version release. Though I think that if we don't only add the step definition dependencies as beans we can still release on a minor version.

    From what I can tell we currently we only add step definitions as beans regardless where they came from. So depending on the CDI implementation, without a beans.xml  their dependencies could still not be injected if they came from a test-jar or some other.

So I'm currently leaning towards dropping a little bit of the proposed developer ergonomics in favour of non-interference but I'm also doing to have to find some time to play around with CDI and understand a bit more what is going on. I'm afraid this will not be for a little while.

@dcendents given that this item will not be resolved for some time, would you benefit from a release with the currently unreleased patches or would you rather wait until we've also resolved this item?

rmannibucau:
Hmm

1. Agree and this means injection points observer can be skipped as mentionnned probably
2. Think either we make it work or fail, there is no other case in the impl so logging is always useless no?
3. This is not true as current impl does - even worse: we want step to be injectable even when not part of the app/cdi bean to enable step libraries and this is what does the impl
4. Agree so maybe dont add steps as bean for next one

dcendents:
@mpkorstanje Following up on your comment I took the liberty to split this PR in two.
I have openend #2256 with the current commits where we can discuss further if/how we should process injection points.

I have removed the processing of injections points from this branch. Which means in its current state this PR will resolve #2241 and there are no regression or added feature from 6.10.0.
(Step classes are added as synthetic beans only when CDI did not already resolve them).

I'd appreciate if this PR could be merged in the next release.

mpkorstanje:
The pro-activeness is appreciated. I've updated the docs with a minimal example explanation of these modules. 

Please give this PR a last good look. If there are no further comments I'll have this merged and released.

dcendents:
It looks good to me

Thanks

PR #2251, opened 2021-03-02
[Cdi2/Jakarta Cdi] Run tests for OpenWebBeans and Weld in CI

dcendents:
@mpkorstanje this is an alternate maven configuration that lets you run the tests with both openwebbeans and weld.
It is also possible to enable just one implmentation with the profiles `cdi2-openwebbeans` and `cdi2-weld` like before.

The only issue is that the default profile will break the tests in the IDE.
I use eclipse and to be able to run unit tests I have to activate the `cdi2-openwebbeans` or `cdi2-weld` profile in the project configuration.

I let you judge if you like it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,8 +30,6 @@ jobs:
         run: mvn install -DskipTests=true -DskipITs=true -Darchetype.test.skip=true -Dmaven.javadoc.skip=true -B -V --toolchains .github/workflows/.toolchains.xml
       - name: test
         run: mvn verify -P-spotless-apply --toolchains .github/workflows/.toolchains.xml
-      - name: test cdi with openwebbeans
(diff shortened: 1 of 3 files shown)


mpkorstanje (changes requested):
I like this. But I don't like that it breaks in an IDE. I would like the project to keep working by default without clicking options.

So what if `cdi2-openwebbeans` is active by default and `cdi2-all-implementations` is active in CI? We know a test is running in CI when the `CI` environment variable is true.

https://docs.github.com/en/actions/reference/environment-variables

dcendents:
> So what if `cdi2-openwebbeans` is active by default and `cdi2-all-implementations` is active in CI?

I've done that and for jakarta-cdi as well, just checked the build logs and it worked, the profile was activated.

If you want to test locally:

- `mvn help:active-profiles`
- `CI=true mvn help:active-profiles`

In eclipse the tests pass with openwebbeans

I cannot modify the github build workflow though.

mpkorstanje:
Much appreciated!

PR #2256, opened 2021-03-07
[cdi2, jakarta-cdi] Cdi injection points

dcendents:
Follow-up on #2248

### DIFF
--- a/cdi2/src/main/java/io/cucumber/cdi2/Cdi2Factory.java
+++ b/cdi2/src/main/java/io/cucumber/cdi2/Cdi2Factory.java
@@ -1,6 +1,8 @@
 package io.cucumber.cdi2;
 
 import io.cucumber.core.backend.ObjectFactory;
+import io.cucumber.core.logging.Logger;
+import io.cucumber.core.logging.LoggerFactory;
 import org.apiguardian.api.API;
 
 import javax.enterprise.context.spi.CreationalContext;
@@ -12,9 +14,13 @@ import javax.enterprise.inject.spi.AfterBeanDiscovery;
 import javax.enterprise.inject.spi.AnnotatedType;
 import javax.enterprise.inject.spi.BeanManager;
 import javax.enterprise.inject.spi.Extension;
+import javax.enterprise.inject.spi.InjectionPoint;
 import javax.enterprise.inject.spi.InjectionTarget;
 import javax.enterprise.inject.spi.Unmanaged;
+import javax.enterprise.inject.spi.configurator.BeanConfigurator;
 
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
@@ -23,6 +29,7 @@ import java.util.Set;
 @API(status = API.Status.STABLE)
 public final class Cdi2Factory implements ObjectFactory, Extension {
 
+    private static final Logger log = LoggerFactory.getLogger(Cdi2Factory.class);
     private final Set<Class<?>> stepClasses = new HashSet<>();
 
     private final Map<Class<?>, Unmanaged.UnmanagedInstance<?>> standaloneInstances = new HashMap<>();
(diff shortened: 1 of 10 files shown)


dcendents:
From @mpkorstanje on #2248

Following the thread #2248 (comment) I get the impression we may have some irreconcilable design goals:

As testing framework it is important that we don't interfere with the application under test. This isn't entirely possible to avoid since we are using the same dependency injection container but we should be quite conservative in what we do.

So while I'm quite comfortable registering step definitions as beans when not picked up by the framework already, it gets more complicated when dealing with any other classes. Now I don't quite know enough about CDI to make a working example but I can imagine a situation where the application dynamically creating or looking up beans might be confused by us manually creating additional beans that wouldn't exists if it weren't for us creating them.

And while it might be reasonable to say that this would be very poor application design or bad use of CDI, a testing framework is still not expected to interfere.

It is important to inform the user about what the framework is doing. But given the importance this non-interference principle I don't think a log message is sufficient in the proposed implementation. Log messages get overlooked. Especially amidst all the other perfectly ordinary log messages that the container emits. It is only when something breaks that people will look at the log. However we are currently creating a situation where things just work and people will not look at the log.

As a framework in general developer ergonomics are important and we would like dependency injection into step definitions to just work. This poses a problem when the dependencies of step definitions are not automatically picked up by the CDI implementation (for example due to a lack of beans.xml).

As contributors we don't want to break existing implementations on a minor version release. Though I think that if we don't only add the step definition dependencies as beans we can still release on a minor version.

From what I can tell we currently we only add step definitions as beans regardless where they came from. So depending on the CDI implementation, without a beans.xml their dependencies could still not be injected if they came from a test-jar or some other.

So I'm currently leaning towards dropping a little bit of the proposed developer ergonomics in favour of non-interference but I'm also doing to have to find some time to play around with CDI and understand a bit more what is going on. I'm afraid this will not be for a little while.

dcendents:
From @rmannibucau on #2248

Hmm

1. Agree and this means injection points observer can be skipped as mentionnned probably
1. Think either we make it work or fail, there is no other case in the impl so logging is always useless no?
1. This is not true as current impl does - even worse: we want step to be injectable even when not part of the app/cdi bean to enable step libraries and this is what does the impl
1. Agree so maybe dont add steps as bean for next one

dcendents:
To addess @mpkorstanje concerns, 

- I think what we do is still quite safe and should not interfere with the application under test. We are only adding beans for step classes and injection points that are not managed by the CDI container. 

- In fact I see the opposite being more of a problem, let's assume the user has a working CDI environment and is using `weld-unit`. By switching the dependency to `cucumber-cdi2` we start interfering with the CDI container and add unmanaged step classes. If the injection points in the step classes cannot be resolved by the container, the tests will crash while they worked previously.

@rmannibucau for the reasons above I think it is important we notify the user when we alter the CDI configuration by adding unmanaged beans and we recommend they make sure all tests classes are CDI discoverable by adding a `beans.xml` file and creating `@Producer`s if necessary. Maybe INFO level message is enough.

In all cases if something goes wrong, the CDI container will throw an exception which the user can use to understand/resolve the problem.

The important point for me is: If the user adds a `beans.xml` file and create `@Producer`s for all CDI clasees coming from dependencies without `beans.xml` file, then `cucumber-cdi2` will not interfere with the CDI configuration as it will not be necessary.

mpkorstanje:
>  We are only adding beans for step classes and injection points that are not managed by the CDI container.

But we are never able to answer the question if that bean was unmanaged because the user uses it to share information between steps and never registered it or if it was unmanaged because the user changed the configuration in such a way that they became unmanaged and should have been removed.

The problem is fundamentally that by trying to be helpful we are making assumptions about things outside of Cucumbers domain. There will always be situations where these assumptions are wrong.

> If the injection points in the step classes cannot be resolved by the container, the tests will crash while they worked previously.

Provided we provide a way to add these classes I think this is acceptable. I believe this can be achieved by using the `SeContainerInitializer` as @rmannibucau implemented in #2257. Is this correct?

rmannibucau:
@mpkorstanje yes, if we enable to provide a custom initializer, we are done. If not we must try to be clever as you mention and it can be hard until we override the scanning of CDI SE to be configured by a cucumber config (not sure we prefer that over a custom initializer, don't think so but both are strictly equivalent in terms of feature covering).

dcendents:
Sorry for my late reply, yes I think explicit configuration as in #2257 would cover all cases where we want to register a bean that is not under our control (third-party library). 

This PR also contains test code that enabled testing without the `beans.xml` file, is this something you want to keep? I could close this PR and the test classes to #2257 maybe?

mpkorstanje:
I've written #2439. Let's see if any one wants to contribute a solution like that.

PR #2257, opened 2021-03-07
enable to customize cdi2 container initializer

rmannibucau:

Enable to use the CDI container initializer configuration using system properties.

Read cucumber.cdi2.* properties and wire them + a SPI to make it code first (easier in some cases, in particular for libraries).

### DIFF
--- a/cdi2/README.md
+++ b/cdi2/README.md
@@ -42,3 +42,15 @@ If you don't know which one to use, you can import the following one but if you 
   <version>2.0</version>
 </dependency>
 ```
+
+## Configuration
+
+System properties are used to customize the CDI `SeContainerInitializer`, here is the list:
+
+* `cucumber.cdi2.packages`: list of packages (comma separated) added in the container discovery, must contain a `package-info.class` file to be deterministic.
+* `cucumber.cdi2.recursivePackages`: same as previous one but recursively.
+* `cucumber.cdi2.classes`: list of classes added in the discovery.
+* `cucumber.cdi2.extensions`: list of extension classes added to the container (if not automatically registered).
+* `cucumber.cdi2.<key>`: will add the system property as a key/value property of the initializer (container specific).
+
+Note that you can also implement `io.cucumber.cdi2.spi.SeContainerInitializerCustomizer` and register your implementation by putting its fully qualified name in `META-INF/services/io.cucumber.cdi2.spi.SeContainerInitializerCustomizer` to customize the initializer.
--- a/cdi2/src/main/java/io/cucumber/cdi2/Cdi2Factory.java
+++ b/cdi2/src/main/java/io/cucumber/cdi2/Cdi2Factory.java
@@ -1,5 +1,7 @@
 package io.cucumber.cdi2;
 
+import io.cucumber.cdi2.internal.CustomizableInitializer;
+import io.cucumber.cdi2.spi.SeContainerInitializerCustomizer;
 import io.cucumber.core.backend.ObjectFactory;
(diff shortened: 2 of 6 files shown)


mpkorstanje (changes requested):
So a few problems:

1. System properties aren't quite universal. We have a number of different ways to pass properties, this would be yet another.
2. I'm not happy with all the logic involved in `findPackage` and `toClasses`. I would rather see that we burden the user of Cucumber with these problems. 

I would suggest we do something like the [`InjectorSource`](
https://github.com/cucumber/cucumber-jvm/blob/main/guice/src/main/java/io/cucumber/guice/InjectorSource.java) in `cucumber-guice`. And I would suggest doing something like we do with `cucumber-spring` where the [`SpringBackend`](https://github.com/cucumber/cucumber-jvm/blob/main/spring/src/main/java/io/cucumber/spring/SpringBackend.java#L28) tries to find a specific class and the [`SpringFactory`](https://github.com/cucumber/cucumber-jvm/blob/main/spring/src/main/java/io/cucumber/spring/SpringFactory.java#L66) looks out for that class.

So as a quick sketch:


```java
package com.example.stepdefinitions;


class MyCucumberSeContainerInitializer implements CucumberSeContainerCustomizer {

    public void customize(SeContainerInitializer sci) {
       // do customizations here    
    }

}
```

rmannibucau:
@mpkorstanje
1. I'm fine reading properties from anywhere, not sure where you prefer here. overall rational is that configuring system properties to *refine* a container (and not always configure "first") is very neat since it can be done in surefire or any test launcher just by configuration and build launcher overrides (maven for ex) whereas other solutions are more intrusive AFAIK.
2. about toClasses I'm not sure, it is quite common and direct. About findPackage I can agree but it is also what CDI impls do generally so don't think it is a big deal and it enables the feature which is quite common. Worse case we can document it better I guess? What should absolutely avoided is a 3rd scanning IMHO (cucumber scans, cdi will scan, we don't want to scan another time).
3. the SPI option was nice cause it enables to compose them instead of just having a single class - which means coding again to import multiple customizers. That said as a first step a class can be ok-ish, will just enforce code for each setup and do a custom composition instead of a built-in one but it solves the "setup" issue this PR was coming from.

mpkorstanje:
I'm not suggesting SPI, I'm suggesting the customizer is part of the glue. This means there can be multiple customizers and depending on how Cucumber is configured different once are used.

rmannibucau:
@mpkorstanje I see, I'm fine with that for a first version. Do you want to take over the PR to do it?

mpkorstanje:
I wasn't intending to.

Would you be interested in adding a cdi2 backend that looks for implementations of `io.cucumber.cdi2.SeContainerInitializer` and adds them to the object factory?

rmannibucau:
@mpkorstanje I played a bit more with such pattern and came back to an old issue (fixed now): ObjectFactory = Container + Lookup. So overall this PR is useless if we get a CDILookup and CustomCDIContainer (which is way more easier to manage than a SPI whereever it is) or just a PassthroughContainer assuming the container is started with another extension before cucumber runs (which is also generally easy to do with junit like runners).
With that element I totally forgot when doing this PR, I tend to think we should package better CDI to enable this split and get rid of the customizers, wdyt?

mpkorstanje:
@iliadelo if you have a feature request you may want to start a new issue.

mpkorstanje:
I think this got stuck. So I've written #2439. Let's see if any one wants to contribute a solution like that.

PR #1876, opened 2020-01-26
[Java] Add BeforeAll and AfterAll hooks

mpkorstanje:
## Summary

`BeforeAll` and `AfterAll` hooks are executed before all scenarios are executed and
after all scenarios have been executed. A hook is declared by annotating a method.
This methods must be static and do not take any arguments.

Hooks are global, all hooks declared in any step definition class will be 
executed. The order in which hooks are executed is not defined. An explicit
order can be provided by using the `order` property in the annotation.

```java
package io.cucumber.example;

import io.cucumber.java.AfterAll;
import io.cucumber.java.BeforeAll;

public class StepDefinitions {

    @BeforeAll
    public static void beforeAll() {
        // Runs before all scenarios
    }

    @AfterAll
    public static void afterAll() {
        // Runs after all scenarios
    }
}
```

Notes:

 1. When used in combination with Junit 5, Maven Surefire, and/or Failsafe use 
    version `3.0.0-M5` or later.
 2. When used in combination with Junit 5 and InteliJ IDEA failures in before
    all and after all hooks do not fail a test run.

Fixes: #515.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue).
- [X] New feature (non-breaking change which adds functionality).
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected).

## Checklist:

- [X] I've added tests for my code.
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

Issue #1834, opened 2019-12-06
Improved Plugin API


mpkorstanje:
**Issue with the plugin system**

1. Concrete plugin implementations should not be in core. Plugins tend to use third party dependencies. We have to shade these in or suffer the dependency. Doing this in core make the core heavier then it needs to be.
2. We can't list all available plugins (for CLI options).
3. Third party plugins must be referenced by FQN while our own plugins have short names.
4. Input validation should happen when parsing. Currently happens at plugin instantiation.
5. `URLOutputStream` is an abomination.
6. No support for `Path` and `FileSystem` APIs.
7. The use of standard out is implicit in the constructor arguments of the plugin.
8. `Appendable` can't be closed but we trick around that in `NiceAppendable`.
9. `OutputStream` argument must always be closed. Plugins however can't tell if Cucumber crashed and thus can't clean it up. This should be done by Cucumber (it created the stream so it owns it).

**Solution**

1. Register and create plugins with SPI.
2. Provide interfaces to create an instance of a plugin from specific arguments
3. Remove `Appendable` as valid argument.

Something in that direction anyway.

naveenchandra91:
Is there also a plan to add an api which provides a minimal representation of some pickles(For some detailed third party reporting it needs Feature pickle etc) as a part of this feature request?

mpkorstanje:
Currently the pickles are available as `TestCase` objects in the `TestCaseStarted/Finished` events. Each event has a collection of `TestStep` objects that contain the information about either steps or hooks.

There currently is no good way to access how the pickles are organized in a feature and I hope to unlock that with https://github.com/cucumber/cucumber-jvm/issues/1870. But that will be a very generic tree structure. It won't for example contain the outline steps in a Scenario outline.

So it depends on how much detail you'll need. For AST level detail you'll have to parse the features yourself or wait for [cucumber messages](https://github.com/cucumber/cucumber/tree/master/messages ) to be integrated with Cucumber JVM which will include a well defined json schema for the AST.

mmuzikar:
I am working on a project that'd benefit from a more flexible plugin structure, can we make a deal where I'll rework the plugins and you let me add some features (such as reading registered stepdefs from glue, accessing type transformers etc.)

I started working on separating the plugins to their own small modules and not surprisingly this will break compatibility so it might be a good opportunity to introduce some wanted changes.

I am proposing these changes:
[...]

Issue #1711, opened 2019-07-29
JavaBackend should provide getter for SnippetGenerator and TypeRegistry


laeubi:
This is related to #368

The JavaBackend currently has no way to access SnippetGenerator and TypeRegistry that makes it impossible for IDEs or other code to reuse the information collected by the code.

mpkorstanje:
What kind of information are you looking for? Directly accessing the `Backend` might not be the best way to acquire it.

laeubi:
In general I'd like to reuse the parsing glue as well as generation of code snippets.
Currently we parse that data by our self for the cucumber-eclipse plugin, the problem with that approach is, that this might produce inconsistencies. So it seems natural to reuse the JavaBackend for this as all the logic about finding glue-code is already there and we don't want to catch up with changes done here.

To summarize: Using the eclipse plugin should produce the same errors/output as running against cucumber-cli.

mpkorstanje:
> In general I'd like to reuse the parsing glue as well as generation of code snippets.

In general that is going to be a problem because not all glue can be parsed without executing the code. Lamda step definitions for instance require the class that defines them is instantiated. This can be a problem when building the test context is a slow process.

But lets disregard Lambda step definitions for now.

>  it seems natural to reuse the JavaBackend for this as all the logic about finding glue-code is already there and we don't want to catch up with changes done here.

Specifically which bits of information are you looking for? And in what format would you like to receive them? I'd rather not open up all objects and let you look inside them but we could come up with a JSON format that makes it easier.

laeubi:
I'm sure your using some kind of IDE for development (maybe Idea or Eclipse), so simply think about what would you expect from an IDE to help you write Java code. Of course it is nice to have syntax highlight, but even more helpful are features that provide you with quick-fix and suggestions for autocomplete and you want to get error/warning hints if something is probably wrong.

Especially you want to support the "Specification Guys" so the don't need to *know* what steps are there but can find a list on there hand, automatically generated from projects code.

Lets assume I have a feature file and want to add rich edit capabilities for the user. 
So the user want's to know:
- If I run this feature with cucumber: will it produce any error (e.g. missing step defs, wrong parameter types)
- What steps are available and what parameter does it require
- Get autocompletion proposals when typing, jumping from one parameter to the next
- If a step is missing, generate me the snippet and insert it into a GlueCode file so it can be implemented

[...]

PR #2258, opened 2021-03-08
Custom predicate

raulmiraoliver:

https://community.smartbear.com/t5/Cucumber-Open/Custom-filters/td-p/213203
My team needs a way to have scenarios filtered in a more customizable way than static tagging. This allows setting new Pickle filters.

Created a new config called customPredicate that can be set to a Predicate<Pickle> to filter out scenarios.


raulmiraoliver:
Getting
```
[ERROR] Failed to execute goal org.revapi:revapi-maven-plugin:0.13.2:check (check) on project cucumber-junit: The following API problems caused the build to fail:
[ERROR] java.class.externalClassExposedInAPI: interface io.cucumber.core.gherkin.Argument: A class from supplementary archives is used in a public capacity in the API. [io.cucumber:cucumber-gherkin:jar:6.10.1-SNAPSHOT]
[ERROR] java.class.externalClassExposedInAPI: interface io.cucumber.core.gherkin.Pickle: A class from supplementary archives is used in a public capacity in the API. [io.cucumber:cucumber-gherkin:jar:6.10.1-SNAPSHOT]
[ERROR] java.class.externalClassExposedInAPI: interface io.cucumber.core.gherkin.Step: A class from supplementary archives is used in a public capacity in the API. [io.cucumber:cucumber-gherkin:jar:6.10.1-SNAPSHOT]
[ERROR] java.class.externalClassExposedInAPI: enum io.cucumber.core.gherkin.StepType: A class from supplementary archives is used in a public capacity in the API. [io.cucumber:cucumber-gherkin:jar:6.10.1-SNAPSHOT]
[ERROR] java.class.externalClassExposedInAPI: class io.cucumber.plugin.event.Location: A class from supplementary archives is used in a public capacity in the API. [io.cucumber:cucumber-plugin:jar:6.10.1-SNAPSHOT]
```

Please let me know if the PR has any future so I invest time on creating classes for the API to detach it from the supplementary archives.

mpkorstanje:
There several ways to achieve this using the frameworks Cucumber integrates with:

* If you are using TestNG I would suggest filtering the scenario in the `@DataProvider` annotated method.

 * If you are using JUnit 5 I would suggest using the JUnit platform in combination with the [Cucumber JUnit Platform Engine](https://github.com/cucumber/cucumber-jvm/tree/main/junit-platform-engine). That way the [JUnit Platform Launcher ](https://junit.org/junit5/docs/current/api/org.junit.platform.launcher/module-summary.html) will allow you to use a [`PostDiscoveryFilter`](https://junit.org/junit5/docs/current/api/org.junit.platform.launcher/org/junit/platform/launcher/PostDiscoveryFilter.html). This should allow you to dynamically filter on tags and names. The `PostDiscoveryFilter` can be discovered via SPI so should be enabled by default if used that way.

    Note that JUnit Platform is not a feature complete replacement for JUnit 4 yet. If you using multiple runners (suites) you'll be waiting for the release of https://github.com/junit-team/junit5/pull/2416 and if you need before/after all hooks you'll be waiting for https://github.com/cucumber/cucumber-jvm/pull/1876.

So 'm not sure if it is worth the additional complexity of integrating this with the core.

raulmiraoliver:
@mpkorstanje, I understand - thank you for your comment! I'll discuss with the team whether it's worth in our case to upgrade to JUnit 5, as that seems to be the most fit option.

Even if we do so, if you see any value in this PR I'm willing to invest some time on it. If not, please just say/feel free to close it.

Thanks!

mpkorstanje:
The offer is appreciated. Though as it stands the use case is somewhat of a niche.

Additionally loading classes by name through the CLI is something I'd like to replace with SPI (https://github.com/cucumber/cucumber-jvm/issues/1834) in the long term and possibly through a public API for running Cucumber (https://github.com/cucumber/cucumber-jvm/issues/1711).

PR #2281, opened 2021-04-10
[Core] Extract throwable collector

mpkorstanje:
Try and clean up the logic around throwing/re-throwing exceptions at various
abstraction levels a bit


core/src/main/java/io/cucumber/core/runner/TestAbortedExceptions.java
```diff
@@ -0,0 +1,27 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+final class TestAbortedExceptions {
+    private static final String[] TEST_ABORTED_EXCEPTIONS = {
+            "org.junit.AssumptionViolatedException",
+            "org.junit.internal.AssumptionViolatedException",
+            "org.opentest4j.TestAbortedException",
+            "org.testng.SkipException",
+    };
```
mpkorstanje, line 13: This needs a better solution that allows `instance of` to be used.

PR #2332, opened 2021-07-06
Provides an ObjectFactorySupplier to Runtime

sidharta:
This PR tries to solve two issues:
- There's no way to provide your own `ObjectFactory` which was possible in previous versions (4.x and below)
- If you provide a BackendSupplier, that object is not necessarily related to the `ObjectFactory` that is created on `Runtime`

Following the steps of `BackendSupplier` and `FeatureSupplier`, I added a new method to the builder to receive the ObjectFactorySupplier and I'm checking if that is provider it's not necessary to create a new one on `.build()`.

### DIFF
--- a/core/src/main/java/io/cucumber/core/runtime/Runtime.java
+++ b/core/src/main/java/io/cucumber/core/runtime/Runtime.java
@@ -121,6 +121,7 @@ public final class Runtime {
         private RuntimeOptions runtimeOptions = RuntimeOptions.defaultOptions();
         private BackendSupplier backendSupplier;
         private FeatureSupplier featureSupplier;
+        private ObjectFactorySupplier objectFactorySupplier;
         private List<Plugin> additionalPlugins = emptyList();
 
         private Builder() {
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/cucumber/core/runtime/Runtime.java
```diff
@@ -157,12 +163,18 @@ public Builder withEventBus(final EventBus eventBus) {
...
+            } else {
+                final ObjectFactoryServiceLoader objectFactoryServiceLoader = new ObjectFactoryServiceLoader(classLoader,
+                    runtimeOptions);
 
-            final ObjectFactorySupplier objectFactorySupplier = runtimeOptions.isMultiThreaded()
-                    ? new ThreadLocalObjectFactorySupplier(objectFactoryServiceLoader)
-                    : new SingletonObjectFactorySupplier(objectFactoryServiceLoader);
+                objectFactorySupplier = runtimeOptions.isMultiThreaded()
+                        ? new ThreadLocalObjectFactorySupplier(objectFactoryServiceLoader)
+                        : new SingletonObjectFactorySupplier(objectFactoryServiceLoader);
```
mpkorstanje, line 176: By replacing the  `ObjectFactorySupplier` the `ThreadLocalObjectFactorySupplier` and `SingletonObjectFactorySupplier` are no longer correctly applied. Considering making the `ObjectFactoryServiceLoader` implement the `ObjectFactorySupplier` and so that `ThreadLocalObjectFactorySupplier` and `SingletonObjectFactorySupplier` can be implemented as proper decorators.

mpkorstanje (changes requested):
Other then the practical problems that you are solving, what are you trying achieve by programatically adding an `ObjectFactorySupplier`?

Issue #2343, opened 2021-07-17
Support multiple DocStringTypes with the same contentType


PostelnicuGeorge:
As far as documentation goes, the fact that jackson json library is used and DataTableType can have multiple return types, it's not that obvious why DocStringType has only 1 return type for all feature files/project?

```feature
Given the following example:
  Scenario: Test whether DocStringType allows conversion of 2 different types of records
    Given I have the following cars
    """json
    [
    {
    "name": "Murcielago",
    "brand": "Lamborghini"
    },
    {
    "name": "Sandero",
    "brand": "Dacia"
    }
    ]
    """
    And one animal I like
    """json
    {
    "name": "Lion",
    "domesticated": false
    }
    """
```
```java
public class DemoStepDef {

	@DocStringType(contentType = "json")
	public List<Car> convertCars(String json) throws JsonProcessingException {
		return new ObjectMapper().readValue(json,
				new TypeReference<List<Car>>() {});
	}

	@DocStringType(contentType = "json")
	public Animal convertAnimal(String json) throws JsonProcessingException {
		return new ObjectMapper().readValue(json,
				new TypeReference<Animal>() {});
	}

	@Given("I have the following cars")
	public void i_have_cars(List<Car> cars) {
		System.out.println(cars);
	}

	@And("one animal I like")
	public void one_animal_I_like(Animal animal) {
		System.out.println(animal);
	}
}

public class Animal {
	private String name;
	private boolean domesticated;
// setters and getters omitted 
}

public class Car {
	private String name;
	private String brand;
// setters and getters omitted
}
```

Gives the following error:
```
**Caused by: io.cucumber.docstring.CucumberDocStringException: There is already docstring type registered for 'json' and com.lumeon.cpm.models.records.Animal.**
```
My feeling is that if jackson library is allowed for DocStringType annotation, then we should be able to leverage the full potential of it, and not be forced to have JsonNode as the only option. Having extra intermediary methods to convert to the actual types needed for the step definitions.

Work-around at the moment:
```java
public class DemoThatWorksStepDef {

	@DocStringType(contentType = "json")
	public JsonNode json(String json) throws JsonProcessingException {
		return new ObjectMapper().readTree(json);
	}

	public List<Car> convertCars(JsonNode jsonNode) throws IOException {
		return new ObjectMapper().readValue(jsonNode.traverse(),
				new TypeReference<List<Car>>() {});
	}

[...]

PR #2344, opened 2021-07-17
Fixes cucumber/cucumber-jvm#2343

PostelnicuGeorge:
Support multiple DocStringTypes with the same contentType #2343


docstring/src/main/java/io/cucumber/docstring/DocStringTypeRegistry.java
```diff
@@ -5,31 +5,37 @@
...
+        Map<Type, DocStringType> map;
+        if (docStringTypes.containsKey(docStringType.getContentType())) {
+            map = docStringTypes.get(docStringType.getContentType());
+        } else {
+            map = new HashMap<>();
         }
-        docStringTypesByContentType.put(docStringType.getContentType(), docStringType);
-        docStringTypesByType.put(docStringType.getType(), docStringType);
+        map.put(docStringType.getType(), docStringType);
+        docStringTypes.put(docStringType.getContentType(), map);
```
mpkorstanje, line 38: Consider using `Map.computeIfAbsent` to simplify section.

https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#computeIfAbsent-K-java.util.function.Function-

mpkorstanje (commented):
I don't know if you are done or not. If you're still working on this and would rather wait for feedback until you are done, you can mark the PR as a draft using the drop down on the green button. 

For now I've added a few suggestions. Most of them are related to code style, but don't worry about that too much until the functionality and tests are done.

PR #2354, opened 2021-08-05
[TestNG] Automatically pick up properties from testng.xml

mpkorstanje:
When using TestNG suites can be declared in `testng.xml`. These suites can be
configured with parameters. When using `AbstractTestNGCucumberTests` these
will be picked up automatically.

When using the `TestNGCucumberRunner` they can be read through `ITestContext`.

```
    @BeforeClass(alwaysRun = true)
    public void setUpClass(ITestContext context) {
        XmlTest currentXmlTest = context.getCurrentXmlTest();
        CucumberPropertiesProvider properties = currentXmlTest::getParameter;
        testNGCucumberRunner = new TestNGCucumberRunner(this.getClass(), properties);
    }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
      - Replace `DataTable.asList()` with -> `DataTable.values()`
      - Replace `DataTable.asLists()` with -> `DataTable.cells()`
      - Replace `DataTable.asMaps()` with -> `DataTable.entries()`
+ * [TestNG] Automatically pick up properties from `testng.xml` ([#2354](https://github.com/cucumber/cucumber-jvm/pull/2354) M.P. Korstanje, Gayan Sandaruwan)
 
 ### Deprecated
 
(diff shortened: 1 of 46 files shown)


kalaikovan-QA:
@mpkorstanje Hi,
In this file there is @test method which converts every scenario into test.
```
@Test(groups = "cucumber", description = "Runs Cucumber Scenarios", dataProvider = "scenarios")
	public void scenario(PickleWrapper pickle, FeatureWrapper cucumberFeature) {
		getRunner().runScenario(pickle.getPickle());
	}
```
am using ITestListener ,so the result.getName() gives as 'scenario' instead of actual scenario name.
when i debugged `pickle.getPickle().getName()` returns the actual scenario name.

Is it possible to send the scenario name as well to the `runScenario` method

mpkorstanje:
I dunno, but if you have a feature request it would be good to 1) create a separate issue, and 2) explain how such a thing would be done normally with TestNG.

kalaikovan-QA:
@mpkorstanje Thanks for the quick response.
#2487

Issue #2044, opened 2020-07-07
Add Bill of Materials (BOM)


mpkorstanje:
A typical problem I see on Stack Overflow is that people use a mismatched set of Cucumber dependencies. This could be avoided by introducing a [Bill of Material (BOM)](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html) artefact and changing the documentation to reference this BOM.

I would like to see an implementation this BOM. Preferably one that is either automatically generated or used by the project itself to ensure it stays up to date without manual maintenance. A PR that adds a BOM should also update the documentation in Cucumber-JVM and change the `cucumber-archetype` to use the BOM. Updating the documentation outside of of this project is not required to complete this merge request but listed for completeness:
  * https://github.com/cucumber/docs.cucumber.io
  * https://github.com/cucumber/cucumber-java-skeleton

An alternative would be to use `cucumber-jvm` as a BOM. However to avoid the issues from https://github.com/cucumber/cucumber-jvm/pull/1900 it should be cleaned from all non essential dependencies which I don't think is feasible.

There are a few styles to implement a BOM:

 * A root module as describe by https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html
 * A module listing the versions of all modules in the project https://search.maven.org/artifact/org.junit/junit-bom/5.7.0-M1/pom
 * A module with third party dependencies https://search.maven.org/artifact/org.springframework.boot/spring-boot-dependencies/2.3.1.RELEASE/pom

If possible I'd like to avoid introducing another root module. So the JUnit 5 style seems to be most appropriate. It will avoid introducing non-cucumber dependencies. However unlike JUnit 5, not all `io.cucumber` dependencies are part of `cucumber-jvm`. This may require some additional work.

---

A valid conclusion may also be that it's not worth the effort to add a Bill of Materials. A typical project only needs 3 cucumber dependencies. All others are transitive. The typical problem on stack overflow includes everything and the kitchen sink as a dependency.

```xml
        <dependency>
            <groupId>io.cucumber</groupId>
            <artifactId>cucumber-spring</artifactId>
            <scope>test</scope>
        </dependency>
        <dependency>
            <groupId>io.cucumber</groupId>
            <artifactId>cucumber-java</artifactId>
            <scope>test</scope>
        </dependency>
        <dependency>
            <groupId>io.cucumber</groupId>
            <artifactId>cucumber-junit-platform-engine</artifactId>
[...]

PR #2363, opened 2021-08-17
[BOM] Add Bill of Materials

mpkorstanje:
## Summary

It is fairly common for one project to use more than one cucumber dependency.
To keep these version in sync a Bill of Materials can be used.

## Usage

```xml
<dependencyManagement>
    <dependencies>
        <dependency>
            <artifactId>cucumber-bom</artifactId>
            <type>pom</type>
            <scope>import</scope>
        </dependency>
    </dependencies>
</dependencyManagement>

<!--Then for example-->

<dependencies>
    <dependency>
        <scope>test</scope>
    </dependency>
    <dependency>
        <scope>test</scope>
    </dependency>
</dependencies>
```

## Details

When using `<scope>import</scope>` all dependencies from the parent pom are
imported. This unfortunately include transitive dependencies from the parent
pom. As such all dependencies from `cucumber-jvm` had to be pushed down to
individual modules.

While this does result in some duplication, it also allows dependencies in
modules to be upgraded independently. This is an acceptable trade off as the
modules already have relatively flew overlapping dependencies. Additionally
with Renovate Bot, the tedium of upgrading is mostly automated.

Closes: #2044


mpkorstanje:
Also waiting for JUnit 5.8

PR #2378, opened 2021-09-09
[DataTable] Merge datatable from cucumber/common

mpkorstanje:
With https://github.com/cucumber/common/issues/1724 in mind and no plans for an other implementation of the data table module we merge it into cucumber-jvm..


aslakhellesoy:
LGTM! Will you send another PR to remove it from `cucumber/common`?

PR #2399, opened 2021-10-07
[JUnit Platform] Use Kotlin dsl in Gradle scripts

jamietanna:
As per comment from M.P. Korstanje, we should align our examples for
Gradle documentation as just Kotlin, as the Gradle team are recommending
Kotlin going forward.

### DIFF
--- a/junit-platform-engine/README.md
+++ b/junit-platform-engine/README.md
@@ -57,10 +57,10 @@ parameter. This will include the feature name as part of the tests name.
 
 #### Gradle
 
-```groovy
-test {
+```kotlin
+tasks.test {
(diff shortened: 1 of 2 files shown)


jamietanna:
Ah, I'd not seen that before from them - I'll go through and check the other examples and update this PR 👍🏽

mpkorstanje:
Cheers. There should be quite a few in the docs.cucumber.io repo.  I'm updating the java-skeleton.

mpkorstanje:
No. Documentation is not saying it explicitly. But look at the JUnit and Gradle projects. They're using Kotlin, not Groovy.

Issue #2370, opened 2021-08-31
hooks should not require public visibility -- at least should log an error/warning


elonderin:
**Is your feature request related to a problem? Please describe.**
I was frustrated when i tried to figure out why my tests would fail.
Until i notices that my setup was not public. From the old days in junit i know that they used to require that too, but no more and nowadays it's so common to have all that test stuff with package visibility (to avoid cluttering autocompletion in IDEs i guess) I just did it here too w/o even giving it a 2nd though.

So now my code looks like so to make my team mates (we are all new CC) aware of the fact.
```java
   @Before //method must be public!! otherwise it's silently ignored!
    public void setUp() {
        msUniteTestHarness.reset();
    }
```

**Describe the solution you'd like**
Also allow package visibility but by all means don't just silently drop the execution

**Describe alternatives you've considered**
Throwing an error would be OK too but behavior is not anymore in alignment with junit.

mpkorstanje:
This is an acceptable suggestion, but:

1. This would be a breaking change. So it would have to go into the next major version. We are currently in the process of releasing v7 RC1 so this would have to be resolved in relatively short other. Otherwise the next opportunity would be v8. Possibly next year but impossible to say as this is an open source project.

2.  [The motivation behind JUnit 5 not requiring methods to be public was to reduce the amount of ceremony needed to write tests](https://stackoverflow.com/a/55230350/3945473). Specifically test classes and methods are can be `package-private`, `projected` or `public`.
Merely allowing methods to not be public but classes would only lead to more confusion. So any pull requests that address this issue would have to be equally motivated to reduce ceremony and address both method and class visibility.

    Fortunately both are covered by the same section of code:

    https://github.com/cucumber/cucumber-jvm/blob/6a4567a7a32d904debbfda2d77feab0a9569e547/java/src/main/java/io/cucumber/java/MethodScanner.java#L23-L45

Would you be interested in creating a pull request?

elonderin:
I wish i had the time then i would. Not seeing this though...

> This would be a breaking change.

May i ask why ? After all, the methods which are now public would still work or is there some impl. detail i dont understand?

mpkorstanje:
Step definitions and hooks that would previously be ignored would now be discovered. Even though this may not have been intentional, it would change an existing setup.

elonderin:
is there a timeline on v8 ? I guess it's still long way out/unknown, or ?

maybe have an opt-in property for v7?

[...]

PR #2415, opened 2021-11-07
[Java] Define step definitions and hooks with minimal ceremony

rozenshteyn:

[Java] Allow non-private classes to contain hooks and allow hook annotations in non-public methods (#2370)

1. When scanning methods of a class, allow the class to have package-private and protected visibility in addition to the previously allowed public visibility.
2. When scanning methods of a class, allow the hook-annotated methods to have package-private and protected visibility in addition to the previously allowed public visibility.

**Additional context**
None.


mpkorstanje:
Wow. I'll have to find some time to review this. It is likely that I won't have any in the next two weeks.

java/src/test/java/io/cucumber/java/MethodScannerTest.java
```diff
@@ -29,10 +29,45 @@ void createBackend() {
...
-    void scan_finds_annotated_methods() throws NoSuchMethodException {
-        Method method = BaseSteps.class.getMethod("m");
+    void scan_finds_annotated_methods_in_public_class() throws NoSuchMethodException {
+        Method publicMethod = BaseSteps.class.getMethod("m");
+        Method packagePrivateMethod = BaseSteps.class.getDeclaredMethod("n");
+        Method protectedMethod = BaseSteps.class.getDeclaredMethod("o");
         MethodScanner.scan(BaseSteps.class, backend);
-        assertThat(scanResult, contains(new SimpleEntry<>(method, method.getAnnotations()[0])));
+        assertThat(scanResult,
+            contains(new SimpleEntry<>(publicMethod, publicMethod.getAnnotations()[0]),
```
mpkorstanje, line 38: Consider using `org.hamcrest.collection.IsMapContaining.hasEntry`.

mpkorstanje, line 38: Consider using `org.hamcrest.collection.IsMapContaining.hasEntry` for all these.

Would have been better if I had done that the first time round, but better late then never.

mpkorstanje (approved):
Nitpick aside. I don't see any immediate problems with this.

I think this can be merged with the next major version.

However prior to merging we should update the code in the `examples` module, the `java` module documentation and the `archetype` so it is clear that Cucumber with minimal ceremony is preferred. 

Given that it will be some time before the next major version rolls around, and given that those changes mentioned above are fragile I think it would be prudent to delay them until merging.

After merging the https://github.com/cucumber/cucumber-java-skeleton should also be updated.

mpkorstanje:
Note to self: Drop [0158029](https://github.com/cucumber/cucumber-jvm/pull/2415/commits/0158029962852af48c72f03db20516ebb03029dd) before merging. Looks like it was an attempted fix for a different issue.

mpkorstanje:
Cherry picked into #3200. It's a bit hard to get this PR up to date after 4 years. :sweat_smile:

PR #2421, opened 2021-11-21
Support multiple DocStringTypes with the same contentType (#2343)

PostelnicuGeorge:
It's the implementation to: https://github.com/cucumber/cucumber-jvm/issues/2343

With this implementation is now possible to have DocStrings of default type ("") or a particular contentType like "application/json" to have multiple DocStringTypes (different return types). This gives leverage for users to take full advantage of Jackson ObjectMapper conversions.
Tests were implemented and updated to cover the new situations where it should be allowed to have multiple DocStringTypes per same contentType, this creates much more concise transformers but doesn't allow any more ambiguity around Object as a return type.

### DIFF
--- a/core/src/test/java/io/cucumber/core/plugin/JsonFormatterTest.java
+++ b/core/src/test/java/io/cucumber/core/plugin/JsonFormatterTest.java
@@ -12,6 +12,7 @@ import io.cucumber.core.runtime.StubBackendSupplier;
 import io.cucumber.core.runtime.StubFeatureSupplier;
 import io.cucumber.core.runtime.TimeServiceEventBus;
 import io.cucumber.datatable.DataTable;
+import io.cucumber.docstring.DocString;
 import org.junit.jupiter.api.Test;
 
 import java.io.ByteArrayOutputStream;
@@ -1262,7 +1263,7 @@ class JsonFormatterTest {
                 .withAdditionalPlugins(timeService, new JsonFormatter(out))
                 .withEventBus(new TimeServiceEventBus(timeService, UUID::randomUUID))
                 .withBackendSupplier(new StubBackendSupplier(
-                    new StubStepDefinition("there are bananas", "StepDefs.there_are_bananas()", String.class)))
+                    new StubStepDefinition("there are bananas", "StepDefs.there_are_bananas()", DocString.class)))
                 .build()
                 .run();
 
--- a/core/src/test/java/io/cucumber/core/plugin/JsonFormatterTest.java
+++ b/core/src/test/java/io/cucumber/core/plugin/JsonFormatterTest.java
@@ -1252,7 +1252,7 @@ class JsonFormatterTest {
                 "\n" +
                 "  Scenario: Monkey eats bananas\n" +
                 "    Given there are bananas\n" +
-                "    \"\"\"doc\n" +
+                "    \"\"\"text/plain\n" +
                 "    doc string content\n" +
                 "    \"\"\"\n");
 
@@ -1291,7 +1291,7 @@ class JsonFormatterTest {
                 "            \"name\": \"there are bananas\",\n" +
                 "            \"line\": 4,\n" +
                 "            \"doc_string\": {\n" +
-                "              \"content_type\": \"doc\",\n" +
+                "              \"content_type\": \"text/plain\",\n" +
                 "              \"value\": \"doc string content\",\n" +
                 "              \"line\": 5\n" +
                 "            },\n" +
--- a/core/src/test/java/io/cucumber/core/stepexpression/StepExpressionFactoryTest.java
+++ b/core/src/test/java/io/cucumber/core/stepexpression/StepExpressionFactoryTest.java
@@ -167,6 +167,17 @@ class StepExpressionFactoryTest {
     }
 
     @Test
+    void docstring_and_datatable_match_same_step_definition() {
+        String docString = "A rather long and boring string of documentation";
+        StepDefinition stepDefinition = new StubStepDefinition("Given some stuff:", UNKNOWN_TYPE);
+        StepExpression expression = stepExpressionFactory.createExpression(stepDefinition);
+        List<Argument> match = expression.match("Given some stuff:", docString, null);
+        assertThat(match.get(0).getValue(), is(equalTo(DocString.create(docString))));
+        match = expression.match("Given some stuff:", table);
+        assertThat(match.get(0).getValue(), is(equalTo(DataTable.create(table))));
+    }
+
+    @Test
     void docstring_expression_transform_doc_string_to_json_node() {
         String docString = "{\"hello\": \"world\"}";
         String contentType = "json";
--- a/core/src/test/java/io/cucumber/core/stepexpression/StepExpressionFactoryTest.java
+++ b/core/src/test/java/io/cucumber/core/stepexpression/StepExpressionFactoryTest.java
@@ -158,12 +158,11 @@ class StepExpressionFactoryTest {
     }
 
     @Test
-    void docstring_expression_transform_doc_string_with_content_type_to_string() {
+    void docstring_expression_transform_doc_string_to_string() {
         String docString = "A rather long and boring string of documentation";
-        String contentType = "doc";
         StepDefinition stepDefinition = new StubStepDefinition("Given some stuff:", String.class);
         StepExpression expression = stepExpressionFactory.createExpression(stepDefinition);
-        List<Argument> match = expression.match("Given some stuff:", docString, contentType);
+        List<Argument> match = expression.match("Given some stuff:", docString, null);
         assertThat(match.get(0).getValue(), is(equalTo(docString)));
     }
 
--- a/docstring/src/main/java/io/cucumber/docstring/DocStringType.java
+++ b/docstring/src/main/java/io/cucumber/docstring/DocStringType.java
@@ -37,11 +37,11 @@ public final class DocStringType {
         this.transformer = requireNonNull(transformer);
     }
 
-    public String getContentType() {
+    String getContentType() {
         return contentType;
     }
 
-    public Type getType() {
+    Type getType() {
         return type;
     }
 
--- a/docstring/src/main/java/io/cucumber/docstring/DocStringType.java
+++ b/docstring/src/main/java/io/cucumber/docstring/DocStringType.java
@@ -37,11 +37,11 @@ public final class DocStringType {
         this.transformer = requireNonNull(transformer);
     }
 
-    String getContentType() {
+    public String getContentType() {
         return contentType;
     }
 
-    Type getType() {
+    public Type getType() {
         return type;
     }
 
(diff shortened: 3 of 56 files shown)


mpkorstanje:
> But given that I saw that the java8 project is marked for deprecation I'm awaiting further details on how to deal with this one.

We're planning for deprecation but the project hasn't been deprecated yet - due to a lack of replacement. Until it is removed `cucumber-java8` should remain functioning in the same way as `cucumber-java`.

PostelnicuGeorge:
> > But given that I saw that the java8 project is marked for deprecation I'm awaiting further details on how to deal with this one.
> 
> We're planning for deprecation but the project hasn't been deprecated yet - due to a lack of replacement. Until it is removed `cucumber-java8` should remain functioning in the same way as `cucumber-java`.

Got it, will investigate that method and see if I manage to make it return the actual GenericReturnType of DocStringType lambdas.

PostelnicuGeorge:
Fixed the "type erasure"  issue, all test pass, I'm able to do a "clean install" at the project level.

java8/src/main/java/io/cucumber/java8/LambdaGlue.java
```diff
@@ -406,7 +406,7 @@ default void AfterStep(String tagExpression, int order, final HookNoArgsBody bod
      *                    <code>type</code> from the doc string
      * @see               io.cucumber.docstring.DocStringType
      */
-    default void DocStringType(String contentType, DocStringDefinitionBody<?> body) {
+    default <T> void DocStringType(String contentType, DocStringDefinitionBody<T> body) {
```
mpkorstanje, line 409: Neat!

java/src/test/java/io/cucumber/java/JavaDocStringTypeDefinitionTest.java
```diff
@@ -28,16 +96,39 @@
     private final DocStringTypeRegistryDocStringConverter converter = new DocStringTypeRegistryDocStringConverter(
         registry);
 
+    @Test
+    void doc_string_type_gets_correctly_init() throws NoSuchMethodException {
```
mpkorstanje, line 100: This test is testing the implementation details of `DocString`. This requires making more `DocStrings` methods public then needed. Instead consider testing if the correct conversion method is invoked for simple and complex types (e.g. `List<Sting>` and `List`.

docstring/src/test/java/io/cucumber/docstring/DocStringTest.java
```diff
@@ -5,20 +5,9 @@
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.junit.jupiter.api.Assertions.assertThrows;
 
 class DocStringTest {
 
-    @Test
-    void throws_when_no_converter_defined() {
```
mpkorstanje, line 13: This test still seems correct. Why was it removed?

docstring/src/test/java/io/cucumber/docstring/DocStringTypeRegistryDocStringConverterTest.java
```diff
@@ -16,49 +19,47 @@
...
     }
 
     @Test
-    void uses_target_type_when_available() {
-        registry.defineDocStringType(new DocStringType(
-            JsonNode.class,
-            "json",
-            (String s) -> new ObjectMapper().readTree(s)));
-
+    void converts_doc_string_to_doc_string() {
```
mpkorstanje, line 31: It looks like some methods got shuffled around in this file. This makes reviewing rather difficult. Can you put them back in the same order as much as possible?

docstring/src/test/java/io/cucumber/docstring/DocStringTypeRegistryDocStringConverterTest.java
```diff
@@ -16,49 +19,47 @@
...
+
+        JsonNode converted = converter.convert(docString, JsonNode.class);
+        assertThat(converted.asText(), equalTo(docString.getContent()));
     }
 
     @Test
-    void converts_doc_string_to_doc_string() {
-        DocString docString = DocString.create(
-            "{\"hello\":\"world\"}");
+    void throws_when_no_converter_defined() {
```
mpkorstanje, line 54: This test does not involve the converter. I think it should move back to `DocStringTest`.

java8/src/test/java/io/cucumber/java8/TypeDefinitionsStepDefinitions.java
```diff
@@ -18,8 +19,8 @@
 
     public TypeDefinitionsStepDefinitions() {
         Given("docstring, defined by lambda",
-            (StringBuilder builder) -> assertThat(builder.getClass(), equalTo(StringBuilder.class)));
-        DocStringType("doc", (String docString) -> new StringBuilder(docString));
+            (DocString builder) -> assertThat(builder.getClass(), equalTo(DocString.class)));
+        DocStringType("doc", DocString::create);
```
mpkorstanje, line 23: I don't understand why this was changed. Can you explain?

docstring/src/main/java/io/cucumber/docstring/DocStringType.java
```diff
@@ -37,11 +37,11 @@
         this.transformer = requireNonNull(transformer);
     }
 
-    String getContentType() {
+    public String getContentType() {
         return contentType;
     }
 
-    Type getType() {
+    public Type getType() {
```
mpkorstanje, line 44:
```suggestion
    Type getType() {
```

docstring/src/main/java/io/cucumber/docstring/DocStringType.java
```diff
@@ -37,11 +37,11 @@
         this.transformer = requireNonNull(transformer);
     }
 
-    String getContentType() {
+    public String getContentType() {
```
mpkorstanje, line 40:
```suggestion
    String getContentType() {
```

core/src/test/java/io/cucumber/core/stepexpression/StepExpressionFactoryTest.java
```diff
@@ -158,12 +158,11 @@ void unknown_target_type_transform_doc_string_to_doc_string() {
     }
 
     @Test
-    void docstring_expression_transform_doc_string_with_content_type_to_string() {
+    void docstring_expression_transform_doc_string_to_string() {
```
mpkorstanje, line 161: This test is now identical to the test above `unknown_target_type_transform_doc_string_to_doc_string`. Why did you change this?

core/src/test/java/io/cucumber/core/plugin/JsonFormatterTest.java
```diff
@@ -1262,7 +1263,7 @@ void should_format_scenario_with_a_step_with_a_doc_string_and_content_type() {
                 .withAdditionalPlugins(timeService, new JsonFormatter(out))
                 .withEventBus(new TimeServiceEventBus(timeService, UUID::randomUUID))
                 .withBackendSupplier(new StubBackendSupplier(
-                    new StubStepDefinition("there are bananas", "StepDefs.there_are_bananas()", String.class)))
+                    new StubStepDefinition("there are bananas", "StepDefs.there_are_bananas()", DocString.class)))
```
mpkorstanje, line 1266: Why was this changed?

mpkorstanje (changes requested):
Heya, looks really good!

I've pushed some changes to resolve some of nitpicks I had with the `DocStringTypeRegistry` and `DocStringTypeRegistryDocStringConverter`. By making `docStringTypeRegistry.lookup` return a list it is possible to handle all error cases at once.

I've got some questions about the other changes.

mpkorstanje:
Btw, don't worry about force pushing. The default is squash merge so we don't have to keep branches clean.

core/src/test/java/io/cucumber/core/stepexpression/StepExpressionFactoryTest.java
PostelnicuGeorge: I can start by saying we can do without it, but they're not quite identical, they're covering both situations were you either want to convert the DocString that doesn't have a contentType to either a String (the default converter) or to an Object which gets converted to a DocString in io/cucumber/core/stepexpression/StepExpressionFactory.java:72

docstring/src/test/java/io/cucumber/docstring/DocStringTest.java
PostelnicuGeorge: I thought it's better suited in DocStringTypeRegistryDocStringConverterTest, but you're right it doesn't belong there, I will revert this.

docstring/src/test/java/io/cucumber/docstring/DocStringTypeRegistryDocStringConverterTest.java
PostelnicuGeorge: Re-arranged them, to be in the same order as originally, though the first test from before has to be changed to throw an exception given that the functionality no longer works the same.

docstring/src/test/java/io/cucumber/docstring/DocStringTypeRegistryDocStringConverterTest.java
PostelnicuGeorge: You are right, moving it back to it's original place.

java8/src/test/java/io/cucumber/java8/TypeDefinitionsStepDefinitions.java
PostelnicuGeorge: This was me trying to make it work before making io/cucumber/java8/LambdaGlue.java:409 generic:
 default <T> void DocStringType(String contentType, DocStringDefinitionBody<T> body) {} 
 I reverted this

PostelnicuGeorge:
I have added a test to check that if we use the same parameter type like List<Grocery> in a step definition, we're able to use interchangeably that a DocString or DataTable can be used that convert to that particular type. Is there a better place to put this kind of test?
io.cucumber.core.stepexpression.StepExpressionFactoryTest#docstring_and_datatable_match_same_step_definition

core/src/test/java/io/cucumber/core/stepexpression/StepExpressionFactoryTest.java
mpkorstanje: Right. Makes sense.

docstring/src/test/java/io/cucumber/docstring/DocStringTypeRegistryDocStringConverterTest.java
mpkorstanje: Appreciated!

mpkorstanje:
I'd say that `StepExpressionFactoryTest` is the right place for that test. It's where both converters come together.

I'd say that all that is left now is to add an example to the documentation to illustrate this usecase. 

https://github.com/cucumber/cucumber-jvm/tree/main/java#docstring-type

I think one one example where two json doc strings are converted to different objects would be sufficient.

mpkorstanje:
I'll have to find the time to give it another review and then it will be merged and released.

All the issues tagged with `help wanted` should be one you could pick up.

Issue #2278, opened 2021-04-03
Automatically detect Guice InjectorSource


mpkorstanje:
Currently the [InjectorSource](https://github.com/cucumber/cucumber-jvm/blob/main/guice/src/main/java/io/cucumber/guice/InjectorSource.java) is created by looking up the `guice.injector-source` properties from the `cucumber.properties` file. 

Unfortunately this system is separated from the usual properties parsing. This means that for example when using JUnit5 which uses `junit-platform.properties` both `junit-platform.properties` and `cucumber.properties` have to be used.

As solution to this would be similar to https://github.com/cucumber/cucumber-jvm/pull/1911 where a `Backend` implementation is used to scan for classes that implement `io.cucumber.guic.InjectorSource`. The `GuiceFactory` can then look for `InjectorSource` implementations and when there is exactly one, use it as the injector source.

This would remove the need to use a property all together and allow the `InjectorSource` to be treated as glue.

PR #2432, opened 2021-12-04
[Guice] Automatically detect InjectorSource

PostelnicuGeorge:
Implementation of #2278


PostelnicuGeorge:
I have finished the tests and implementation, can I get advice on how to fix the java.class.externalClassExposedInAPI: interface io.cucumber.core.backend.Backend: A class from supplementary archives is used in a public capacity in the API. [io.cucumber:cucumber-core:jar:7.1.0-SNAPSHOT] errors from Test (Semver check) build step?

guice/src/main/java/io/cucumber/guice/GuiceBackendProviderService.java
```diff
@@ -0,0 +1,17 @@
+package io.cucumber.guice;
+
+import io.cucumber.core.backend.Backend;
+import io.cucumber.core.backend.BackendProviderService;
+import io.cucumber.core.backend.Container;
+import io.cucumber.core.backend.Lookup;
+
+import java.util.function.Supplier;
+
+public final class GuiceBackendProviderService implements BackendProviderService {
```
mpkorstanje, line 10: The `BackendProviderService` is load through [Java's SPI mechanism](https://docs.oracle.com/javase/tutorial/sound/SPI-intro.html).

https://github.com/cucumber/cucumber-jvm/blob/5598e3441409878d9de40be7df67d62e3c314602/core/src/main/java/io/cucumber/core/runtime/BackendServiceLoader.java#L31-L34

This requires a registration in `META-INF/services`, for example the `SpringBackendProviderService` is registered with:

https://github.com/cucumber/cucumber-jvm/blob/5598e3441409878d9de40be7df67d62e3c314602/spring/src/main/resources/META-INF/services/io.cucumber.core.backend.BackendProviderService#L1

I suspect this is also why the semantic version check is complaining.

mpkorstanje (changes requested):

guice/src/main/java/io/cucumber/guice/GuiceBackendProviderService.java
PostelnicuGeorge: I've added the same, and looked at all differences between cucumber-spring project and cucumber-guice and for the life of me I don't get why I'm getting externalClassExposedInAPI, my feeling is that it's reporting binary differences that actually exist and we now have for guice a GuiceBackendProviderService, which exposes Backed, Lookup and Container APIs and thus the errors are real, BUT still don't get why these are not happening for spring project where we have a SpringBackendProviderService?

PostelnicuGeorge: I've updated a couple of small things that I've missed before, updated to the latest main branch, but with this issue I'm blocked if you can point me on some documentation to read or some other examples to look at, I would be very happy to take another look.

mpkorstanje: Perhaps because those classes have already been exposed in the previous version.Though I'd think that implementing something with SPI should be enough to let revapi know this exposure is intentional.

Perhaps it could be considered a bug?

 * https://revapi.org/revapi-site/main/index.html
 * https://github.com/revapi/revapi/
 
 For the short term I would be perfectly happy with adding the suggested exclusions.
 
 When you run the semantic version check it will output a few json blobs:
 
 ```
 mvn verify -Pcheck-semantic-version -DskipTests=true -DskipITs=true -Darchetype.test.skip=true
 ```
 
 These can be added here:
 
 https://github.com/cucumber/cucumber-jvm/blob/main/.revapi/api-changes.json#L24
 
 There are only a few fields actually required. The others aren't too interesting.

PostelnicuGeorge: Hey sorry for not replying, I was busy finishing work for this year, I will open a bug for revapi project and give this PR as an example. Let me know if there is any more work available, I'm in "vacation" until mid January.

mpkorstanje:
Cheers again!

Issue #2285, opened 2021-04-19
Allow using a HTTP proxy to send results


aslakhellesoy:
We should allow users behind a HTTP(S) proxy to publish results. See https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html

aslakhellesoy:
We should make sure all implementations use the same environment variables to configure proxy settings.

* https://github.com/cucumber/cucumber-ruby/issues/1526
* https://github.com/cucumber/cucumber-js/issues/1644

mpkorstanje:
At this point it would be prudent to replace the `HttpURLConnection` with a proper shaded http client. When using the `http.*` properties will affect all components in `java.net` including those of the system under tests. So this may not be feasible. Especially in more complicated setups.

mpkorstanje:
Java 11 does come with a `HttpClient` that is an overall improvement over the `HttpURLConnection`, but as Cucumber is currently still on Java 8 we can not use this yet.  We do have a major version in the works that would allow us to increase the minimum version of the JRE on which Cucumber is ran.

https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpClient.html

aslakhellesoy:
Can't we just use a `java.net.Proxy` object and take the configuration from cucumber-specific environment variables? For example:

* `CUCUMBER_PUBLISH_HTTP_PROXY_HOST`
* `CUCUMBER_PUBLISH_HTTP_PROXY_PORT`

mpkorstanje:
Not on a `HttpURLConnection`, you could create a `URL` instance with a custom implementation of `URLStreamHandler` but that's all sorts painful.

aslakhellesoy:
> Not on a `HttpURLConnection`, you could create a `URL` instance with a custom implementation of `URLStreamHandler` but that's all sorts painful.

Why not? 

At https://github.com/cucumber/cucumber-jvm/blob/0a101a30464e9ba68432dd21d67e426aeb2d01bb/core/src/main/java/io/cucumber/core/plugin/UrlOutputStream.java#L72 we could pass in a proxy object. See https://docs.oracle.com/javase/7/docs/api/java/net/URL.html#openConnection%28java.net.Proxy%29

mpkorstanje:
Didn't even know you could do that.

mpkorstanje:
We can probably extend the curl option syntax with the curl option for proxy:

```
[...]

PR #2452, opened 2022-01-01
[Core] Support proxy for publish plugin

mpkorstanje:
Users looking to publish cucumber reports behind a proxy can add to
`cucumber.properties`:

```
cucumber.publish.proxy=https://proxy.example.com:3129
```

Note: The publish-plugin only reads properties from the `cucumber.properties`,
the environment, system properties. And not `junit-platform.properties`. 

Fixes: #2285

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   - When transforming doc strings in addition to the content type Cucumber will
     also look at the type used in the step definition to disambiguate between
     doc string types.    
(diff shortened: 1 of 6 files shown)


mpkorstanje:
@aslakhellesoy do you have a proxy to test with?

aslakhellesoy:
@mpkorstanje seems to work fine! I tried with one of the proxies listed at https://hidemy.name/en/proxy-list/?country=GB&type=s#list

I got an error when I didn't put an explicit port in the proxy URL. Should we make it default to `443` for `https` and `80` for `http`?

mpkorstanje:
Proxies don't typically have a default port.

PR #2497, opened 2022-03-08
[Core] Use null safe-messages

mpkorstanje:
With https://github.com/cucumber/common/pull/1879 `messages` now uses
`Optional<T>` for optional types avoiding some unexpected null references.
This resulted in some updating of `gherkin` and `html-formatter`.

Additionally with this change the dependency on Jackson has been removed from
`messages`. However we still have to serialize the message objects into json.
So for now Jackson is shaded into `core` instead. And eventually we'll be able
to spin this out to it's own module.


pom.xml
```diff
@@ -21,8 +21,8 @@
 
         <!-- io.cucumber dependencies -->
         <!-- Whenever messages is updated run `make update-cck` -->
-        <messages.version>17.1.1</messages.version>
-        <gherkin.version>22.0.0</gherkin.version>
+        <messages.version>17.1.2-SNAPSHOT</messages.version>
```
aslakhellesoy, line 24: We can change this to `18.0.0` now that it's released.

bom/pom.xml
```diff
@@ -15,7 +15,7 @@
         <ci-environment.version>9.0.4</ci-environment.version>
         <cucumber-expressions.version>15.0.2</cucumber-expressions.version>
         <datatable.version>4.1.0</datatable.version>
-        <html-formatter.version>18.0.0</html-formatter.version>
+        <html-formatter.version>19.0.1-SNAPSHOT</html-formatter.version>
```
mpkorstanje, line 18: See https://github.com/cucumber/html-formatter

aslakhellesoy:
Very nice!!

Issue #2536, opened 2022-04-28
Datatables output by the pretty formatter are displayed incorrectly by gradle


scottadav:
### 👓 What did you see?

Problem is:
- Data tables are displayed in console with excessive newlines

Defect occurs when:
- The tests contain a DataTable
- The tests are run using Gradle and the JUnit5 Platform Engine
- The pretty plugin is enabled
- Test logging is enabled

Here is an example of the console output from running a sample test with gradle:
```
$ ./gradlew test
Configuration cache is an incubating feature.
Calculating task graph as no configuration cache is available for tasks: test

> Task :test

RunAllCucumberTests > Cucumber > Demonstrate datatable display bug > com.example.RunAllCucumberTests.Print datatable STANDARD_OUT

    Scenario: Print datatable   # features/DataTable.feature:3
      Given a data table exists # com.example.steps.Steps.aDataTableExists(io.cucumber.datatable.DataTable)
    
    | 
    ColumnA
     | 
    ColumnB
     | 
    ColumnC
     |

    
    | 
    1
 
 
 
 
 
 
     | 
    2
 
 
 
 
 
 
     | 
    3
 
 
 
 
 
 
     |

    
    | 
    4
 
 
 
 
 
 
     | 
    5
 
 
 
 
 
 
     | 
    6
 
 
 
 
 
 
     |

    
    | 
    7
 
 
 
 
 
 
     | 
    8
 
 
 
 
 
 
     | 
    9
 
 
 
 
 
 
     |

    
    | 
    *
 
 
 
 
 
 
     | 
    0
 
 
 
 
 
 
     | 
    #
 
 
 
 
 
 
     |


RunAllCucumberTests > Cucumber STANDARD_OUT

    1 Scenarios (1 passed)
    1 Steps (1 passed)
    0m0.146s


```

### ✅ What did you expect to see?

I expected to see this:

```
$ ./gradlew test
Configuration cache is an incubating feature.
Calculating task graph as no configuration cache is available for tasks: test

> Task :test

RunAllCucumberTests > Cucumber > Demonstrate datatable display bug > com.example.RunAllCucumberTests.Print datatable STANDARD_OUT

    Scenario: Print datatable   # features/DataTable.feature:3
      Given a data table exists # com.example.steps.Steps.aDataTableExists(io.cucumber.datatable.DataTable)
        | ColumnA | ColumnB | ColumnC |
        | 1       | 2       | 3       |
        | 4       | 5       | 6       |
[...]

Issue #2481, opened 2022-02-08
Pretty print plugin performance issues


vkuzel:
Pretty print plugin is slow, because:
* Calls flush method on every printed line.
* Use unbuffered output stream.

Combination of previous points results in poor performance. In my situation the plugin doubles the run time.

### Single thread

When running Cucumber tests in a single thread, the bottleneck is in the `OutputStreamWriter.flush()` method.

![cucumber-junit5-pretty-print-issue-sample-singlethread](https://user-images.githubusercontent.com/3496234/152949059-5bd097a7-7cbf-4164-9290-3c88dd956606.png)

### Multi-threading

When running Cucumber tests in multiple threads, the bottleneck is still in the `flush()` method. Because printing is called from synchronised `SynchronizedEventBus.send()` method, a thread has to wait for `flush()` in other threads, which subsequently kills  performance gains from multi-threaded solution.

![cucumber-junit5-pretty-print-issue-sample-multithread](https://user-images.githubusercontent.com/3496234/152949806-2f0e2b30-ef8b-4103-8a4c-8732a9011062.png)

## Expected behaviour

Performance of the `NiceAppendable` class should be improved to be close to running tests without the plugin. A good idea may be not to flush on every line.

- Version: `io.cucumber:cucumber-java:7.2.3`
- Tests executed from IntelliJ IDEA 2021.3.2
- Build tool: Gradle 5.3.1
- OS: macOS Monterey 12.1

mpkorstanje:
Very interesting. I'll have to take a good look.

scottadav:
I couldn't find a way to attach a zip file, so instead here is a gitlab link to download the sample project from. This was taken from a contributed cucumber example project and modified to demonstrate the defect.

https://github.com/scottadav/cucumber-junit5-example/tree/demonstrate-datatable-gradle-display-defect

mpkorstanje:
The `NiceAppendable` provides:
 * "Real time" output to `System.out` by flushing every append
 * "Real time" output to `System.out` by flushing every line.
 * Wrapping of exceptions
 
The first is important important for the progress formatter which only outputs dots. Not flushing immediately would hide the progress indication in the buffer.

The second is important for the pretty formatter. It doesn't need a flush after every append, but the information for each step should be flushed. However the pretty formatter doesn't need to flush every append and I would expect that the underlying print stream already flushes on new lines.

The third is important for the rerun and unused steps formatters. However neither cares about timely output.

[...]

PR #2537, opened 2022-04-28
Reduce pretty formatter output flushes for DataTables (#2536)

scottadav:
### 🤔 What's changed?

See details in #2536 and #2481 

DataTables were formatted incorrectly by the Pretty formatter when using Gradle. 

The problem stemmed from the fact that Gradle adds a newline with every flush of the output stream, and cucumber-jvm was flushing the output stream very frequently. For DataTables, it was flushing with every cell, padding, and delimiter. This caused an unreadable output. It also reduces I/O performance by flushing excessively.

Make sure it follows project standards.

### 📋 Checklist:

----

### DIFF
--- a/datatable/src/main/java/io/cucumber/datatable/DataTableFormatter.java
+++ b/datatable/src/main/java/io/cucumber/datatable/DataTableFormatter.java
@@ -65,23 +65,31 @@ public final class DataTableFormatter {
             }
         }
 
+        // Buffer the output with a StringBuilder, to avoid excess calls to
+        // flush() from the NiceAppender used by PrettyFormatter.
+        // Gradle adds a newline for every call to flush() unless the most
+        // recent character written was already a newline. That leads
+        // to some VERY difficult to read tables on the console.
+        StringBuilder buffer = new StringBuilder();
+
         // print the rendered cells with padding
         for (int rowIndex = 0; rowIndex < height; rowIndex++) {
-            printRowPrefix(appendable, rowIndex);
-            appendable.append("| ");
+            printRowPrefix(buffer, rowIndex);
+            buffer.append("| ");
             for (int colIndex = 0; colIndex < width; colIndex++) {
                 String cellText = renderedCells[rowIndex][colIndex];
-                appendable.append(cellText);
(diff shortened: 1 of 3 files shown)


mpkorstanje (changes requested):
While this solution reduces the number of flushes, it tries to solve a problem that happens several abstractions down. As a result the problem is not solved fundamentally and will re-occur whenever a flush does not coincide with a newline.

Please consider all my comments from #2481 to develop your solution.

datatable/src/main/java/io/cucumber/datatable/DataTableFormatter.java
```diff
@@ -65,23 +65,31 @@ public void formatTo(DataTable table, Appendable appendable) throws IOException
             }
         }
 
+        // Buffer the output with a StringBuilder, to avoid excess calls to
+        // flush() from the NiceAppender used by PrettyFormatter.
+        // Gradle adds a newline for every call to flush() unless the most
+        // recent character written was already a newline. That leads
+        // to some VERY difficult to read tables on the console.
+        StringBuilder buffer = new StringBuilder();
```
mpkorstanje, line 73: Normally it could be assumed that the `Appendable` is well behaved.  However for the formatter we know that lines may interleave with other output as it writes to `System.out`. Consider pushing this buffer up to the pretty formatter instead.

scottadav:
[mpkorstanje](https://github.com/mpkorstanje), I created a separate pull request https://github.com/cucumber/cucumber-jvm/pull/2541, which might be more acceptable as a general-purpose solution. Could you take a look at that one? If you prefer the other solution, feel free to close this one. Either of them solves my display problem.

datatable/src/main/java/io/cucumber/datatable/DataTableFormatter.java
scottadav: Pushed the buffer up to the pretty formatter by commit f4a2d42 . Only caveat - I can't decide how to test that particular change or whether a unit test is warranted.

mpkorstanje:
Superseded by #2481

Issue #2546, opened 2022-05-02
Reduce duplicated code to test log records


mpkorstanje:
### 🤔 What's the problem you've observed?

We use a `LogRecordListener` to capture log messages and test them. This results in a bit of boiler plate code repeated over several tests:

https://github.com/cucumber/cucumber-jvm/blob/78adb8fe2e4c9d80ccfc36b1a69a8a27bab4ace7/core/src/test/java/io/cucumber/core/runtime/FeaturePathFeatureSupplierTest.java#L30-L39

### ✨ Do you have a proposal for making it better?

Implement [a JUnit 5 extension](https://junit.org/junit5/docs/current/user-guide/#extensions) that provides the `LogRecordListener` as a method argument to tests.

```java
@WithLogRecordListener
class Example {

    @Test
    void example(LogRecordListener logRecordListener){

    }
}
```    

Then replace the existing usages with this extension.

### :hammer_and_wrench: Implementation hints

JUnit 5 supports meta-annotations. So something like this can make it possible in a single annotation:

```java
@Target({ ElementType.TYPE, ElementType.METHOD })
@Retention(RetentionPolicy.RUNTIME)
@ExtendWith(WithLogRecordListener.Extension.class)
public @interface WithLogRecordListener {

	class Extension implements BeforeEachCallback, AfterEachCallback, ParameterResolver {
        // Implement methods 
	}
}
```

#### :test_tube: Testing hints

Since we're re-factoring the existing tests I don't expect any additional tests for the extension itself.

PR #2547, opened 2022-05-03
Warn when feature / glue paths passed as file schemes

### LINKED ISSUE #2543: Warn when src/main or src/test is used in a glue or feature path
### 👓 What did you see?

Novice users often do not understand the difference between the classpath and the filesystem path.

For example this mistake often shows up on stack overflow:

```java
@RunWith(Cucumber.class)
@CucumberOptions(
        features = "src/test/resources/features",
        glue = "src/test/java/com/example"
)
```

This should be

```java
@RunWith(Cucumber.class)
@CucumberOptions(
        features = "classpath:features",
        glue = "com.example"
)
```

Here `scr/main/java/com/example` is a filesystem path while `classpath:com/example` a classpath path.

Due to misunderstanding the difference between these paths novice users often specify paths that are positively not glue paths.

Feature paths often just happen to work because unlike glue features don't have to be on the classpath.

Additionally while classpath uris are supported for glue paths the package name format is preferred to avoid confusion.

When passing any feature path starting with `src/{test,main}/resources` emit a warning that suggests replacing it with `classpath:` prefix.

Whe passing a glue path starts with  `src/{test,main}/{java,kotlin,scala,groovy}` emit a warning that advices the use of the package name.

The warning should include the original path used and the suggested replacement.

For the feature path it should also include an explanation that mentions the possible ambiguity and advice the user to avoid it. Either by using the suggested replacement or moving the feature files away from well known filesystem paths that are put on the classpath.

For glue paths it should inform users that files in these paths are typically compiled by their build tool before the compiled files are put on the classpath when executing. As such they probably meant to use the package name.

### 🛠️ Implementation hints

Feature path parsing is implemented in [`FeaturePath`](https://github.com/cucumber/cucumber-jvm/blob/main/core/src/main/java/io/cucumber/core/feature/FeaturePath.java). Checking for the prefix in `parseAssumeFileScheme` seems like the right place.

Glue paths are parsed in [`GluePath`](https://github.com/cucumber/cucumber-jvm/blob/main/core/src/main/java/io/cucumber/core/feature/GluePath.java). The `parseAssumeClasspathScheme` seems like the right place.

Logging warnings should be done using Cucumbers logging utility.  For example see: https://github.com/cucumber/cucumber-jvm/blob/549fe862b42ac3f0d7aab0a4cc91f05052ef737c/core/src/main/java/cucumber/api/cli/Main.java#L29

### :test_tube: Testing hints

You can test log messages by adding a `LogRecordListener` to the `LoggerFactory`. See the surrounding code for more context:

https://github.com/cucumber/cucumber-jvm/blob/78adb8fe2e4c9d80ccfc36b1a69a8a27bab4ace7/core/src/test/java/io/cucumber/core/runtime/FeaturePathFeatureSupplierTest.java#L30-L39

tal66:
Hi, this closes #2543. changes:
- warn when 'src/{test,main}/resources' used in features path, 
- warn when 'src/{test,main}/{java,kotlin,scala,groovy}' used in glue path
- tests for the above
- had to change core/runtime/FeaturePathFeatureSupplierTest as well, because it checked log messages at a specific index - which now changed. so also added a more convenient static method for this in LogRecordListener, used in my tests as well.

### DIFF
--- a/core/src/main/java/io/cucumber/core/feature/FeaturePath.java
+++ b/core/src/main/java/io/cucumber/core/feature/FeaturePath.java
@@ -34,7 +34,7 @@ public class FeaturePath {
 
     private static final Logger log = LoggerFactory.getLogger(FeaturePath.class);
 
-    private static final Pattern FILESYSTEM_PATH_TO_RESOURCES = Pattern.compile("/*src/(?:main|test)/resources/(.*)");
+    private static final Pattern FILESYSTEM_PATH_TO_RESOURCES = Pattern.compile("src/(?:main|test)/resources/?(.*)");
 
     private FeaturePath() {
 
--- a/core/src/main/java/io/cucumber/core/feature/FeaturePath.java
+++ b/core/src/main/java/io/cucumber/core/feature/FeaturePath.java
@@ -1,8 +1,13 @@
 package io.cucumber.core.feature;
 
+import io.cucumber.core.logging.Logger;
+import io.cucumber.core.logging.LoggerFactory;
+
 import java.io.File;
 import java.net.URI;
 import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static io.cucumber.core.resource.ClasspathSupport.CLASSPATH_SCHEME_PREFIX;
 import static io.cucumber.core.resource.ClasspathSupport.RESOURCE_SEPARATOR_CHAR;
@@ -26,6 +31,10 @@ import static java.util.Objects.requireNonNull;
  * @see FeatureWithLines
  */
 public class FeaturePath {
+
+    private static final Logger log = LoggerFactory.getLogger(FeaturePath.class);
+
+    private static final Pattern FILESYSTEM_PATH_TO_RESOURCES = Pattern.compile("/*src/(?:main|test)/resources/(.*)");
 
     private FeaturePath() {
 
@@ -69,6 +78,7 @@ public class FeaturePath {
     }
 
     private static URI parseAssumeFileScheme(String featureIdentifier) {
+        warnWhenFileSystemPathToResources(featureIdentifier);
         File featureFile = new File(featureIdentifier);
         return featureFile.toURI();
     }
@@ -99,6 +109,15 @@ public class FeaturePath {
             return "";
         }
         return value.toLowerCase(Locale.US).replaceAll("[^a-z0-9]+", "");
+    }
+
+    private static void warnWhenFileSystemPathToResources(String featureIdentifier) {
+        Matcher matcher = FILESYSTEM_PATH_TO_RESOURCES.matcher(featureIdentifier);
+        if (matcher.matches()) {
+            log.warn(() -> String.format("Please replace feature path '%s' with the classpath '%s' to avoid ambiguity.",
+                featureIdentifier,
+                matcher.replaceAll("classpath:$1")));
+        }
     }
 
 }
--- a/core/src/main/java/io/cucumber/core/feature/FeaturePath.java
+++ b/core/src/main/java/io/cucumber/core/feature/FeaturePath.java
@@ -34,7 +34,7 @@ public class FeaturePath {
 
     private static final Logger log = LoggerFactory.getLogger(FeaturePath.class);
 
-    private static final Pattern FILESYSTEM_PATH_TO_RESOURCES = Pattern.compile("src/(?:main|test)/resources/?(.*)");
+    private static final Pattern FILESYSTEM_PATH_TO_RESOURCES = Pattern.compile("src/(?:main|test)/resources(/?)(.*)");
 
     private FeaturePath() {
 
@@ -114,9 +114,13 @@ public class FeaturePath {
     private static void warnWhenFileSystemPathToResources(String featureIdentifier) {
         Matcher matcher = FILESYSTEM_PATH_TO_RESOURCES.matcher(featureIdentifier);
         if (matcher.matches()) {
-            log.warn(() -> String.format("Please replace feature path '%s' with the classpath '%s' to avoid ambiguity.",
-                featureIdentifier,
-                matcher.replaceAll("classpath:$1")));
+            if (!matcher.group(1).equals("/") && matcher.group(2).length() > 0) {
+                return;
+            }
+            log.warn(
+                () -> String.format("Consider replacing feature path '%s' with '%s'.",
+                    featureIdentifier,
+                    matcher.replaceAll("classpath:$2")));
         }
     }
 
--- a/core/src/main/java/io/cucumber/core/feature/FeaturePath.java
+++ b/core/src/main/java/io/cucumber/core/feature/FeaturePath.java
@@ -1,13 +1,8 @@
 package io.cucumber.core.feature;
 
-import io.cucumber.core.logging.Logger;
-import io.cucumber.core.logging.LoggerFactory;
-
 import java.io.File;
 import java.net.URI;
 import java.util.Locale;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import static io.cucumber.core.resource.ClasspathSupport.CLASSPATH_SCHEME_PREFIX;
 import static io.cucumber.core.resource.ClasspathSupport.RESOURCE_SEPARATOR_CHAR;
@@ -31,10 +26,6 @@ import static java.util.Objects.requireNonNull;
  * @see FeatureWithLines
  */
 public class FeaturePath {
-
-    private static final Logger log = LoggerFactory.getLogger(FeaturePath.class);
-
-    private static final Pattern FILESYSTEM_PATH_TO_RESOURCES = Pattern.compile("src/(?:main|test)/resources(/?)(.*)");
 
     private FeaturePath() {
 
@@ -78,7 +69,6 @@ public class FeaturePath {
     }
 
     private static URI parseAssumeFileScheme(String featureIdentifier) {
-        warnWhenFileSystemPathToResources(featureIdentifier);
         File featureFile = new File(featureIdentifier);
         return featureFile.toURI();
     }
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/cucumber/core/logging/LogRecordListener.java
```diff
@@ -17,4 +17,9 @@ public List<LogRecord> getLogRecords() {
         return Arrays.asList(logRecords.toArray(new LogRecord[0]));
     }
 
+    public static boolean anyRecordMessageMatch(LogRecordListener listener, String regex) {
```
mpkorstanje, line 20: You can rewrite any `static` method that has an instance of its own class as an argument to an instance method.


```java
    public boolean anyRecordMessageMatch(String regex) { }
```

And then invoke as

```
logRecordListener.anyRecordMessageMatch(".*");
```

core/src/main/java/io/cucumber/core/feature/GluePath.java
```diff
@@ -29,6 +34,11 @@
  */
 public class GluePath {
 
+    private static final Logger log = LoggerFactory.getLogger(GluePath.class);
+
+    private static final Pattern FILESYSTEM_PATH_TO_PACKAGES = Pattern
+            .compile("/*src/(?:main|test)/(?:java|kotlin|scala|groovy)/(.*)");
```
mpkorstanje, line 40: This pattern suggests that `/src/main/java/com/example` is a file system path to a package. But I think it is unlikely that the file system root is part of any project. So I don't think we have to warn users about this. They'll be warned did not find any glue.

The pattern also suggests that `src/main/java` is not a path to a package (it's the root package). Consider adding a test case for this.

core/src/main/java/io/cucumber/core/feature/FeaturePath.java
```diff
@@ -27,6 +32,10 @@
  */
 public class FeaturePath {
 
+    private static final Logger log = LoggerFactory.getLogger(FeaturePath.class);
+
+    private static final Pattern FILESYSTEM_PATH_TO_RESOURCES = Pattern.compile("/*src/(?:main|test)/resources/(.*)");
```
mpkorstanje, line 37: This pattern suggests that `/src/test/resources/com/example` is a file system path to a package. But I think it is unlikely that the file system root is part of any project. So I don't think we have to warn users about this. They'll be warned about cucumber not being able to find any feature files.

The pattern also suggests that `src/test/resources` is not a path to a package (it's the root package). Consider adding a test case for this.

core/src/main/java/io/cucumber/core/feature/FeaturePath.java
```diff
@@ -101,4 +111,13 @@ private static String normalize(final String value) {
         return value.toLowerCase(Locale.US).replaceAll("[^a-z0-9]+", "");
     }
 
+    private static void warnWhenFileSystemPathToResources(String featureIdentifier) {
+        Matcher matcher = FILESYSTEM_PATH_TO_RESOURCES.matcher(featureIdentifier);
+        if (matcher.matches()) {
+            log.warn(() -> String.format("Please replace feature path '%s' with the classpath '%s' to avoid ambiguity.",
+                featureIdentifier,
+                matcher.replaceAll("classpath:$1")));
```
mpkorstanje, line 119: I think we're missing a bit of explanation here. Some one reading this might wonder:

* What did I do wrong?
* Why should I follow this suggestion, my code works right?

core/src/test/java/io/cucumber/core/options/CucumberOptionsAnnotationParserTest.java
```diff
@@ -218,6 +220,44 @@ void create_with_glue() {
...
+        GluePath.parse("src/main/java/com/example");
+        GluePath.parse("src/test/java/com/package/other_package");
+
+        LoggerFactory.removeListener(logRecordListener);
+
+        assertAll(
+            () -> assertTrue(LogRecordListener.anyRecordMessageMatch(logRecordListener,
+                ".*replace.*src/main/java/com/example.*'com.example'.*")),
+            () -> assertTrue(LogRecordListener.anyRecordMessageMatch(logRecordListener,
+                ".*replace.*src/test/java/com/package/other_package.*'com.package.other_package'.*")));
```
mpkorstanje, line 258: When this assertion fails, the assertion failed exception will not contain any useful information. Consider using matchers instead. They usually generate better exception messages. For example:

```
var records = logRecordListener.getLogRecords().stream()
  .map(LogRecord::getMessage)
  .collect(toList());
  
asserThat(records, containsInAnyOrder(matchesPattern(".*some-pattern.*)));
```  

See:

 * http://hamcrest.org/JavaHamcrest/javadoc/2.2/org/hamcrest/collection/IsIterableContainingInAnyOrder.html#containsInAnyOrder-org.hamcrest.Matcher...-
 * http://hamcrest.org/JavaHamcrest/javadoc/2.2/org/hamcrest/Matchers.html#matchesPattern-java.util.regex.Pattern-

core/src/test/java/io/cucumber/core/options/CucumberOptionsAnnotationParserTest.java
```diff
@@ -218,6 +220,44 @@ void create_with_glue() {
...
+
+    @Test
+    void warn_when_glue_as_filesystem_path() {
+        // warn when 'src/{test,main}/{java,kotlin,scala,groovy}' is used
+
+        LogRecordListener logRecordListener = new LogRecordListener();
+        LoggerFactory.addListener(logRecordListener);
+
+        GluePath.parse("src/main/java/com/example");
+        GluePath.parse("src/test/java/com/package/other_package");
```
mpkorstanje, line 250: Because we're testing multiple inputs for a function of `path -> log message` consider using [Parameterized tests](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests), specifically the `testWithMultiArgMethodSource` example.

This makes it possible to test multiple different inputs/outputs for a function while also having a single test for each case.

mpkorstanje (changes requested):
Thanks! That was quick!

Looks like we're going in the right direction over all but there are a few things to consider:

* It is usually a good idea to keep the tests as close to the code under test as possible. There are dedicated unit tests in `FeaturePathTest` and `GluePathTest`. Consider moving the tests there.

* I do think the `anyRecordMatch` method is pretty neat but I also think that after adding parameterized tests and fixing the root cause of the failure in the `FeaturePathFeatureSupplierTest` it won't be needed anymore. If done right there should only be one log record per test case.

core/src/test/java/io/cucumber/core/options/CucumberOptionsAnnotationParserTest.java
mpkorstanje: And unfortunately we're still using `Hamcrest`. If you're familiar with `AssertJ` feel free to use that.

tal66:
thanks for the feedback, changed accordingly (except for adding explanation about the feature path, because it does work as file scheme, so not sure what more to say)

mpkorstanje:
No problem. I'll think about the explanation a bit.

mpkorstanje (commented):
I think we're almost there now. Currently the regex patterns also match a few things they shouldn't match. Consider adding tests where nothing is logged. 

Hint: You can pass matchers as an argument to the test method.

tal66:
changed it. 
but maybe reconsider whether the feature path warning is needed at all? if it's just a best practice, and has no noticeable effect on the execution, than maybe a warning is not the place to mention that. especially if someone already put their feature files under resources (and those with `src/main/resources-hello` don't get a warning).

core/src/main/java/io/cucumber/core/feature/GluePath.java
```diff
@@ -29,6 +34,11 @@
  */
 public class GluePath {
 
+    private static final Logger log = LoggerFactory.getLogger(GluePath.class);
+
+    private static final Pattern FILESYSTEM_PATH_TO_PACKAGES = Pattern
+            .compile("src/(?:main|test)/(?:java|kotlin|scala|groovy)(|/|/.+)");
```
mpkorstanje, line 40: By having the capture group match either `''`, `'/'` or '/anything else'` we can avoid checking a few edge cases in the warning. We now know the string is either empty or starts with a `/`.

core/src/test/java/io/cucumber/core/feature/GluePathTest.java
```diff
@@ -111,7 +127,44 @@ void absolute_windows_path_form_is_not_valid() {
         Executable testMethod = () -> GluePath.parse("C:\\com\\example\\app");
         IllegalArgumentException actualThrown = assertThrows(IllegalArgumentException.class, testMethod);
         assertThat("Unexpected exception message", actualThrown.getMessage(), is(equalTo(
-            "The glue path must have a classpath scheme C:/com/example/app")));
+                "The glue path must have a classpath scheme C:/com/example/app")));
+    }
+
+    @ParameterizedTest
+    @MethodSource("GluePathAndPatternProvider")
+    void warn_when_glue_as_filesystem_path(String gluePath, Matcher<String> logPattern) {
```
mpkorstanje, line 135: Passing a matcher here instead of a string gives us some flexibility. We can now pass any sort of check. This means that we don't need to have an empty string if there is no message.

mpkorstanje (commented):
I've pushed a few improvements to the `GluePath`. Please let me know what you think about the explanation.

I still have to look at the `FeaturePath`. Hopefully that will come later.

tal66:
> let me know what you think about the explanation

I prefer shorter explanations, that start with a short action "change `<x>` to `<y>`" and might continue with a longer explanations for those interested. but it's a style thing, and your choice :)

mpkorstanje:
That's a good idea.
 - Put the action first in the message. Then the explanation.
 - Removed the warning for `FeaturePath`. While it's not quite right, to reference files in `src/{main,test}/resources` because the build tool might still process them, it is a bit pedantic.
 
 If you can add an entry in the changelog under "Added"  I think we're done.

tal66:
closing this?

mpkorstanje:
Will be soon. It's good!

mpkorstanje (approved):

aslakhellesoy:
Hi @tal66,

Aslak Hellesøy
Creator of Cucumber

mpkorstanje:
Cheer @tal66! Sorry for the delay. Very happy with this. If you're looking for more to do, you may like #2546.

PR #2554, opened 2022-05-20
The project should be possible to build without prior knowledge ..

vitamindit:
The project should be possible to build without prior knowledge about necessary maven goals. 
In a Java developers environment conflicts with dependencies on other Java/Maven projects should be avoided.

- Introduce a variable to specify the required minimum maven version .. The variable is configured to a version which is assumed to be neither too old nor too new and is preferably API compatible to "modern maven-plugins".
- Configure an isolated local Maven repository cache using the maven -s option in .mvn/maven.config (Since maven 3.3.1+) - see https://maven.apache.org/configure.html
- Configure a project-specific settings(...).xml with minimum requirements - this is where the localRepository element points to an isolated local repository. There is no need to create the repository path manually ( this is done by maven).
- .... After cloning the project entering 'mvn' should build it without any further arguments.

### 🤔 What's changed?

- Add .mvn/maven.config ... points to the next file:
- Add .mvn/settings-cucumber-jvm.xml
- Change root pom.xml .. manage version of the enforcer plugin (3.0.0) and enforce mimimum maven version 3.5.4 (from 2018) it should work with 3.3.1 too

When I first started working on the project there were conflicts with the state of my local repository as the cucumber project did not have any minimum requirements. The unmanaged enforcer plugin was not API compatible with my newer maven version. It should be as easy as possible for newcomers to the project to build the project even without prior knowledge.

### DIFF
--- a/.mvn/settings-cucumber-jvm.xml
+++ b/.mvn/settings-cucumber-jvm.xml
@@ -1,0 +1,6 @@
+<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 https://maven.apache.org/xsd/settings-1.0.0.xsd">
+    <localRepository>${user.home}/.m2/repository-cucumber-jvm</localRepository>
+    <interactiveMode>true</interactiveMode>
+    <offline>false</offline>
+</settings>
--- a/pom.xml
+++ b/pom.xml
@@ -151,32 +151,32 @@
     </profiles>
 
     <build>
-    	
-    	<!-- You may run a build without any goal .. just type 'mvn' -->
-    	<defaultGoal>clean install</defaultGoal>
-    	
-    	<plugins>
-    		<plugin>
+
+        <!-- You may run a build without any goal .. just type 'mvn' -->
+        <defaultGoal>clean install</defaultGoal>
+
+        <plugins>
+            <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-enforcer-plugin</artifactId>
                 <executions>
-					<execution>
-						<id>enforce-maven-3-3-1-plus</id>
-						<goals>
-							<goal>enforce</goal>
-						</goals>
-						<configuration>
-							<fail>true</fail>
-							<rules>
-								<requireMavenVersion>
-									<version>${maven.version.min}</version>
-								</requireMavenVersion>                
-							</rules>
-						</configuration>
-					</execution>
-				</executions>
-    		</plugin>
-    		
+                    <execution>
+                        <id>enforce-maven-3-3-1-plus</id>
+                        <goals>
+                            <goal>enforce</goal>
+                        </goals>
+                        <configuration>
+                            <fail>true</fail>
+                            <rules>
+                                <requireMavenVersion>
+                                    <version>${maven.version.min}</version>
+                                </requireMavenVersion>
+                            </rules>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-resources-plugin</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -26,7 +26,11 @@
 
         <!--Maven plugins-->
         <groovy.version>2.5.15</groovy.version>
+        
+        <!--Minimum required maven version - 3.5.4 was released 2018-06-21 --> 
+        <maven.version.min>3.5.4</maven.version.min>
     </properties>
+    
     <scm>
         <connection>scm:git:git://github.com/cucumber/cucumber-jvm.git</connection>
         <developerConnection>scm:git:git@github.com:cucumber/cucumber-jvm.git</developerConnection>
@@ -114,6 +118,11 @@
                             <artifactId>maven-failsafe-plugin</artifactId>
                             <version>3.0.0-M6</version>
                         </plugin>
+                        <plugin>
+			                <groupId>org.apache.maven.plugins</groupId>
+			                <artifactId>maven-enforcer-plugin</artifactId>
+                            <version>3.0.0</version>
+                        </plugin>
                     </plugins>
                 </pluginManagement>
             </build>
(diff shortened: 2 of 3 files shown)


vitamindit:
Since I do not have knowledge about the CI of the project perhaps the following adjustments are still necessary:

The CI should either configure its settings-cucumber-ci.xml (??) explicitly via maven option -s or it might be possible to use profiles within a single settings.xml. I would recommend the first option.

pom.xml
```diff
@@ -26,7 +26,11 @@
 
         <!--Maven plugins-->
         <groovy.version>2.5.15</groovy.version>
+        
+        <!--Minimum required maven version - 3.5.4 was released 2018-06-21 --> 
+        <maven.version.min>3.5.4</maven.version.min>
```
mpkorstanje, line 31: Fever versions to support means fewer candidates for bugs. Currently 3.6.3 is the version used to release Cucumber so that would be preferable.

pom.xml
```diff
@@ -138,10 +147,36 @@
...
+					<execution>
+						<id>enforce-maven-3-3-1-plus</id>
+						<goals>
+							<goal>enforce</goal>
+						</goals>
+						<configuration>
+							<fail>true</fail>
+							<rules>
+								<requireMavenVersion>
+									<version>${maven.version.min}</version>
```
mpkorstanje, line 172: Please in-line this version. The variable is only used once.

pom.xml
```diff
@@ -114,6 +118,11 @@
                             <artifactId>maven-failsafe-plugin</artifactId>
                             <version>3.0.0-M6</version>
                         </plugin>
+                        <plugin>
+			                <groupId>org.apache.maven.plugins</groupId>
+			                <artifactId>maven-enforcer-plugin</artifactId>
+                            <version>3.0.0</version>
```
mpkorstanje, line 124: > When I first started working on the project there were conflicts with the state of my local repository as the cucumber project did not have any minimum requirements. The unmanaged enforcer plugin was not API compatible with my newer maven version.

The enforcer plugin _is managed_ through the parent pom and needs an upgrade there, not in this project: https://github.com/cucumber/cucumber-parent/pull/25. Please remove this.

pom.xml
```diff
@@ -193,7 +228,6 @@
             <plugin>
                 <groupId>org.commonjava.maven.plugins</groupId>
                 <artifactId>directory-maven-plugin</artifactId>
-                <version>1.0</version>
```
mpkorstanje, line 196: This looks like a mistake.

.mvn/settings-cucumber-jvm.xml
```diff
@@ -0,0 +1,6 @@
+<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 https://maven.apache.org/xsd/settings-1.0.0.xsd">
+    <localRepository>${user.home}/.m2/repository-cucumber-jvm</localRepository>
```
mpkorstanje, line 3: > Configure an isolated local Maven repository cache using the maven -s option in .mvn/maven.config (Since maven 3.3.1+) - see https://maven.apache.org/configure.html

I'm missing a motivating reason. Unless you are doing terrible things with your cache this should not be needed.

Please remove.

pom.xml
```diff
@@ -138,10 +147,36 @@
...
             </build>
         </profile>
+        
     </profiles>
 
     <build>
-        <plugins>
+    	
+    	<!-- You may run a build without any goal .. just type 'mvn' -->
+    	<defaultGoal>clean install</defaultGoal>
```
mpkorstanje, line 156: Could you provide some examples or other large projects that use this?

For example the Maven Surefire Project only uses this in the context of a profile rather then a project default. This leads me to believe that this is intended for specific uses rather then a generic `mvn` command.

https://github.com/apache/maven-surefire/blob/acc1e7d907803b1de4f6692194500b5487b6dc91/maven-failsafe-plugin/pom.xml#L225-L228

mpkorstanje (changes requested):
Cheers, but I've got some comments a questions.

Please also fix the indentation.

mpkorstanje:
> The unmanaged enforcer plugin was not API compatible with my newer maven version.

Comes to mind that rather then fixing this project we can provide users with a Maven wrapper instead. However this will result in binaries in git and would have to be backed up with a Github action to verify the Maven wrapper integrity.

The CI is github actions based. You can find the CI files here:

https://github.com/cucumber/cucumber-jvm/blob/main/.github/workflows/build.yml

pom.xml
vitamindit: Unfortunately I didnt see the plugin management of the enforcer in the parent. Did not notice that there is a parent outside the project.

pom.xml
vitamindit: It is - but it seems that this plugin is also managed in the parent?

pom.xml
vitamindit: I explained the main motivation in the PR. I think it is a good practice if the project is self-documenting the way you build it. The whole PR is honestly based on lack of knowledge of project details and pitfalls that I ran into when I tried to build it for the first time.

.mvn/settings-cucumber-jvm.xml
vitamindit: Maybe this is the wrong place to discuss what is terrible and what is not :) However - not terrible is having total clarity about what project the artifacts belong to. I have more than one of them and space problems are no longer an argument.

mpkorstanje: I'm not convinced.

The current size of my personal Maven cache is 5GB. Professionally around 15GB. These are not trivial amounts of data. Especially not once we add duplicates for each project. People joke about `node_modules` and for good reason.

Not to mention the practical aspects. 

- Using individual caches will prevent the use of snapshot versions build locally. This means that snapshot builds can not be used in another project. For the Cucumber org this could be solved by adding the same setup to all other java projects, but then we are losing the advantage of using Maven defaults by having to configure all these projects. And it still doesn't cover the use of snapshots in non-Cucumber Org projects.

- Using individual caches prevents the reuse of common dependencies already downloaded. Given that a significant part of Cucumber was developed on poor internet connections I've come to appreciate this efficiency.

Please remove this the MR.

pom.xml
mpkorstanje: It is. Good.

pom.xml
mpkorstanje: I'm not looking for a motivation. That is reasonably clear here.

Maven works with conventions. So we can assume people will known the main goals of Maven and their purpose. As such we should try to avoid customizing the setup as much as possible. So what I'm looking for is some evidence that setting `defaultGoal` is a well known convention. If you don't know any major projects that use this, a a subject matter expert (such as a conference speaker) recommending this is fine too.

mpkorstanje: Wait, did you start this PR because `mvn test` currently doesn't work?

mpkorstanje:
I've cherry picked the non-controversial changes.

PR #2616, opened 2022-09-18
Always color Maven output in GitHub workflows with -Dstyle.color=always

timtebeek:
Consistently added just behind --batch-mode, as that normally disables color

### 🤔 What's changed?

Always color Maven output in GitHub workflows with -Dstyle.color=always

### ⚡️ What's your motivation? 

Makes it easier to spot any issues, even with the default imperfect red/green color scheme.

### 🏷️ What kind of change is this?

- :bank: Refactoring/debt/DX (improvement to code design, tooling, documentation etc. without changing behaviour)

### ♻️ Anything particular you want feedback on?

Pushed as a branch rather than through a fork; or would you have preferred me to use a fork still?

### 📋 Checklist:

- [x] I agree to respect and uphold the [Cucumber Community Code of Conduct](https://cucumber.io/conduct/)
- [ ] I've changed the behaviour of the code
  - [ ] I have added/updated tests to cover my changes.
- [ ] My change requires a change to the documentation.
  - [ ] I have updated the documentation accordingly.
- [ ] Users should know about my change
  - [x] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../CHANGELOG.md), linking to this pull request.

PR #2617, opened 2022-09-18
Always color all Maven steps, not just the ones using --batch-mode

timtebeek:
Continuation of #2616


mpkorstanje:
Cheers! Also looks like you got the 1200th request that was merged.

timtebeek:
Glad to help, even if it's just in a small way these days. Cheers!

PR #2618, opened 2022-09-20
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/release-java.yml
+++ b/.github/workflows/release-java.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches: [release/*]
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   publish-mvn:
     name: Publish Maven Package
--- a/.github/workflows/test-java.yml
+++ b/.github/workflows/test-java.yml
@@ -14,6 +14,9 @@ on:
       - v7.x.x
       - renovate/**
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   build:
     name: 'Build'


mpkorstanje:
This doesn't look like the correct solution you are highlighting. Looking at [How the permissions are calculated for a workflow job](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#how-the-permissions-are-calculated-for-a-workflow-job).

> The permissions for the GITHUB_TOKEN are initially set to the default setting for the enterprise, organization, or repository. If the default is set to the restricted permissions at any of these levels then this will apply to the relevant repositories. For example, if you choose the restricted default at the organization level then all repositories in that organization will use the restricted permissions as the default. The permissions are then adjusted based on any configuration within the workflow file, first at the workflow level and then at the job level. Finally, if the workflow was triggered by a pull request from a forked repository, and the Send write tokens to workflows from pull requests setting is not selected, the permissions are adjusted to change any write permissions to read only.

sashashura:
It is one of options. It is possible to switch defaults to read-only per org or repository level and I encourage you doing so. Just keep in mind that changing it per org, may break the workflows that really need a subset of write permissions (`issues: write` to comment on issues or `pull-requests: write` to label pull requests, etc.). The decision of how to implement the policy: changing workflows one by one and making sure nothing is broken and then changing it for the whole org or just changing it on the org level and fixing what is needed is totally up to you.

mpkorstanje:
Superseded by https://github.com/cucumber/common/issues/2139

Issue #2637, opened 2022-11-10
Hardcoded scenario (pickle) order


hovi:
### 🤔 What's the problem you've observed?

The library has only 3 hardcoded orderings "lexical", "reverse" and "random" (with possible seed parameter).
The library is not suited for extending this.

### ✨ Do you have a proposal for making it better?

Probably the simplest way to do it with the least amount of code affected by this, is to try to interpret this string as a fully qualified class name and try to create instance of the class in `PickleOrderParser.parse`.

Of course it would be better to allow customizable factory, but I don't know, how much would that affect the whole ecosystem.

### 📚 Any additional context?

I would like to give specific tags order preference over others and then give lexical ordering within the same tag.
The relevant project is here:
https://github.com/iv4xr-project/iv4xr-se-plugin/

Happy to create PR if this makes sense to you.

mpkorstanje:
A PR would be welcome.

Though I would prefer a solution based on [Service Provider Interfaces](https://docs.oracle.com/javase/tutorial/sound/SPI-intro.html). It hides all problems associated with class loading and should play nicely with modules in the future. The `PickleOrder` class could be the service interface. For inspiration consider looking at `ObjectFactoryServiceLoader`.

mpkorstanje:
> All test scenarios should be independent. Thus, there is no need for specific execution order.

There are uses for ordering that are not related to scenario dependence. For example running slow scenarios first, or frequently failing ones first, recently changed first, ect.

Especially when combined with a fail-fast (which Cucumber also doesn't have yet) these help getting feedback faster.

Since we can't quite predict which of these people will want I reckon it is useful to provide an extension point.

hovi:
We have a couple of reasons:
- Our system is very expensive to reset and sometimes non-deterministic
- Some of these tests indeed affect each other so reset is necessary afterwards, completely random execution can cause failures
- Some tests can be run repeatedly without reset
- Some tests can be run only once and then need reset
- We reflect all this with tags
- We can then optimize ordering of tests (while still randomize it to some degree). Even then our test suite takes hours. 
- If we were to do "independent test scenarios", tests would take almost a day due to reset times

jkronegg:
Thanks for clarification. Indeed, very specific needs imply custom sort capability.
For curiosity : what is your Cucumber volumetry (in terms of parameter types, step definitions, features, rules, test scenarios) ?

PR #2639, opened 2022-11-10
PickleOrder extensibility proposition.

hovi:
Playing with #2637 haven't tested the code yet.

DId I understand correctly, as I never used `ServiceLoader` before. You basically want to use it as an alternative to `Class.forName` call?
- Create PickleOrder implementation, register it in `META-INF/services` of my lib.
- In PickleOrderParser iterate through `ServiceLoader<PickleOrder>` and find the one with same class name as `String` argument provided?


mpkorstanje:
Yes, and ideally also list what the available options are when an invalid class name is provided.

If you want to make it fancy you could also add a `name` to the interface and use that instead of the FQN. Then the other orders can use SPI too instead of being hardcoded.

hovi:
+1 to the list of options

SPI classes are created from no param constructor, so `random` implementation would lose possible `seed` parameter or has to stay hardcoded. I am not ure if the `name` is worth it though. For the premade `PickleOrder` classes, simple Map<> is simpler, I wouldn't overcomplicate the interface (I like more direction of throwing it away and using just `Comparator<PickleOrder>`).

Another way is to use SPI for `Factory`, creating `PickleOrder` instances from the `String` argument. That way the creation of `PickleOrder` is customizable and the str argument is available in the factory method and the method can be extended. But honestly I am not sure if it's worth it since cucumber has been here so many years and none requested custom ordering yet :)

It's your call, happy to take either direction.

mpkorstanje:
> (I like more direction of throwing it away and using just Comparator<PickleOrder>).

It's impossible to create a comparator that results in properly random list when using  `Collections.sort`. It requires breaking the contract of the `Comparator`.

At best you can pick a derived property like a hash of a pickle, but that's not what we want.

> SPI classes are created from no param constructor, so random implementation would lose possible seed parameter or has to stay hardcoded. 

Extending the syntax to `plugin-name[:argument]` is also an option. Then we can use a `setArgument` method. If a plugin does something with the argument or not doesn't matter.

> Another way is to use SPI for Factory, creating PickleOrder instances from the String argument. 

I'd prefer avoiding the factory. While immutable objects are nice, that does seem like a lot more overhead.

mpkorstanje:
> But honestly I am not sure if it's worth it since cucumber has been here so many years and none requested custom ordering yet :)

And to make sure we'll never get another request a little bit of over engineering doesn't hurt. 😄

hovi:
Ah right, didn't think of random sorting issue.
Extending the syntax to possible argument sounds good.

You basically suggest solution (SPI) that creates classes with empty constructor, while we want them to have parameter in some cases. And then to add `setArgument` method to help with that. Does that sound good to you? Are you sure that's better than having simple factory (1 method interface + 1 default implementation), which fixes this all and creates nice encapsulated, immutable objects and gives possibility to client to give full control of creating `PickleOrder` objects if they wish so?

mpkorstanje:
I think you've thought about it more then me now. Let's try the factory solution.

hovi:
Looks like I haven't. It would actually require separate configuration for the factory (in case of customization) and for PickleOrder.

mpkorstanje:
Maybe you can try a single pickle order factory for each order. The factory would have a method `String name()` and a method `PickleOrder create(String argument)`.

Then you can use the first half of `name[:argument]` to find the right factory and the second half to create an instance.

hovi:
Good idea! I finally managed to find the time. I see the current state close to final. Just need to throw nicer exception on invalid name and adjust the test accordingly.

cucumber-core/src/main/java/io/cucumber/core/order/factory/LexicalUriPickleOrderFactory.java
```diff
@@ -0,0 +1,17 @@
...
+public class LexicalUriPickleOrderFactory implements PickleOrderFactory {
+
+    @Override
+    public String getName() {
+        return "normal";
+    }
+
+    @Override
+    public PickleOrder create(String argument) {
+        return StandardPickleOrders.lexicalUriOrder();
```
mpkorstanje, line 15: Feel free to inline the implementation from `StandardPickleOrders` and have `StandardPickleOrders` invoke the factories. Then we can deprecate `StandardPickleOrders` and remove it in the future.

cucumber-core/src/main/java/io/cucumber/core/options/PickleOrderParser.java
```diff
@@ -1,43 +1,32 @@
...
+import java.util.ServiceLoader;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 final class PickleOrderParser {
 
-    private static final Logger log = LoggerFactory.getLogger(PickleOrderParser.class);
-
-    private static final Pattern RANDOM_AND_SEED_PATTERN = Pattern.compile("random(?::(\\d+))?");
+    private static final Pattern NAME_VS_ARGUMENT_PATTERN = Pattern.compile("([a-z_]+)(?::(.+))?");
```
mpkorstanje, line 12: I don't think versus is the right word. I would consider `NAME_WITH_ARGUMENT_PATTERN`.

cucumber-core/src/main/java/io/cucumber/core/order/factory/PickleOrderFactory.java
```diff
@@ -0,0 +1,10 @@
+package io.cucumber.core.order.factory;
+
+import io.cucumber.core.order.PickleOrder;
+
+public interface PickleOrderFactory {
```
mpkorstanje, line 5: This class is part of Cucumber public API. 

We are using the API Guardian annotations to document that. Please add  `@API(status = API.Status.EXPERIMENTAL)` to this class. Same for `PickleOrder`.

mpkorstanje (commented):
This looks pretty decent!

Unfortunately aside from the tests we also have a few boring items to pick up.

- Please update the CHANGELOG (best to do that at the very last)
- Please add a section about ordering pickles here: https://github.com/cucumber/cucumber-jvm/tree/main/cucumber-core#cucumber-core
- Write some Javadoc to explain what `PickleOrderFactory` and `PickleOrder` are about.
- Update this bit of documentation: https://github.com/cucumber/cucumber-jvm/tree/main/cucumber-core#properties-environment-variables-system-options
- And unfortunately its duplicate to https://github.com/cucumber/cucumber-jvm/blob/main/cucumber-core/src/main/resources/io/cucumber/core/options/USAGE.txt

cucumber-core/src/main/java/io/cucumber/core/order/factory/LexicalUriPickleOrderFactory.java
```diff
@@ -0,0 +1,17 @@
+package io.cucumber.core.order.factory;
+
+import io.cucumber.core.order.PickleOrder;
+import io.cucumber.core.order.StandardPickleOrders;
+
+public class LexicalUriPickleOrderFactory implements PickleOrderFactory {
```
mpkorstanje, line 6: Please make all classes `final`. This will allow us to make changes in the future without breaking peoples code unexpectedly.

hovi:
No problem, agree with all the comments and will do the remaining necessary changes :-)

hovi:
Okay, the code should be fine now. Not too familiar with the project documentation standards so I just put what made sense to me.

mpkorstanje:
Cheers! I'll try and have a look soon.

cucumber-core/README.md
```diff
@@ -79,6 +80,22 @@ They are  respectively responsible for discovering glue classes, registering
 step definitions, and creating instances of said glue classes. Backend and
 object factory implementations are discovered via SPI.
 
+## Order ##
```
mpkorstanje, line 83: Consider structuring this as 

- Explain what we are talking about (i.e. we can change the order of pickles).
- Explain each build in ordering individually and in detail 
- Explain that custom orders can be implemented by implementing the `PickleFactory` interface.

Move all other details to the Javadoc. People who are going to implement the interface will see it there.

cucumber-core/README.md
```diff
@@ -27,6 +27,7 @@ cucumber.execution.dry-run=     # true or false. default: false
 cucumber.execution.limit=       # number of scenarios to execute (CLI only).
   
 cucumber.execution.order=       # lexical, reverse, random or random:[seed] (CLI only). default: lexical
+                                # Can be customized. To do that, implement `PickleOrderFactory` and register it for [ServiceLoader](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html)
```
mpkorstanje, line 30:
```suggestion
                                # Note: Can be customized. See the documentation below.
```

cucumber-core/src/main/java/io/cucumber/core/order/factory/PickleOrderFactory.java
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Responsible for creating {@link PickleOrder}. Factories use {@link java.util.ServiceLoader}.
+ */
+@API(status = API.Status.EXPERIMENTAL)
+public interface PickleOrderFactory {
+
+    /** The name is used as a lookup when providing the name of order from a command line. */
+    String getName();
+
+    PickleOrder create(String argument);
```
mpkorstanje, line 19: This needs some explanation about the possible `null` value of argument and what it means.

mpkorstanje (commented):
Looks good! 

Unfortunately the docs have some duplication in them. Don't forget to update
 - https://github.com/cucumber/cucumber-jvm/blob/main/cucumber-core/src/main/java/io/cucumber/core/options/Constants.java#L46
 - https://github.com/cucumber/cucumber-jvm/blob/main/cucumber-core/src/main/resources/io/cucumber/core/options/USAGE.txt#L59
 - https://github.com/cucumber/cucumber-jvm/blob/main/cucumber-core/src/main/resources/io/cucumber/core/options/USAGE.txt#L118

cucumber-core/src/main/java/io/cucumber/core/options/PickleOrderParser.java
```diff
@@ -1,43 +1,38 @@
...
-        }
-
-        Matcher matcher = RANDOM_AND_SEED_PATTERN.matcher(argument);
-        if (!matcher.matches()) {
-            throw new IllegalArgumentException("Invalid order. Must be either reverse, random or random:<long>");
+        Matcher matcher = NAME_WITH_ARGUMENT_PATTERN.matcher(argument);
+        if (matcher.matches()) {
+            return parse(matcher.group(1), matcher.group(2));
         }
+        return parse(argument, null);
```
mpkorstanje, line 21: This does need a test. The others can be ignored.

mpkorstanje (changes requested):

cucumber-core/src/main/java/io/cucumber/core/options/PickleOrderParser.java
hovi: Okay, I created specific class for RandomPickleOrder to make it easier to test, that the mechanism returns this kind of order. I like concrete classes rather than anonymous lambdas for this reason :)

This documentation is a bit confusing since it's spread so much. I made some effort.

hovi:
I see there's a lot of fails related to formatting (and were the last time I tried to commit). I use intellij. Is there a format config file for Intellij so that my autoformatted code also passes?

mpkorstanje:
If you run `mvn verify` or `mvn spotless:apply` the code will be formatted. You can find the settings here:

https://github.com/cucumber/cucumber-jvm/tree/main/.spotless

cucumber-core/src/main/java/io/cucumber/core/options/PickleOrderParser.java
mpkorstanje: That's fine. As long as they're package private.

mpkorstanje:
It looks like you forced pushed my changes out. I had added some fixes for Revapi so we had a passing build. Can you get them back in?

hovi:
> If you run `mvn verify` or `mvn spotless:apply` the code will be formatted. You can find the settings here:
> 
> https://github.com/cucumber/cucumber-jvm/tree/main/.spotless

Ah, I tried to run the command before and I got error so I assumed the file is not in the repo without further examination:


```
~/git/cucumber-jvm$ mvn spotless:apply
...
[ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.27.2:apply (default-cli) on project cucumber-jvm: Execution default-cli of goal com.diffplug.spotless:spotless-maven-plugin:2.27.2:apply failed: Unable to locate file with path: ${main.basedir}/.spotless/eclipse-formatter-settings.xml: Could not find resource '${main.basedir}/.spotless/eclipse-formatter-settings.xml'. -> [Help 1]
```

```
~/git/cucumber-jvm$ ls .spotless/eclipse-formatter-settings.xml
.spotless/eclipse-formatter-settings.xml
```

Do I run it from the root of the repo?

> It looks like you forced pushed my changes out. I had added some fixes for Revapi so we had a passing build. Can you get them back in?

Pretty sure I only force-pushed fixes of own commits. I can see your commits still on the branch.

mpkorstanje:
> Do I run it from the root of the repo?

Unfortunately that is the case. The `directory-maven-plugin` doesn't kick in otherwise. It's a bit of a limitation of Maven.

> Pretty sure I only force-pushed fixes of own commits. 

Ah indeed. My bad. The UI on my phone shows things differently.

cucumber-core/src/main/java/io/cucumber/core/options/PickleOrderParser.java
mpkorstanje: The USAGE.txt is still a bit of an embarrassment. I'd love to replace it with something like Pico CLI but that's a long term project that requires fixing quite a few of Cucumbers internal first.

But for the other docs, that is the unfortunate nature of documentation. It's not so bad if you keep in mind that the README's are for users while the Javadoc is for developers. You have to explain somethings twice, but you can keep the details in the Javadoc.

hovi:
> > Do I run it from the root of the repo?
> 
> Unfortunately that is the case. The `directory-maven-plugin` doesn't kick in otherwise. It's a bit of a limitation of Maven.
> 

I asked, because I am running it from the root of the repo, but still getting the error, even on fresh checkout. Even when I try `main `branch. I tried on 2 computers. One with ubuntu and one ubuntu subsystem in windows.

mpkorstanje:
Mmh.  What about running `mvn install`?

hovi:
Let's wrap this one up if possible. :) Doesn't matter if the mvn command doesn't work for me if it works for you and you can check yourself. It is weird thought.

mpkorstanje:
Cheers! Unfortunately I have been a bit short on time for the last few weeks and the time available has gone into more urgent matters.

mpkorstanje:
> Doesn't matter if the mvn command doesn't work for me if it works for you and you can check yourself. It is weird thought.

The build is currently failing on a lack of formatting.  It also appears that some comments have gone unaddressed. 

In other circumstances I would be happy to fix these small imperfections but I'm a little pressed for time at the moment.

Issue #2654, opened 2022-12-06
@MockBean not Working with SpringBootTest and Junit5


sashokbg:
### 👓 What did you see?

When running a SpringBootTest with JUnit5 (Jupiter) all @MockBean annotations are not populated.

### ✅ What did you expect to see?

Running the same test without cucumber configuration works with no issue.

### 📦 Which tool/library version are you using?

Java - JDK 17
SpringBoot 2.7
Junit Jupiter - 5.8.2
cucumber-java - 7.8.1
cucumber-spring - 7.8.1
cucumber-junit-platform-engine - 7.8.1

### 🔬 How could we reproduce it?

Cucumber Test as described in the documentation
```
@Suite
@IncludeEngines("cucumber")
@SelectClasspathResource("features")
@ConfigurationParameter(key = GLUE_PROPERTY_NAME, value = "com.proxym.bankerise.backend.messagecenter.domain")
@ConfigurationParameter(key = FILTER_TAGS_PROPERTY_NAME, value = "@Active")
public class MessageCenterApplicationIT {

}
```

StepsFile

```
@CucumberContextConfiguration
@SpringBootTest(classes = TestConfig.class, webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
@SpringJUnitConfig
public class MessageCenterSteps {
    @Autowired
    private ApplicationContext applicationContext; // <-- works properly injected

    @MockBean
    private EmailChannelAdapter emailChannelAdapter; <-- not working

```

----
 
*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/ISSUE_TEMPLATE/bug_report.md)*

mpkorstanje:
Could you sprinkle some breakpoints in the `MockitoTestExecutionListener` and see what happens?

sashokbg:
Hello, thank you for prompt response !

I was already debugging and it turns out the spring context gets properly populated with the mocked beans.

I put a breakpoint in my @BeforeAll and did inject the Spring ApplicationContext then did in evaluate mode "context.getBean(EmailChannelAdapter.class)" and it returns the mock.

This means that the problem is in the injection of the beans.

After following your advice:

[...]

PR #2655, opened 2022-12-08
Fixed mock beans not injected with @MockBean in JUnit 5

sashokbg:
Fix suggested by https://github.com/mpkorstanje in https://github.com/cucumber/cucumber-jvm/issues/2654

### 🤔 What's changed?

As suggested by @mpkorstanje I have added a call to prepareTestInstance which properly injects the mocked beans from the spring context. 

Fixing #2654 

### 📋 Checklist:

----

### DIFF
--- a/cucumber-spring/src/main/java/io/cucumber/spring/TestContextAdaptor.java
+++ b/cucumber-spring/src/main/java/io/cucumber/spring/TestContextAdaptor.java
@@ -53,6 +53,7 @@ class TestContextAdaptor {
         try {
             Class<?> delegateTestClass = delegate.getTestContext().getTestClass();
             delegateTestInstance = applicationContext.getBean(delegateTestClass);
+            delegate.prepareTestInstance(delegateTestInstance);
             Method dummyMethod = TestContextAdaptor.class.getMethod("cucumberDoesNotHaveASingleTestMethod");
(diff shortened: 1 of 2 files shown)


cucumber-spring/src/main/java/io/cucumber/spring/TestContextAdaptor.java
```diff
@@ -53,6 +53,7 @@ private void notifyTestContextManagerAboutBeforeTestMethod() {
         try {
             Class<?> delegateTestClass = delegate.getTestContext().getTestClass();
             delegateTestInstance = applicationContext.getBean(delegateTestClass);
+            delegate.prepareTestInstance(delegateTestInstance);
```
mpkorstanje, line 56: Note to self: `delegateTestInstance` is used as an implicit condition to determine if `beforeTestMethod` has fired. Consider separating creation+preparation from assignemtn.

mpkorstanje:
Thanks for the PR! I was hoping to get it released today. But while this is part of the solution, I did some research and unless all methods of the test context manager are invoked issues like this will keep popping up.

Now I do know how to invoke those properly, but it will take a bit of time. And will need some proper tests written for it. So I'm afraid you'll have to use the double `@Autowired` and `@MockBean` annotation for a while.

PR #2667, opened 2022-12-20
Improved TestStep creation performance

### LINKED ISSUE #2666: Bad performance during TestStep creation
### 👓 What did you see?

On my Java project with about 400 Cucumber test scenarios and about 150 step definitions, the IntelliJ flame graph shows that creating the TestStep takes about 10% of the total time (that's about 1 second on my project). Most of the time is consumed in the [`TestAbortedExceptions.createIsTestAbortedExceptionPredicate()`](https://github.com/cucumber/cucumber-jvm/blob/main/cucumber-core/src/main/java/io/cucumber/core/runner/TestAbortedExceptions.java) method. See the `flamegraph.png` file in the annexed ZIP file.

I expect the [`TestAbortedExceptions.createIsTestAbortedExceptionPredicate()`](https://github.com/cucumber/cucumber-jvm/blob/main/cucumber-core/src/main/java/io/cucumber/core/runner/TestAbortedExceptions.java) `Predicate<Throwable>` creation to be lightweight, with no noticeable impact on the overall execution time.

I implemented some variants and did some benchmarking to determine what is the best alternative.

Note: some optimisations have been produced using ChatGPT (on 19.12.2022). I could have written the code by myself but wanted to compare what an IA could produce. The produced code is impressive and could have been produced by human developers. But as human developers, ChatGPT makes mistakes (e.g. the first produced code was not able to identify subclasses), so it needs some code review and test harness.

All variants have been tested using the following JUnit5 test (contains a two positive and one negative example):

    @ParameterizedTest
    @MethodSource("methods")
    void testPredicate(String name, Predicate<Throwable> myMethod) {
        assertFalse(myMethod.test(new RuntimeException()));
        assertTrue(myMethod.test(new SkipException()));
        assertTrue(myMethod.test(new SkipExceptionSubClass()));
    }

Even if the implemented variants pass this tests, there are slight differences with the classloading behavior:

- **classloader**: the original Cucumber version takes the current thread classloader (when existing). This way, one Thread could have a classloader with JUnit4 AssumptionViolatedException and another Thread could have a classloader with TestNG SkipException. Such cases are probably rare, but the original code support it. Variants which use a static classloader are not able to support this use-case.
- **caching classes**: the original Cucumber version loads the aborted exceptions classes each time the method is called. Thus, Cucumber is able to manage dynamic classloading (e.g. at first the classloader doesn't know a class, but it loads the class afterwards). Again, this dynamic classloader behavior is probably rare, but the original code support it. 

As these use-cases are very rare, we could agree on changing the original behavior if the performance improvement is significative (to be decided by Cucumber core team).

The table below summarizes the different variants:

| Name                                   | Description                                                                                                                      | classloader | cached classes | code readability | ops/s      |
|----------------------------------------|----------------------------------------------------------------------------------------------------------------------------------|-------------|----------------|------------------|------------|
| createIsTestAbortedExceptionPredicate0 | Original Cucumber 7.10.0 version                                                                                                 | by thread   | no             | low              | 5'517      |
| createIsTestAbortedExceptionPredicate1 | ChatGPT (2nd trial) + human code review + minor human improvements                                                               | by thread   | no             | average          | 261'792'878 |
| createIsTestAbortedExceptionPredicate6 | ChatGPT (3rd trial) + human code review + IntelliJ refactoring + minor human refactoring to make it compile/behave like original | by thread   | no             | high             | 266'445'123 |
| createIsTestAbortedExceptionPredicate5 | Optimized by a human developer, with ThreadLocal cached Predicate                                                                | by thread   | yes            | average          | 260'493'066 |
| createIsTestAbortedExceptionPredicate2 | ChatGPT (3rd trial) + human code review + IntelliJ refactoring + minor human refactoring to make it compile                      | static      | no             | high             | 443'061'213 |
| createIsTestAbortedExceptionPredicate3 | ChatGPT (4rth trial with caching) + human code review + minor human refactoring to make it compile  + minor human improvements   | static      | yes            | low              | 443'764'493 |
| createIsTestAbortedExceptionPredicate4 | Optimized by a human developer                                                                                                   | static      | yes            | average          | 443'207'453 |

Based on the results above, my recommendation would be to use `createIsTestAbortedExceptionPredicate6` which keeps the same classloader behavior ("by thread"), without cached classes. This way, the original behavior is conserved, the code readability is improved and the performance is improved (about 48'000 times faster). We could have even better performance, but this would require to change the original behavior (e.g. `createIsTestAbortedExceptionPredicate5` or `createIsTestAbortedExceptionPredicate3`), which may cause side effects on Cucumber users.

Below is the proposed method:

    static Predicate<Throwable> createIsTestAbortedExceptionPredicate6() {
        ClassLoader defaultClassLoader = ClassLoaders.getDefaultClassLoader();
        return throwable -> Arrays.stream(TEST_ABORTED_EXCEPTIONS)
                .anyMatch(s -> {
                    try {
                        Class<?> aClass = defaultClassLoader.loadClass(s);
                        return aClass.isInstance(throwable);
                    } catch (Throwable t) {
                        rethrowIfUnrecoverable(t);
                        log.debug(t,
                                () -> String.format("Failed to load class %s: will not be supported for aborted executions.", s));
                    }
                    return false;
                });
    }

I can do a PR with the solution preferred by the Cucumber core team.

Cucumber 7.10.0

The ZIP file in annex contains the JMH benchmark and the implemented variants. 
[runner.zip](https://github.com/cucumber/cucumber-jvm/files/10268941/runner.zip)

These files can be copied in a minimal Maven project with the following dependencies:

        <dependency>
            <scope>test</scope>
        </dependency>
        <dependency>
            <artifactId>cucumber-junit-platform-engine</artifactId>
            <scope>test</scope>
            <exclusions>
                <exclusion>
                    <groupId>org.junit.platform</groupId>
                    <artifactId>junit-platform-engine</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
        <dependency>
            <artifactId>cucumber-picocontainer</artifactId>
            <scope>test</scope>
        </dependency>
        <dependency>
            <groupId>org.openjdk.jmh</groupId>
            <artifactId>jmh-generator-annprocess</artifactId>
            <version>1.36</version>
            <scope>test</scope>
        </dependency>

jkronegg:
### 🤔 What's changed?

Refactored `TestAbortedExceptions` `Predicate` creation.

Performance gain (`Predicate` creation is about 50'000 times faster).
Fixes #2666 

- :bank: Refactoring/debt/DX (improvement to code design, tooling, documentation etc. without changing behaviour)

Nothing.

### 📋 Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- [Core] Improve test step creation performance ([#2666](https://github.com/cucumber/cucumber-jvm/issues/2666), Julien Kronegg)
 
 ## [7.10.1] - 2022-12-16
 ### Fixed
@@ -30,7 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - [Core] Emit exceptions on failure to handle test run finished events ([#2651](https://github.com/cucumber/cucumber-jvm/issues/2651) M.P. Korstanje)
 - [Spring] @MockBean annotation not working with JUnit5 ([#2654](https://github.com/cucumber/cucumber-jvm/pull/2654) Alexander Kirilov, M.P. Korstanje)
-- [Core] Improve expression creation performance ([cucumber-expressions/#187](https://github.com/cucumber/cucumber-expressions/pull/187), [cucumber-expressions/#189](https://github.com/cucumber/cucumber-expressions/pull/189), jkrongegg)
+- [Core] Improve expression creation performance ([cucumber-expressions/#187](https://github.com/cucumber/cucumber-expressions/pull/187), [cucumber-expressions/#189](https://github.com/cucumber/cucumber-expressions/pull/189), Julien Kronegg)
 
 ## [7.9.0] - 2022-11-01
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
 - [Core] Improve test step creation performance ([#2666](https://github.com/cucumber/cucumber-jvm/issues/2666), Julien Kronegg)
 
 ## [7.10.1] - 2022-12-16
--- a/cucumber-core/src/main/java/io/cucumber/core/runner/TestAbortedExceptions.java
+++ b/cucumber-core/src/main/java/io/cucumber/core/runner/TestAbortedExceptions.java
@@ -34,7 +34,8 @@ final class TestAbortedExceptions {
                     } catch (Throwable t) {
                         rethrowIfUnrecoverable(t);
                         log.debug(t,
-                                () -> String.format("Failed to load class %s: will not be supported for aborted executions.", s));
+                            () -> String.format(
+                                "Failed to load class %s: will not be supported for aborted executions.", s));
                     }
                     return false;
                 });
--- a/cucumber-core/src/main/java/io/cucumber/core/runner/TestAbortedExceptions.java
+++ b/cucumber-core/src/main/java/io/cucumber/core/runner/TestAbortedExceptions.java
@@ -6,7 +6,6 @@ import io.cucumber.core.resource.ClassLoaders;
 
 import java.util.Arrays;
 import java.util.function.Predicate;
-import java.util.stream.Stream;
 
 import static io.cucumber.core.exception.UnrecoverableExceptions.rethrowIfUnrecoverable;
 
--- a/cucumber-core/src/main/java/io/cucumber/core/runner/TestAbortedExceptions.java
+++ b/cucumber-core/src/main/java/io/cucumber/core/runner/TestAbortedExceptions.java
@@ -26,20 +26,18 @@ final class TestAbortedExceptions {
 
     static Predicate<Throwable> createIsTestAbortedExceptionPredicate() {
         ClassLoader defaultClassLoader = ClassLoaders.getDefaultClassLoader();
-        return Arrays.stream(TEST_ABORTED_EXCEPTIONS)
-                .flatMap(s -> {
+        return throwable -> Arrays.stream(TEST_ABORTED_EXCEPTIONS)
+                .anyMatch(s -> {
                     try {
                         Class<?> aClass = defaultClassLoader.loadClass(s);
-                        return Stream.of(aClass);
+                        return aClass.isInstance(throwable);
                     } catch (Throwable t) {
                         rethrowIfUnrecoverable(t);
                         log.debug(t,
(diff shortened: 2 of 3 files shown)


cucumber-core/src/main/java/io/cucumber/core/runner/TestAbortedExceptions.java
```diff
@@ -26,20 +26,18 @@ final class TestAbortedExceptions {
 
     static Predicate<Throwable> createIsTestAbortedExceptionPredicate() {
         ClassLoader defaultClassLoader = ClassLoaders.getDefaultClassLoader();
-        return Arrays.stream(TEST_ABORTED_EXCEPTIONS)
-                .flatMap(s -> {
+        return throwable -> Arrays.stream(TEST_ABORTED_EXCEPTIONS)
```
mpkorstanje, line 29: Would a for loop be faster?

mpkorstanje:
If you run `mvn clean install` it will test and fix the formatting.

cucumber-core/src/main/java/io/cucumber/core/runner/TestAbortedExceptions.java
jkronegg: I benchmarked the following code:

static Predicate<Throwable> createIsTestAbortedExceptionPredicate7() {
        ClassLoader defaultClassLoader = ClassLoaders.getDefaultClassLoader();
        return throwable -> {
            for (String s : TEST_ABORTED_EXCEPTIONS) {
                try {
                    Class<?> aClass = defaultClassLoader.loadClass(s);
                    if (aClass.isInstance(throwable)) {
                        return true;
                    }
                } catch (Throwable t) {
                    rethrowIfUnrecoverable(t);
                    log.debug(t,
                            () -> String.format(
                                    "Failed to load class %s: will not be supported for aborted executions.", s));
                }
            }
            return false;
        };
    }

The JMH benchmark says the variant 7 (the one with a loop) is slightly slower than the proposed version (variant 6), but not significantly :

Benchmark                                                               Mode  Cnt          Score          Error  Units
TestAbortedExceptionsBenchmark.createIsTestAbortedExceptionPredicate5  thrpt   25  264824252,996 ±  4775342,657  ops/s
TestAbortedExceptionsBenchmark.createIsTestAbortedExceptionPredicate6  thrpt   25  259886661,320 ±  1766902,768  ops/s
TestAbortedExceptionsBenchmark.createIsTestAbortedExceptionPredicate7  thrpt   25  254627500,665 ± 12269987,245  ops/s

I think the code readability is better with the Stream+anyMatch than with the loop, so I would prefer to keep the proposed version..

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- [Core] Improve test step creation performance ([#2666](https://github.com/cucumber/cucumber-jvm/issues/2666), Julien Kronegg)
```
mpkorstanje, line 13:
```suggestion
### Fixed
- [Core] Improve test step creation performance ([#2666](https://github.com/cucumber/cucumber-jvm/issues/2666), Julien Kronegg)
```

aslakhellesoy:
Hi @jkronegg,

Aslak Hellesøy
Creator of Cucumber

mpkorstanje:
Cheers!

PR #2703, opened 2023-03-14
[Core] Add faster UUID generator selectable through SPI

### LINKED ISSUE #2698: Improve feature loading performance by changing UUID generator
### 👓 What did you see?

The Cucumber Java event bus uses UUID as event identifiers. These UUIDs are generated using the JDK `java.util.UUID.randomUUID()`.
This kind of generators known to be slow (e.g. https://stackoverflow.com/questions/14532976/performance-of-random-uuid-generation-with-java-7-or-java-6). On my project (~450 test scenarios, ~50 rules) this UUID generator takes about 25% of the feature file loading time on the IntelliJ profiler flame graph (~2% of the total CPU time).

This generator is used for each each rule and each scenario step line : that corresponds to ~100'000 calls for my project.

Using a faster UUID generator would improve the feature file loading performance. For example, https://github.com/cowtowncoder/java-uuid-generator provides a time sequence UUID generator which is much faster. 

From what I've seen, the solution could be very simple as the `EventBus` is instanciated in `io.cucumber.core.runtime.Runtime.Builder` with `UUID::randomUUID` : we need to change this generator to get a good performance improvement. Note that other calls to `UUID.randomUUID()` are also made elsewhere in the project but I didn't saw big performance impact.

Cucumber Java 7.11.1

1. Run IntelliJ profiler on a project with about 50 rules with cucumber-java 7.11.1.
2. Look for the `loadFeatures` block on the flame graph: on the top is the `java.util.UUID.randomUUID()` call
[uuid_flame_graph.zip](https://github.com/cucumber/cucumber-expressions/files/10825917/uuid_flame_graph.zip)

### 📚 Any additional context?

N/A

jkronegg:
### 🤔 What's changed?

The cucumber event bus UUID generator can now be configured through a SPI using a properties from:
- command-line
- environment variables
- system properties
- `cucumber.properties`
- `junit-platform.properties`
- `@CucumberOptions`

Two UUID generators are provided:

An event has a UUID. The UUID generator can be configured using the `cucumber.uuid-generator` property:

| UUID generator                                 | Features                | Performance [Millions UUID/second] | Typical usage example                                                                                                                                                                                                                                             | 
|------------------------------------------------|-------------------------|------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| io.cucumber.core.eventbus.RandomUuidGenerator | Thread-safe, collision-free, multi-jvm  | ~1                                 | Reports may be generated on different JVMs at the same time. A typical example would be one suite that tests against Firefox and another against Safari. The exact browser is configured through a property. These are then executed concurrently on different Gitlab runners. |
| io.cucumber.core.eventbus.IncrementingUuidGenerator    | Thread-safe, collision-free, single-jvm | ~130                               | Reports are generated on a single JVM                                                                                                                                                                                                                                          |

Performance on real projects depends on the size (e.g. on a project with 3095 Given, 445 When, 1177 Then, 445 Scenarios, 48 Rules, the `IncrementingUuidGenerator` improves the global performance by about 3-5% and the feature file parsing performance of about 37%).

The `IncrementingUuidGenerator` is a very simple UUID generator based on `AtomicLong` counters. It is thread-safe and collision-free within a single JVM.

The `RandomUuidGenerator` is a usual `UUID.randomUUID()` generator.

Fixes #2698

Files of particular interest are:

- `IncrementingUuidGenerator`
- `RandomUuidGenerator`
- `README.md`
- `UuidGeneratorServiceLoader`

After this PR, I'll add a description on https://cucumber.io/docs/cucumber/state/?lang=java#the-cucumber-object-factory


### 📋 Checklist:


jkronegg:
How to solve the Semver build error? Should I just change the version to `7.12.1-SNAPSHOT` ? Most probably it is caused by 

    [ERROR] java.method.addedToInterface: method java.lang.Class<? extends io.cucumber.core.eventbus.UuidGenerator> io.cucumber.core.options.CucumberOptionsAnnotationParser.CucumberOptions::uuidGenerator(): Method was added to an interface. https://revapi.org/revapi-java/differences.html#java.method.addedToInterface

Also I saw that the test coverage is not so good, I'll work on that...

mpkorstanje:
Incrementing the minor version should make the semver check pass. You can use `mvn versions:set` for this.

cucumber-core/src/main/java/io/cucumber/core/eventbus/FastUuidGenerator.java
```diff
@@ -0,0 +1,32 @@
...
+     * @throws CucumberException when out of capacity
+     */
+    @Override
+    public UUID get() {
+        long leastSigBits = counter.incrementAndGet();
+        if (leastSigBits == Long.MAX_VALUE) {
+            throw new CucumberException(
+                "Out of FastUuidGenerator capacity. Please use the DefaultUuidGenerator instead.");
+        }
+        return new UUID(Thread.currentThread().getId(), leastSigBits);
```
mpkorstanje, line 30: What is the purpose of the current thread ID? I reckon we are tempting Hyrum's Law by using it.

If a value must be used here, we could use a random value that is determined on initialization (a session Id of sorts). That would solve some of the collision concerns.

cucumber-core/src/main/java/io/cucumber/core/eventbus/DefaultUuidGenerator.java
```diff
@@ -0,0 +1,14 @@
+package io.cucumber.core.eventbus;
+
+import java.util.UUID;
+
+/**
+ * Default UUID generator. The generator is thread-safe and supports multi-jvm
+ * usage of Cucumber.
+ */
+public class DefaultUuidGenerator implements UuidGenerator {
```
mpkorstanje, line 9: Please rename this to `RandomUuidGenerator`. Should we ever switch to the fast generator as a default we won't get stuck on that.

cucumber-core/src/main/java/io/cucumber/core/runtime/UuidGeneratorServiceLoader.java
```diff
@@ -0,0 +1,115 @@
...
+                "the classpath?");
+    }
+
+    private static String getMultipleUuidGeneratorLogMessage(UuidGenerator... uuidGenerators) {
+        String factoryNames = Stream.of(uuidGenerators)
+                .map(Object::getClass)
+                .map(Class::getName)
+                .collect(Collectors.joining(", "));
+
+        return "More than one Cucumber UuidGenerator was found on the classpath\n" +
```
mpkorstanje, line 104: This exception needs some consideration. We have multiple generators by default so I don't think it is applicable.

mpkorstanje (commented):
Didn't look at everything yet but these stood out immediately.

jkronegg:
Incrementing to `7.12.0-SNAPSHOT` solved some semver issues, but I still have:

```
[ERROR] Failed to execute goal org.revapi:revapi-maven-plugin:0.15.0:check (check) on project cucumber-core: The following API problems caused the build to fail:
[ERROR] java.method.addedToInterface: method java.lang.Class<? extends io.cucumber.core.eventbus.UuidGenerator> io.cucumber.core.options.CucumberOptionsAnnotationParser.CucumberOptions::uuidGenerator(): Method was added to an interface. https://revapi.org/revapi-java/differences.html#java.method.addedToInterface
[ERROR] java.method.addedToInterface: method java.lang.Class<? extends io.cucumber.core.eventbus.UuidGenerator> io.cucumber.core.runner.Options::getUuidGeneratorClass(): Method was added to an interface. https://revapi.org/revapi-java/differences.html#java.method.addedToInterface
[ERROR] 
```

As the `CucumberOptionsAnnotationParser` and the `Options` are internal Cucumber classes, is it a false positive ?

mpkorstanje:
We can consider them false positives. You can add it to the `internal` section in this file:

https://github.com/cucumber/cucumber-jvm/blob/main/.revapi/api-changes.json

cucumber-core/src/main/java/io/cucumber/core/eventbus/FastUuidGenerator.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Thread-safe UUID generator for single JVM. This is a sequence generator and
+ * each instance has its own counter. This generator is about 100 times faster
+ * than #DefaultUuidGenerator. If you use Cucumber in multi-JVM setup, you
+ * should use #DefaultUuidGenerator instead.
+ */
+public class FastUuidGenerator implements UuidGenerator {
```
mpkorstanje, line 14: While it's hard to imagine that a faster generator could be created, I don't think the implementation speed is the property that should be used to name this. I think something like `IncrementingUuidGenerator` would be better.

That is, assuming we don't change the implementation to be something else.

cucumber-core/src/main/java/io/cucumber/core/eventbus/FastUuidGenerator.java
jkronegg: True, I made a mistake here, mixing the "thread-safety" with the "collision" features. The thread-safe feature is provided through the `AtomicLong`, so we don't need the current thread identifier to at the generation time. 

To avoid the collisions, using the current thread identifier may not be the best approach, because one thread could have two `IncrementingUuidGenerator`s, When using a pure random session identifier is not ideal as well, as there is a small collision risk. But an incrementing session identifier (e.g. `static final AtomicLong` field) would do the trick. I'll refactor.

cucumber-core/src/main/java/io/cucumber/core/runtime/UuidGeneratorServiceLoader.java
jkronegg: I'll continue to work on this part.

jkronegg: @mpkorstanje I added some test cases on the `UuidGeneratorServiceLoader` (which revealed one issue regarding the exception message above).

Then I applied the same concepts on `ObjectFactoryServiceLoader`, but kept the tests `@Disabled` because they are not passing: https://github.com/cucumber/cucumber-jvm/pull/2703/files#diff-cbb57a5f00daaf5e93ea9bf64b59ae727113b0af3e5a98b70d06ad5118864766 . Thus, either we have an issue in the tests or in the code.

Could you please confirm me the expected `ObjectFactoryServiceLoader` behavior ?

| Option property `object-factory` | `ObjectFactory` in META-INF/services/... file | Expected behavior                    |
| ----------------------------------- | ------------------------------------------------ | --------------------------------- |
| not specified                               | DefaultObjectFactory                                      | uses DefaultObjectFactory       |
| not specified                               | OtherObjectFactory                                         | uses OtherObjectFactory        |
| not specified                               | none                                                               | exception "could not find any" |
| not specified                               | DefaultObjectFactory + OtherObjectFactory  | exception "more than one"      |
| not specified                               | OtherObjectFactory + DefaultObjectFactory  | exception "more than one"      |

(the two new `@Disabled` test cases cover the last two lines of the table above).

mpkorstanje: I think the current problem is caused by having two "defaults". The easy way around it would be to keep the fast/incrementing UUID generator in your own project.

If we do want to keep both implementations then things get rather complicated because we no longer have a good way to make the current default gracefully step aside. 

So the correct solution would be to 1) non-defaults are prioritized. 2) When only default are available, use the primary default.

But in the interest of simplicity I think having only one "default" implementation is better.

jkronegg: I don't think there is a problem with the UUID generator. 

Let's make some use-cases... (I'll keep my question on `ObjectFactory` aside for the moment).
Note: the use-cases are ordered by decreasing estimated popularity.

### Use-case 1: multiple UUID generators from Cucumber

My application uses Cucumber and no other UUID generator is provided. It will be only one  `META-INF/services/io.cucumber.core.eventbus.UuidGenerator` files on the classpath. Thus, when the `UuidGeneratorServiceLoader` runs, the `ServiceLoader.iterator()` will have the following UUID generators (package name not mentioned for simplicity):
- `RandomUuidGenerator` (coming from Cucumber META-INF)
- `IncrementingUuidGenerator` (coming from Cucumber META-INF)

The `uuid-generator` selection parameter is left unspecified.

### Use-case 2: multiple UUID generators from multiple sources

My application uses Cucumber with two other third-party libraries which implement Cucumber UUID generator SPI (respectively `LibraryOneUuidGenerator` and `LibraryTwoUuidGenerator`). 

When the application is packaged in a classical way (i.e. not a fat jar), it will be 3 `META-INF/services/io.cucumber.core.eventbus.UuidGenerator` files on the classpath. 
Thus, when the `UuidGeneratorServiceLoader` runs, the `ServiceLoader.iterator()` will have the following UUID generators (package name not mentioned for simplicity):
- `RandomUuidGenerator` (coming from Cucumber META-INF)
- `IncrementingUuidGenerator` (coming from Cucumber META-INF)
- `LibraryOneUuidGenerator` (coming from LibraryOne META-INF)
- `LibraryTwoUuidGenerator` (coming from LibraryTwo META-INF)

The `uuid-generator` selection parameter is left unspecified.

### Use-case 3: only one UUID generator

My application uses Cucumber and I packaged it as a fat jar with a single one custom `META-INF/services/io.cucumber.core.eventbus.UuidGenerator` files on the classpath. Let's say I specified only one UUID generator in this file, a custom `MyApplicationUuidGenerator`. Thus when the `UuidGeneratorServiceLoader` runs, the `ServiceLoader.iterator()` will have the UUID generators defined in this file:
- `MyApplicationUuidGenerator`

The `uuid-generator` selection parameter is left unspecified.

### Use-case 4: multiple UUID generators without the `RandomUuidGenerator`

My application uses Cucumber and I packaged it as a fat jar with a single one custom `META-INF/services/io.cucumber.core.eventbus.UuidGenerator` files on the classpath. Let's say I specified only two UUID generators in this file, a custom `MyApplicationUuidGenerator` and Cucumber's `IncrementingUuidGenerator`. Thus when the `UuidGeneratorServiceLoader` runs, the `ServiceLoader.iterator()` will have the UUID generators defined in this file:
- `MyApplicationUuidGenerator`
-  `IncrementingUuidGenerator`

The `uuid-generator` selection parameter is left unspecified.

### Use-case 5: no UUID generator

My application uses Cucumber and I packaged it as a fat jar with a single one custom `META-INF/services/io.cucumber.core.eventbus.UuidGenerator` files on the classpath. This file is empty. Thus when the `UuidGeneratorServiceLoader` runs, the `ServiceLoader.iterator()` will have the no UUID generators defined.

The `uuid-generator` selection parameter is left unspecified.

### Expected behavior

| Use-case | UUID generators available on SPI | Expected selected UUID generator |
| -----------| ------------------------------------| ---------------------------------------|
| Use-case 1 | `RandomUuidGenerator` <br> `IncrementingUuidGenerator` |  `RandomUuidGenerator`  |
| Use-case 2 | `RandomUuidGenerator` <br> `IncrementingUuidGenerator` <br> `LibraryOneUuidGenerator` <br> `LibraryTwoUuidGenerator` |  `RandomUuidGenerator`  |
| Use-case 3 | `MyApplicationUuidGenerator` | `MyApplicationUuidGenerator` |
| Use-case 4 | `MyApplicationUuidGenerator` <br> `IncrementingUuidGenerator` | error "more than one UUID generator without selection" |
| Use-case 5 | none | error "no UUID generator found as SPI" |

So the rule is (when no "uuid-generator" selection parameter is provided) :
- when the `RandomUuidGenerator` is present, use it
- when a single `UuidGenerator` is present, use it
- when multiple `UuidGenerator` are present (without `RandomUuidGenerator`), raise an error "more than one UUID generator without selection"
- when no `UuidGenerator` is present, raise an error "no UUID generator found as SPI"

(note that the code must be robust to the UUID generators order in the SPI iterator: it must work no matter in the `RandomUuidGenerator` is the first or the last).

The code in this PR is implemented for the rules and test use-cases above.

jkronegg: That's fine for me, I just need to use `IncrementingUuidGenerator` only when specified in the uuid-generator property.

That is (same as you wrote, with changes in **bold**) :

| # | uuid-generator property | Available services | Result |
|-|-|-|-|
 | 1 | undefined |  none | exception, no generators available |
 | 2 | undefined |  RandomUuidGenerator, **IncrementingUuidGenerator** | RandomUuidGenerator used |
 | 3 | RandomUuidGenerator |  RandomUuidGenerator, **IncrementingUuidGenerator** | RandomUuidGenerator used |
 | 4 | undefined |  RandomUuidGenerator, **IncrementingUuidGenerator**, OtherGenerator | OtherGenerator used |
 | 5 | RandomUuidGenerator |  RandomUuidGenerator, **IncrementingUuidGenerator**, OtherGenerator | RandomUuidGenerator used |
 | 6 | undefined |  RandomUuidGenerator, **IncrementingUuidGenerator**, OtherGenerator, YetAnotherGenerator |  exception, cucumber couldn't  decide multiple (non default) generators available |
 | 7 | OtherGenerator |  RandomUuidGenerator, **IncrementingUuidGenerator**, OtherGenerator, YetAnotherGenerator | OtherGenerator used |
 | 8 | **IncrementingUuidGenerator** |  RandomUuidGenerator, **IncrementingUuidGenerator**, OtherGenerator, YetAnotherGenerator | **IncrementingUuidGenerator** used |

I think the `IncrementingUuidGenerator` must be included in Cucumber to be developer-friendly (the developer just need to set a property).

mpkorstanje: If you can make that work, sure.

jkronegg:
@mpkorstanje I've finished the work I planned to do on this PR. Do you see anything to change ?

cucumber-core/src/test/java/io/cucumber/core/eventbus/IncrementingUuidGeneratorTest.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Test
+    void raises_exception_when_out_of_range() throws NoSuchFieldException, IllegalAccessException {
+        // Given
+        UuidGenerator generator = new IncrementingUuidGenerator();
+        Field counterField = IncrementingUuidGenerator.class.getDeclaredField("counter");
+        counterField.setAccessible(true);
+        AtomicLong counter = (AtomicLong) counterField.get(generator);
+        counter.set(Long.MAX_VALUE - 1);
```
mpkorstanje, line 37: Consider a package private constructor to initialize a certificain problem for testing instead of using reflection.

cucumber-core/src/test/java/io/cucumber/core/options/CommandlineOptionsParserTest.java
```diff
@@ -84,7 +95,7 @@ void has_version_from_properties_file() {
     }
 
     private String output() {
-        return new String(out.toByteArray(), StandardCharsets.UTF_8);
+        return out.toString(StandardCharsets.UTF_8);
```
mpkorstanje, line 98: I don't think this works on 8. I would be okay with using a different test source and target level though.

https://maven.apache.org/plugins/maven-compiler-plugin/testCompile-mojo.html#testSource

cucumber-core/src/test/java/io/cucumber/core/runtime/UuidGeneratorServiceLoaderTest.md
```diff
@@ -0,0 +1,17 @@
+# Testcases for `UuidGeneratorServiceLoader`
```
mpkorstanje, line 1: Feels like this should be Javadoc on UuidGeneratorServiceLoader.

cucumber-core/src/test/java/io/cucumber/core/runtime/ObjectFactoryServiceLoaderTest.md
```diff
@@ -0,0 +1,18 @@
+# Testcases for `ObjectFactoryServiceLoader`
```
mpkorstanje, line 1: This too should be Javadoc

cucumber-junit/src/main/java/io/cucumber/junit/NoUuidGenerator.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+final class NoUuidGenerator implements UuidGenerator {
+
+    private NoUuidGenerator() {
+        // No need for instantiation
+    }
+
+    @Override
+    public UUID get() {
+        return null;
```
mpkorstanje, line 18: Note to self: Write an issue with the considerations of using a null-safe API, returning null only on methods annotated with nullable. Specifically investigate automated checks.

mpkorstanje (commented):
Reviewed this from my phone. Looks good overal. A few nitpicks and some minor stuff. I didn't see anything major.

I also left a few items for myself, please leave those open.

cucumber-core/src/test/java/io/cucumber/core/options/CommandlineOptionsParserTest.java
jkronegg: True, this is Java 10, so I reverted.

cucumber-core/src/test/java/io/cucumber/core/eventbus/IncrementingUuidGeneratorTest.java
jkronegg: Used package private visibility.

cucumber-core/src/main/java/io/cucumber/core/eventbus/UuidGenerator.java
```diff
@@ -0,0 +1,13 @@
...
+import org.apiguardian.api.API;
+
+import java.util.UUID;
+import java.util.function.Supplier;
+
+/**
+ * SPI (Service Provider Interface) to generate UUIDs.
+ */
+@API(status = API.Status.STABLE)
+public interface UuidGenerator extends Supplier<UUID> {
```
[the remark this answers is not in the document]
jkronegg: Added `UUID generateId()` method on the interface and a `default UUID get()` to keep the `Supplier<UUID>` interface.

cucumber-core/src/main/java/io/cucumber/core/eventbus/IncrementingUuidGenerator.java
```diff
@@ -0,0 +1,40 @@
...
+ * UUID version and variant is not guaranteed to be stable.
+ */
+public class IncrementingUuidGenerator implements UuidGenerator {
+    private static final AtomicLong sessionCounter = new AtomicLong(Long.MIN_VALUE);
+
+    private final long sessionId;
+    private final AtomicLong counter = new AtomicLong(Long.MIN_VALUE);
+
+    public IncrementingUuidGenerator() {
+        sessionId = sessionCounter.incrementAndGet();
```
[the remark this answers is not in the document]
jkronegg: Fully refactored `IncrementingUuidGenerator` with the following structure:

|       40 bits      |      8 bits    |  4 bits |    12 bits    |  2 bits | 62 bits |
| -------------------| -------------- | ------- | ------------- | ------- | ------- |
| LSBs of epoch-time | sessionCounter | version | classloaderId | variant | counter |

cucumber-core/src/test/java/io/cucumber/core/runtime/UuidGeneratorServiceLoaderTest.md
jkronegg: Done

cucumber-core/src/test/java/io/cucumber/core/runtime/ObjectFactoryServiceLoaderTest.md
jkronegg: Done

cucumber-core/src/main/java/io/cucumber/core/eventbus/IncrementingUuidGenerator.java
```diff
@@ -0,0 +1,101 @@
...
+ * |       40 bits      |      8 bits    |  4 bits |    12 bits    |  2 bits | 62 bits |
+ * | -------------------| -------------- | ------- | ------------- | ------- | ------- |
+ * | LSBs of epoch-time | sessionCounter | version | classloaderId | variant | counter |
+ * <!-- @formatter:on -->
+ */
+public class IncrementingUuidGenerator implements UuidGenerator {
+    /**
+     * 40 bits mask for the epoch-time part (MSB).
+     */
+    public static final long MAX_EPOCH_TIME = 0x0ffffffffffL;
```
mpkorstanje, line 34: Can this be made private?

mpkorstanje (approved):
Looks good. I'll take care of the last few comments and merge this tomorrow in all likelihood.

cucumber-core/src/main/java/io/cucumber/core/eventbus/IncrementingUuidGenerator.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Random;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Thread-safe and collision-free UUID generator for single JVM. This is a
+ * sequence generator and each instance has its own counter. This generator is
+ * about 100 times faster than #RandomUuidGenerator. If you use Cucumber in
+ * multi-JVM setup, you should use #RandomUuidGenerator instead. Note that the
+ * UUID version and variant is not guaranteed to be stable.
```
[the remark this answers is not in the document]
jkronegg: Oops, I forgot to update the comment. The version and variant are now stable.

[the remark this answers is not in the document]
jkronegg: I the cleaned up the javadoc and made the field private.

Issue #2709, opened 2023-03-20
Corrupted Cucumber.xml when using surefire.rerunFailingTestsCount parameter


carmenloghin:
### 👓 What did you see?

Using Cucumber 7.11.0 or 7.11.1  and **surefire.rerunFailingTestsCount** parameter in order to run the UI tests pack, I noticed the report is no longer generated on Azure DevOps and the following warning message is displayed:

##[warning]Failed to read /home/vsts/work/1/s/ui/target/cucumber-xml-reports/cucumber.xml. Error : Unexpected end of file while parsing Name has occurred. Line 44, position 10..

### ✅ What did you expect to see?

With other version of Cucumber (tested with versions from 7.5 to 7.10), I saw the report generated like 

![Screenshot 2023-03-17 at 11 13 05](https://user-images.githubusercontent.com/97223571/225862418-539b2ae2-2d68-4810-b3c7-3c096ab316ab.png)

### 📦 Which tool/library version are you using?

Cucumber 7.11.0/7.11.1, Selenium 4.8, Java 11, Maven compiler 3.10, Maven Surefire Plugin 3.0.0-M7, Azure DevOps

### 🔬 How could we reproduce it?

1. Install 'Cucumber' version '7.11.0 or 7.11.1', Java version 11, Selenium version 4.8, Maven compiler version 3.10, Maven Surefire Plugin version 3.0.0-M7
2. Have a feature file (with 4 scenarios, 2 of them need to fail in order to be rerun again and the cucumber report to be generated with the new results ) which run in an Azure DevOps pipeline, on a Docker machine, with the maven command: '-Dcucumber.filter.tags="@Smoke" **-Dsurefire.rerunFailingTestsCount=1**' and test results file being created at '$(Build.SourcesDirectory)/ui/target/cucumber-xml-reports/cucumber.xml'
3. Run the pipeline and inspect the results
4. The build is green, even if there are failed tests. When looking at the results in the logs, I could see: 
[WARNING] ForkStarter IOException: Element name cannot be empty
Element name cannot be empty. See the dump file /home/vsts/work/1/s/ui/target/surefire-reports/2023-03-20T06-39-40_033-jvmRun1.dumpstream

6. See warning 
##[warning]Failed to read xxxxxx/ui/target/cucumber-xml-reports/cucumber.xml. Error : Unexpected end of file while parsing Name has occurred. Line 44, position 10..
***********************************************
This defect is reproducing on my local machine as well:
[...]

PR #2711, opened 2023-03-21
[JUnit Platform Engine] Cache parsed features

mpkorstanje:
### 🤔 What's changed?

The JUnit Platform allows Scenarios and Examples to be selected by their
unique id. This id is stable between test executions and can be used to rerun
failing Scenarios and Examples.

For practical reasons each unique id is processed individually[+] and results
in a feature file being parsed. The parsed feature is then compiled into
pickles. Both are mapped to a hierarchy of JUnit 5 test descriptors. These are
then merged. The merge process will discard any duplicate nodes.

Each time a feature file is parsed the parser will assign unique identifiers
to all ast nodes and pickles. As a result the merged hierarchy of junit test
descriptors contained pickles that belonged to a different feature file.

This poses a problem for tools such as the junit-xml-formatter that depend on
the identifiers in pickles and features to stitch everything back together. By
caching the parsed feature files we ensure that within a single execution the
internal identifiers do not change.

 + : It would actually matter little if we did process all unique ids at once,
     the problem would persist when uri selectors are used, either alone or in
     combination with unique id selectors.

Fixes: #2709

### 📋 Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- [JUnit Platform Engine] Corrupted junit-xml report when using `surefire.rerunFailingTestsCount` parameter ([#2709](https://github.com/cucumber/cucumber-jvm/pull/2709) M.P. Korstanje)
 
 ## [7.11.1] - 2023-01-27
 ### Added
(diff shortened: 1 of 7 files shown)


jkronegg:
I checked on my project (with CLI `Main` or JUnit5 platform, but without rerun of failed tests) : the feature files are parsed only once. So this caching will make no performance difference. 

I which setup do you expect to have better performance ?

mpkorstanje:
Yeah sorry, this is just a draft. I didn't get around to a full write up of the problem.

I would expect this to have a benefit when tests are selected by their uniqueId. Currently for every uniqueId we reparse the feature file.

jkronegg:
You mean that each failed test makes all features to be parsed again ? In this case, yes, caching will help.

PR #2718, opened 2023-03-30
Update README.md

atdd-bdd:
Added information on using List<Class> for a data table.

### 🏷️ What kind of change is this?

- :bank: Refactoring/debt/DX (improvement to code design, tooling, documentation etc. without changing behaviour)
- :bug: Bug fix (non-breaking change which fixes a defect)
- :zap: New feature (non-breaking change which adds new behaviour)
- :boom: Breaking change (incompatible changes to the API)

### 📋 Checklist:

- [x] I agree to respect and uphold the [Cucumber Community Code of Conduct](https://cucumber.io/conduct/)
- [ ] I've changed the behaviour of the code
  - [ ] I have added/updated tests to cover my changes.
- [ ] My change requires a change to the documentation.
  - [ ] I have updated the documentation accordingly.
- [ ] Users should know about my change
  - [ ] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../CHANGELOG.md), linking to this pull request.

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/PULL_REQUEST_TEMPLATE.md)*

PR #2720, opened 2023-03-31
Update README.md

atdd-bdd:
Altered explanation of data tables to include List<Class>.   Provided an example which follows the introductory page.

### 🤔 What's changed?
Added more detail on data tables 

Little explanation of List<Class> in the documentation. 

Documentation only

Just review and correct as desired

### DIFF
--- a/cucumber-java/README.md
+++ b/cucumber-java/README.md
@@ -66,8 +66,11 @@ Depending on the table shape, it can also be accessed as one of the following co
  * `Map<String, String> table`
  * `Map<String, List<String>> table`
  * `Map<String, Map<String, String>> table`
+ * `List<Class> table`
  
 For examples of each type see: [cucumber/datatable](https://github.com/cucumber/cucumber/tree/master/datatable)
+
+#### Data collections with String 
 
 ```java
 package com.example.app;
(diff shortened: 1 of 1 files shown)


mpkorstanje:
Hi! Thanks for looking to improve the docs. Unfortunately there already is a section about using lists of objects. It's under the "Data Table Type" heading.

Perhaps you could integrate what you're missing there. It might just need to be a reference to that section from the data tables section.

atdd-bdd:
> Hi! Thanks for looking to improve the docs. Unfortunately there already is a section about using lists of objects. It's under the "Data Table Type" heading.
> 
> Perhaps you could integrate what you're missing there. It might just need to be a reference to that section from the data tables section.

I've rearranged the entire page.  All the details about parameter, data table, and doc string transformations are all together.   
That allowed taking out the duplicate code for default transformations.

The problem with the previous arrangement was that data tables appeared in two different places in the page.   Therefore, one read the first example and didn't see the 2nd one.

mpkorstanje:
Cheers! Sorry for the delayed reaction, this is one of the more complicated PRs so it's a bit harder to find a long enough block of free spare time to review it.

First I think there is a small conceptual misunderstanding. 

Cucumber supports transforming data tables to these collections of `String` out of the box:

```
 * `List<List<String>> table`
 * `List<Map<String, String>> table`
 * `Map<String, String> table`
 * `Map<String, List<String>> table`
 * `Map<String, Map<String, String>> table`
```

(As an aside, I think `List<String>` also works, not sure why it's missing).

You may consider these views on the table.

Cucumber then also supports:
```
 * `List<T> table`
 * `List<Map<K, V>> table`
 * `Map<K, V> table`
 * `Map<K, List<V>> table`
 * `Map<K1, Map<K2, V>> table`

```

Which will transform the contents of the table into an arbitrary object if there is a data table type defined for that objects type. We can call these transforms. 

So it is not quite right to say that views on the table includes `List<Class> table` isn't quite right. It is not a view, and also not the only transform.


The second problem I've got when reading the current version is that data tables are introduced as part of the transformers. However data tables and doc strings are perfectly usable on their own and should be introduced separately from their transforms.  If it helps I would accept a version that introduce the data table and doc string  first but only reference the `DataTable` and `DocString` objects leaving out the views and transforms for the transformation section.

It might also be reasonable to split the Parameter types for Cucumber expressions, from those of DocStrings and Datatables. So the outline could roughly look like:

 * Step definitions
 * Cucumber expressions
   * What is / How to use
   * Parameter Types
 * Data Tables
   * What is / How to use
   * Transforming tables
      * The views
      * The transforms
      * Transposing tables
 * Doc Strings
   * What is / How to use
   * Transforming doc strings

atdd-bdd:
I'll work on matching your outline.  I think it makes things clearer.    

The distinction between views and transforms is subtle.  I get your point that views are different than transforms, as no string to object methods are needed.    

If there is a default transformation in place, then it no additional code is necessary to transform objects, so that it works as a view.   

You mentioned previously that you tend to use the default transformations.  I think the documentation should make it clear that doing that simplifies the code.   Creating explicit transformers could be done if more control was needed.

atdd-bdd:
I also keep confusing this PR with https://github.com/cucumber/cucumber-jvm/pull/2718. Should this one supersede the other?

That one is for a different page.   It introduces a list of objects that matches the example in the introductory page.

mpkorstanje:
Cheers! Since both influence `cucumber-java/README.md` could you merge your improvements? Makes reviewing things a bit easier.

atdd-bdd:
Now I'm a little confused.   It's not clear to me how " both influence cucumber-java/README.md could you merge your improvements".   I may have misunderstood that I was changing the same page.   But I thought I was changing two different pages.    

In what appears on:   https://cucumber.io/docs/cucumber/api/?lang=java
I think there should be an example of how to code the introductory example of what is on https://cucumber.io/docs/gherkin/reference/    At least for java.    It would not have the ten lines of default transformation code (which would be on the linked page   (cucumber-java) ).  So this would just be an addition to the Data Table entry by breaking it into two parts - Strings and objects.    Reason - one should show how to do the introductory example.


Then in what appears on:  https://github.com/cucumber/cucumber-jvm/tree/main/cucumber-java#data-tables  the page could be revised per your outline.       This is where the other variations of transformations could be shown. 
List<T> table`
 * `List<Map<K, V>> table`
 * `Map<K, V> table`
 * `Map<K, List<V>> table`
 * `Map<K1, Map<K2, V>> table`
 I don't think they appear anywhere in the documentation.

mpkorstanje:
> Now I'm a little confused. It's not clear to me how " both influence cucumber-java/README.md could you merge your improvements". I may have misunderstood that I was changing the same page. But I thought I was changing two different pages.

Ah. So both [this change](https://github.com/cucumber/cucumber-jvm/pull/2720/files#diff-5a597f444ed51f201a83c033b03653e17bce6b77761a0382e594a6c44c9a9d42) in #2720 and [this change](https://github.com/cucumber/cucumber-jvm/pull/2718/files#diff-5a597f444ed51f201a83c033b03653e17bce6b77761a0382e594a6c44c9a9d42) in #2718 go into the same file name `cucumber-java/README.md` in the `cucumber-jvm` repo. You'probably meant to change the [`content/docs/cucumber/api.md` in the `docs` repo](https://github.com/cucumber/docs/blob/main/content/docs/cucumber/api.md#data-tables).

>  I don't think they appear anywhere in the documentation.

No you are absolutely right. That part is explained a bit more in the [`datatable/README.md`](https://github.com/cucumber/cucumber-jvm/tree/main/datatable) but there is no comprehensive explanation that explains everything end-to-end. So I would very much appreciate it if you were to write it.

>  So this would just be an addition to the Data Table entry by breaking it into two parts - Strings and objects. Reason - one should show how to do the introductory example.
> ...
> Then in what appears on: https://github.com/cucumber/cucumber-jvm/tree/main/cucumber-java#data-tables the page could be revised per your outline. 

That sounds quite reasonable. 

PS: Please don't be disheartened by my irregular replies. While I can do quite a few things on my phone in an idle moment, reviewing documentation changes doesn't work so well. So for those I need to find a bigger time-slot. These are unfortunately a bit scarce.

Issue #2724, opened 2023-04-06
Performance issue in picocontainer


jkronegg:
### 👓 What did you see?

When using the picocontainer, the picocontainer container is fully reconstructed for each test-case:

```java
    public void start() {
        pico = new PicoBuilder()
                .withCaching()
                .withLifecycle()
                .build();
        for (Class<?> clazz : classes) {
            pico.addComponent(clazz);
        }
        pico.start();
    }
```

The impact is that the `PicoFactory.start()` method takes 3.2% of the total CPU time on my project with 446 test scenarios (about 170 ms).

This operation is slow because the picocontainer library inspects all classes at each container creation (see https://github.com/picocontainer/picocontainer/issues/11). 

### ✅ What did you expect to see?

I expect that picocontainer impact is not noticeable on the total CPU time.
This could be done easily by reusing the container in `PicoFactory` (the picocontainer library cannot be corrected because it is not maintained anymore since 2015).

### 📦 Which tool/library version are you using?

Cucumber JVM 7.11.2.

### 🔬 How could we reproduce it?

1. Run any cucumber test based on picocontainer with consequent number of test scenarios (>100) with the IntelliJ Profiler
2. Look the IntelliJ Profiler and look for `PicoFactory.start()` method : the flame graph tells the percentage of total CPU

### 📚 Any additional context?

The picocontainer library 2.x has not been updated since 2014 (http://picocontainer.com/). There is an unreleased picocontainer 3.x project which has no activity since 2014 as well (https://github.com/picocontainer/picocontainer).

PR #2725, opened 2023-04-06
[Pico] Improved performance

jkronegg:
### 🤔 What's changed?

Improved picontainer performance by reusing the library initialization (basically: clear the cache on `stop()`).

Solves #2724 .

The issue cannot be resolved in the picocontainer library itself (https://github.com/picocontainer/picocontainer/issues/11) because the library seems not maintained anymore.

### 📋 Checklist:

### DIFF
--- a/cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
+++ b/cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
@@ -34,19 +34,22 @@ public final class PicoFactory implements ObjectFactory {
         // do nothing (was already started in constructor)
     }
 
+    @Override
     public void stop() {
         pico.getComponentAdapters()
                 .forEach(cached -> ((Cached<?>) cached).flush());
     }
 
+    @Override
     public boolean addClass(Class<?> clazz) {
         if (isInstantiable(clazz) && classes.add(clazz)) {
+            pico.addComponent(clazz);
             addConstructorDependencies(clazz);
-            this.pico.addComponent(clazz);
         }
         return true;
     }
 
+    @Override
     public <T> T getInstance(Class<T> type) {
         return pico.getComponent(type);
     }
--- a/cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
+++ b/cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
@@ -15,7 +15,7 @@ import java.util.Set;
 public final class PicoFactory implements ObjectFactory {
 
     private final Set<Class<?>> classes = new HashSet<>();
-    private final MutablePicoContainer pico  = new PicoBuilder()
+    private final MutablePicoContainer pico = new PicoBuilder()
             .withCaching()
             .withLifecycle()
             .build();
@@ -36,7 +36,7 @@ public final class PicoFactory implements ObjectFactory {
 
     public void stop() {
         pico.getComponentAdapters()
-                .forEach(cached->((Cached<?>) cached).flush());
+                .forEach(cached -> ((Cached<?>) cached).flush());
     }
 
     public boolean addClass(Class<?> clazz) {
--- a/cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
+++ b/cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
@@ -4,6 +4,7 @@ import io.cucumber.core.backend.ObjectFactory;
 import org.apiguardian.api.API;
 import org.picocontainer.MutablePicoContainer;
 import org.picocontainer.PicoBuilder;
+import org.picocontainer.behaviors.Cached;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Modifier;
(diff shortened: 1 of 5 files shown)


cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
```diff
@@ -23,24 +31,18 @@ private static boolean isInstantiable(Class<?> clazz) {
...
-        }
-        pico.start();
+        // do nothing (was already started in constructor)
     }
 
     public void stop() {
-        pico.stop();
-        pico.dispose();
+        pico.getComponentAdapters()
+                .forEach(cached -> ((Cached<?>) cached).flush());
```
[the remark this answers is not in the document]
mpkorstanje: Just fine because `addClass` is recursive.

mpkorstanje (approved):

cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
```diff
@@ -14,7 +15,14 @@
...
 
     private final Set<Class<?>> classes = new HashSet<>();
-    private MutablePicoContainer pico;
+    private final MutablePicoContainer pico = new PicoBuilder()
+            .withCaching()
+            .withLifecycle()
+            .build();
+
+    public PicoFactory() {
+        this.pico.start();
```
mpkorstanje, line 24: This will break step definitions that implement the `Startable` because the factory is no longer stopped/started at the right moment..

jkronegg: What about doing the `pico.start()` only once in `PicoFactory.start()` ?

```java

private boolean firstTime = true;

public void start() {
    if (firstTime) {
      this.pico.start();
      firstTime = false;
    }
}
```

mpkorstanje: And it doesn't because flushing from cache also invokes the life cycle methods.

mpkorstanje (approved):

mpkorstanje:
@jkronegg I can't find flaw this with MR. But it bothers me that we can simply skip invoking the life cycle methods and still see the same effect. Could you provide a brief rationale why this works? Or if brief takes up too much time, a slightly longer explanation is fine too.

jkronegg:
> Could you provide a brief rationale why this works?

I tried to use the `stop()` and `dispose()` methods, with the hope that it will delete the instances but keep the class metadata: not only the instances are kept, but the the picocontainer stays in "stopped" or "disposed" mode and I was not able to restart it. I think this is because the pico lifecycle manager is only dispatching `stop()` and `dispose()` commands, but as they are not implemented, there is no effect. It's only on the container itself (on which the handlers are defined) that an effect can be seen.

As the picocontainer has been created "withCaching", my second trial was to look for a cache eviction method, but I could not find one.

My third approach (in this MR) is to remove the instances held by the `Cached` pico component adapters, in order to make the adapter think that the instances are not yet created. Fortunately, I found the `Cached.flush()` method and by chance it removes the instance. The `DefaultPicoContainer` has several collections that hold the pico adapters, so I empirically use the `getComponentAdapters()` method because it avoids casting. The method returns a pseudo immutable collection (the collection cannot be modified but the `Cached` objects are not themselves immutable), which explain why the component adapter cache can be flushed.

Then I ran the `PicoFactoryTest` and was happy to see that everything was already tested. Then I ran it on my usual project (which has a `TestContext` data object which is shared between 4-5 class files which define the cucumber steps) : as everything ran correctly, I supposed that the trick works properly.

I must admit : there is a part of empirical/brute force in the approach, I'm no picocontainer expert, and I was also surprised that it worked without much effort (it's only about 1 hour of work)😅.

mpkorstanje:
Mmh. I suppose because we are not using nested containers and only components, we can get away with skipping the lifecycle for the containers.

But I'm going to let this MR sit until I have some more time to look into Pico Containers features. Pico container is one of the more common DI containers for Cucumber and has been around for a long time. There is a fair chance it is subject to Hyrum's Law.

mpkorstanje:
Okay. I think this is mostly fine because

1. All added classes and their constructor dependencies are made into components
2. All components go through their life cycle hooks.
3. The internal state of the PicoFactory before and after each stop/start is the same.

One case where Hyrum might come and bite us is the order in which the components are shut down.Pico Container does this in a specific order:

```java
    private void stopAdapters() {
        for (int i = getOrderedComponentAdapters().size() - 1; 0 <= i; i--) {
            ComponentAdapter<?> adapter = getOrderedComponentAdapters().get(i);
            if (adapter instanceof ComponentLifecycle) {
                ComponentLifecycle<?> componentLifecycle = (ComponentLifecycle<?>)adapter;
                if (componentLifecycle.componentHasLifecycle() && componentLifecycle.isStarted()) {
                    componentLifecycle.stop(DefaultPicoContainer.this);
                }
            }
        }
    }
```

This order appears to be the reverse order in which components are defined. But because we define from a unordered collection this should be fine there too.

https://github.com/cucumber/cucumber-jvm/blob/b182cf876a7f44bca7aa4583deefbaa3c4658c44/cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java#L16-L34

PR #2759, opened 2023-05-19
[Java] Fix duplicate step defintion when using interfaces

### LINKED ISSUE #2757: DuplicateStepDefinition error for abstract method implementation.
**Problem**:

Whenever we have a class with Step or Hook definition/annotation, this class cannot be inherited by another but can be the last class of an hierarchy. The problem arises when we have common methods definition across this classes and no abstract method can be implemented to enforce behaviors (abstract class nor interface). i.e:

```
abstract class Parent {
    public abstract void method();
}

class Child1 extends Parent {
    @Given("Child 1")
    public void method(){
        System.out.println("1");
    }
}

class Child2 extends Parent {
    @Given("Child 2")
    public void method(){
        System.out.println("2");
    }
}
```

There is any chance to made cucumber recognize abstract definitions to avoid DuplicateStepDefinition error on implementation and use POO needed implementations such as abstraction and behavior ruling?

**More dependencies context: Maven**
```
<dependency>
    <version>7.12.0</version>
</dependency>

<dependency>
    <artifactId>cucumber-testng</artifactId>
    <version>7.12.0</version>
</dependency>

<dependency>
    <groupId>org.testng</groupId>
    <artifactId>testng</artifactId>
    <version>7.7.1</version>
</dependency>
```

**Class definitions in my scenario:**

BasePageBehaviors interface
```
public interface BasePageBehaviors<T extends BasePage<T>> {
     public T goTo();
}
```

BasePage abstract class
```
public abstract class BasePage<T extends BasePage<T>> implements BasePageBehaviors {
   public void goTo(String url){
   driver.get(url);
   }
}
```
GoogleHomePage class
```
public class GoogleHomePage extends BasePage<GoogleHomePage> {
    @Given("User is in search page")
    public GoogleHomePage goTo() {
        goTo(url.toString());
        return this;
    }
}
```

**Error:** 
```
[ERROR] Failures: 
[ERROR] org.seleniumTest.cucumber.CucumberRunnerTests.runScenario(org.seleniumTest.cucumber.CucumberRunnerTests)
[ERROR]   Run 1: CucumberRunnerTests>AbstractTestNGCucumberTests.runScenario:35 » DuplicateStepDefinition
```
```
[ERROR] runScenario(org.seleniumTest.cucumber.CucumberRunnerTests)  Time elapsed: 0.003 s  <<< FAILURE!
io.cucumber.core.runner.DuplicateStepDefinitionException: Duplicate step definitions in org.seleniumTest.pageObjects.GoogleHomePage.goTo() and org.seleniumTest.pageObjects.GoogleHomePage.goTo()
```

jkronegg:
### 🤔 What's changed?

Corrected wrong exception saying there is a duplicate step definition

Fixes #2757

Tested in TDD mode (the additional test was first red, then green after the code correction).
Tested on a real project using the snapshot version.

### 📋 Checklist:


mpkorstanje (commented):
Thanks for picking this up so quickly!

But I don't think this is the correct solution. Rather `MethodScanner.scan` should not have produced the duplicate definition.

mpkorstanje:
It should also not be possible to write:

```
interface Example {
  
   @Given("a step")
   void step();

}
```

This should result in a `InvalidMethodException`. I think it would currently merely be ignored. But I'm not sure.

jkronegg:
I chose to correct the issue in `JavaBackend.loadGlue()` to reduce the risk of side effects, but I'll move the correction to `MethodScanner.scan`.

Regarding you step definition in the interface, I don't understand the point. The unit test is only defining an interface and a class which implements this interface. 

    public interface StepsInterface {
        Number test();
    }
    public class StepsWithInterface implements StepsInterface {
        @Given("test")
        public Integer test() {
            return 1;
        }
    }

When doing `StepsWithInterface.class.getDeclaredMethods()`, it will return two methods (both behing annotated by `Given("test")`:
- Number test()
- Integer test()

mpkorstanje:
Mmh. Maybe I was a bit to quick with the second example. I'd have to look at the problem a bit more carefully I suppose. I'll do that as soon as I can find some time.

The general idea is that all step definition classes should be concrete and "effectively final".

If they're not it gets really confusing later on because people have trouble figuring out that their duplicate step definitions came from an abstract class or interface that is inherited by two other classes (novices think there is only one).

Normally extending a class with step definitions causes a validation to fail. And I would have expected that to be the case for this method inherited from the interface too.

So in my mind there are two problems, the one reported and the one where the system (wrongfully) allows this method to pass validation.

jkronegg:
> The general idea is that all step definition classes should be concrete and "effectively final".

The `StepsWithInterface` class is concrete and "effectively final" (it's a minimalistic example which reproduces the issue, you don't need to use generics and abstract classes as suggested by @sanflg in the issue). If I write another class which implements the interface and define steps, this will lead to another step definition, without ambiguity, even for beginners.

Of course it's a bit "unusual" to implement an interface when you write a class which contains cucumber step definitions, but I don't see this approach as a problem. While I don't clearly see a situation where implementing an interface gives you an advantage, there is absolutely no issue to allow this in Cucumber.

BTW, I moved the check in the `MethodScanner.scan`.

mpkorstanje:
In terms of how the `MethodScanner` should behave I agree with your description.

In terms of how I understand the implementation I would have behaved, there are a few more things I'd like to check. One being the example I gave. I'll try to find some time for that soon.

mpkorstanje:
Note to self: I was puzzled why this method appears to be volatile while volatile is not a valid method modifier. The actual type is probably bridge which has the same bit mask value.

https://stackoverflow.com/questions/6651867/why-make-a-method-volatile-in-java

jkronegg:
Good point (I'll be smarter when I get the "volatile method" question in a future job interview 😅).

Actually, I think we should use the "synthetic" flag rather than the "bridge" flag, because some synthetic methods generated by the compiler are not bridge methods. For example:

```java
public class TestSyntheticAndBridge {
	@Retention(RetentionPolicy.RUNTIME)
	static @interface Annotation {
		
	}
	
	static interface I {
		Number get();
	}
	
	static class B implements I {
		Function<String, Integer> func = s -> s.length();
		
		@Annotation
		public Integer get() {
			return 1;
		}
    }
	public static final void main(String[] args) {
		Stream.of(B.class.getDeclaredMethods()).forEach(method -> 
			System.out.println(method+
					", synthetic="+method.isSynthetic()+
					", bridge="+method.isBridge()+
					", annotations="+method.getAnnotations().length));
	}
}
```

This displays:

    public java.lang.Number TestSyntheticAndBridge$B.get(), synthetic=true, bridge=true, annotations=0
    public java.lang.Integer TestSyntheticAndBridge$B.get(), synthetic=false, bridge=false, annotations=1
    private static java.lang.Integer TestSyntheticAndBridge$B.lambda$0(java.lang.String), synthetic=true, bridge=false, annotations=0

The lambda method has has been generated for the field "func" and is synthethic but not bridge.
Note that here the method from the interface has no annotation (the test was done with Oracle JVM 10.0.1, while the tests in my previous posts were done using AzulZulu 17). I'll investigate further...

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Core] Fixed `cucumber.publish.enabled=false` ([#2747](https://github.com/cucumber/cucumber-jvm/pull/2747) M.P. Korstanje)
 - [JUnit Platform Engine] Fixed `cucumber.publish.enabled=false` ([#2747](https://github.com/cucumber/cucumber-jvm/pull/2747) M.P. Korstanje)
 
+### Added 
+- [Java] Added support for step definition classes with interfaces ([#2757](https://github.com/cucumber/cucumber-jvm/issues/2757) Julien Kronegg)
```
mpkorstanje, line 18: I would consider this a bug fix.

cucumber-java/src/main/java/io/cucumber/java/MethodScanner.java
```diff
@@ -51,8 +51,13 @@ private static boolean isInstantiable(Class<?> clazz) {
...
     private static void scan(BiConsumer<Method, Annotation> consumer, Class<?> aClass, Method method) {
-        // prevent unnecessary checking of Object methods
-        if (Object.class.equals(method.getDeclaringClass())) {
+        // prevent unnecessary checking of Object methods and generated methods
+        // Note on generated methods: when a class implements a method from the
+        // interface but specializes the return type, two Method will be
+        // generated (one with the return type of the interface and one with
+        // the specialized return type). The one with the return type of the
+        // interface is a synthetic one.
+        if (Object.class.equals(method.getDeclaringClass()) || method.isSynthetic()) {
```
mpkorstanje, line 60: Please split this condition into two statements. It will make the code comments easier to read.

mpkorstanje (commented):
Thanks for the explanation. That does cover most of my concerns and questions.

How do we know that there always is an non-synthetic annotated method? 

For example, what happens in Kotlin?

jkronegg:
> How do we know that there always is an non-synthetic annotated method?

We may may have no non-synthetic method, but that should not be an issue (in this case the `MethodScanner` will process 0 methods).

For Kotlin, I don't have enough knowledge to say...

mpkorstanje:
I did a really quick manual test. Kotlin is going to be fine. I'm not happy with doing this manually, ideally we had an example project for this but that's all the time I've got right now.

I've narrowed the solution to bridge methods only. The bug report was about a bridge method and the tests we've written are also for bridge methods. If other bugs show up involving synthetics that aren't bridge methods, we can test those individually.

mpkorstanje (approved):
@jkronegg I think this will do. But do have a look at my changes too.

jkronegg:
> @jkronegg I think this will do. But do have a look at my changes too.

That's fine for me.

PR #2772, opened 2023-06-28
[Pico] Fixed missing calls to start, stop and dispose handles

### LINKED ISSUE #2771: Pico container lifecycle changes
### 👓 What did you see?

The Pico container lifecylce changed between 7.11.2 and 7.12.1 because of this issue #2724

The dispose method is not called after each Scenario like before, but only once after the first scenario. In the following runs the state is not started anymore and the dispose method is not called anymore. Also affects the start and stop methods.

@After works like before.

The same behavior like in 7.11.2 or a hint that there was a breaking change and disposable should not be used anymore.

Java, cucumber-bom:7.12.1

See example project https://github.com/mszalbach/CucumberDisposable

Readme shows the difference state outputs when running with different versions.

Run with ./mvnw test

jkronegg:
### 🤔 What's changed?

Corrected the pico-container lifecycle: the start, stop and dispose methods are now called for every test scenarios.

The code is the same as before #2725 (so it's safer in terms of lifecycle), except that the pico container is recycled in the `start()` method when it already exists. 

When a second call to `start()` is done (e.g. for the 2nd scenario), the container component instances are all in disposed state and the container lifecycle is in disposed state. The recycling operation consists in removing all container component instances (i.e. flushing the cache) and resetting the container lifecycle.

Fixes #2771

I tested the new version on a project with ~480 test scenarios and saw no performance difference when compared to Cucumber 7.12.0.

### 📋 Checklist:

### DIFF
--- a/cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
+++ b/cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
@@ -5,6 +5,7 @@ import org.apiguardian.api.API;
 import org.picocontainer.MutablePicoContainer;
 import org.picocontainer.PicoBuilder;
 import org.picocontainer.behaviors.Cached;
+import org.picocontainer.lifecycle.DefaultLifecycleState;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Modifier;
@@ -15,14 +16,7 @@ import java.util.Set;
 public final class PicoFactory implements ObjectFactory {
 
     private final Set<Class<?>> classes = new HashSet<>();
-    private final MutablePicoContainer pico = new PicoBuilder()
-            .withCaching()
-            .withLifecycle()
-            .build();
-
-    public PicoFactory() {
-        this.pico.start();
-    }
+    private MutablePicoContainer pico;
 
     private static boolean isInstantiable(Class<?> clazz) {
         boolean isNonStaticInnerClass = !Modifier.isStatic(clazz.getModifiers()) && clazz.getEnclosingClass() != null;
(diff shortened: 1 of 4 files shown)


cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
```diff
@@ -15,36 +16,45 @@
...
+            }
+        } else {
+            // we already get a pico container which is in "disposed" lifecycle,
+            // so recycle it by defining a new lifecycle and removing all
+            // instances
+            pico.setLifecycleState(new DefaultLifecycleState());
+            pico.getComponentAdapters()
+                    .forEach(cached -> ((Cached<?>) cached).flush());
+        }
+        pico.start();
```
mpkorstanje, line 45: Performance wise, how do the different branches compare? If there still an appreciable performance gain?

jkronegg: Yes, performance is still good. JMH micro-benchmark shows:

|Benchmark |                     Mode | Cnt |      Score |      Error | Units|
|--------------|--------------------|-----|------------|---------|-----------|
|PicoFactoryBenchmark.v7_11_1 | thrpt |  25  | 24015,595 |±  1681,477  |ops/s|
|PicoFactoryBenchmark.v7_12_1|  thrpt |  25 | 769958,897 |± 65377,491 | ops/s|
|PicoFactoryBenchmark.v7_13 (this PR)    |thrpt  | 25 | 709328,227 |± 12756,281  |ops/s|

The benchmark consists in instanciating a new PicoFactory ouside the benchmark, then adding 7 classes (containing 183 step defs in total), then start the factory, get one instance and stop the factory:

    public class PicoFactoryBenchmark {
      @Benchmark
      public List<Object> v7_11_1(PicoFactoryPlan plan) {
          return benchmark(plan.picoFactory7_11_1);
      }
  
      @Benchmark
      public List<Object> v7_12_1(PicoFactoryPlan plan) {
          return benchmark(plan.picoFactory7_12_1);
      }
      @Benchmark
      public List<Object> v7_13(PicoFactoryPlan plan) {
          return benchmark(plan.picoFactory7_13);
      }
      private static List<Object> benchmark(ObjectFactory objectFactory) {
          List<Object> instances = new ArrayList<>();
          objectFactory.addClass(A.class);
          objectFactory.addClass(B.class);
          objectFactory.addClass(C.class);
          objectFactory.addClass(D.class);
          objectFactory.addClass(F.class);
          objectFactory.addClass(F.class);
          objectFactory.addClass(G.class);
  
          objectFactory.start();
          instances.add(objectFactory.getInstance(A.class));
          objectFactory.stop();
  
          return instances;
      }
  
    }

I think the version from Cucumber 7.12.1 is about 10% faster than the new code for 7.13 because the pico-container is started in the constructor (that is, outside the benchmark). But this code for 7.13 is still 30 times faster than in Cucumber 7.11.1.

Issue #2052, opened 2020-07-17
Cucumber plugin for setting up a context before every test step


agolubev81:
Hello, can Cucumber plugins be used for setting some context before actual step is executed? 

For instance, I have a requirement to integrate tests with 3rdparty framework which is responsible for logging and error handling. Ideally all test-related information should be available to that integration code, which can be implemented as AOP btw, including text with test step. Will plugins allow to do that ? 

My concern is that plugins API is event-based, unlike hooks, and thus order of test start and end events might be not exactly "around" actual test execution, which is important for managing context.

I'm aware about existing tickets https://github.com/cucumber/cucumber-jvm/issues/1805 for adding step information into hooks, which seems quite straight forward approach for my case, but it's not implemented yet and on other hand plugins  are proposed as alternative.

mpkorstanje:
In brief, no. events are not guaranteed to be provided to the plugin around the invocation of a step. For example when implementing the `EventListener` interface and executing scenarios in parallel, all events will be offered in canonical order to the plugin after test execution has completed. The `ConcurrentEventListener` does not suffer from this problem but also does not guarantee that you'll be able to use `ThreadLocal` to share data between your step execution and the plugin.

Now at a glance it sounds like you already have solution in mind and are looking for help implementing that solution. However I also get the impression that you have not explained your actual problem. This phenomenon is called the [XY-Problem](http://xyproblem.info/).

So which framework are you using? Why do you need invoke around guarantees from the plugin? What problem would these solve?

> I'm aware about existing tickets #1805 for adding step information into hooks, which seems quite straight forward approach for my case, but it's not implemented yet and on other hand plugins are proposed as alternative.

It's open source. If nobody cares enough to implement it, it doesn't get done. Are you looking to submit a pull request?

agolubev81:
> So which framework are you using? Why do you need invoke around guarantees from the plugin? What problem would these solve? What data are you trying to share between the step and plugin and why?
-  The framework is proprietary
-  Logging and error handling are to be delegated to the framework 
-  Integration is a must requirement (problem yeah :)
-  Info about step definition and java method should be available. 

Let's look at the example (it's a not a real but conceptually very close):
```
@When("example step is executed with parameter {string}")
public void example(String parameter) {
  TestFramework.logTestStart("test step: {} , method : {}", 
    String.format("example step is executed with parameter %s", parameter) , 
    "example()");
  try {
// actual test code 
  } catch (Throwable th) {
[...]

PR #2789, opened 2023-09-03
feat: Add Plugin for OpenTelemetry tracing support

deejgregor:
### 🤔 What's changed?

This adds a new plugin that integrates OpenTelemetry tracing with Cucumber. A new trace is started for each test case and each test step is a span underneath.

This is most useful when downstream code is also OpenTelemetry-tracing enabled, as it provides a very clear linkage between cucumber test case/steps and application interaction. E.g.: 
image

See previous discussion in Slack: https://cucumberbdd.slack.com/archives/C5YHPPJMP/p1688522812400979

OpenTelemetry uses a ThreadLocal to pass span context information downstream. If I start a span in the plugin with TestCaseStarted/TestStepStarted events (note: it is the Scope that updates the ThreadLocal) and end the span on TestCaseFinished/TestStepFinished events, this seems to work properly. I've seen in https://github.com/cucumber/cucumber-jvm/issues/2052#issuecomment-659966397, that this isn't guaranteed, however:
> The ConcurrentEventListener does not suffer from this problem but also does not guarantee that you'll be able to use ThreadLocal to share data between your step execution and the plugin.

So, I'm curious if this approach is correct at all (or correct enough for this usage), or if something else should be done.

The other main question I have: because I need to find the same Span/Scope objects that were created in the Test*Started events later when the Test*Finished events come in, I keep a Map containing these and it is keyed on the TestCase ID and codeLocation, which seems to be unique in my testing--but I'm not sure if it is unique enough. Also, I realized that I can get the Span by calling Span.current() (as long as I see this in the same thread), and then I can use that to lookup the Scope with a simple map and then eliminate the need for the Pair class altogether. Unless there seem to be any issues with this approach, I will plan on making this change, which should nicely simplify things a good bit.

Do the names of the Maven module, Java package, and Java class for the plugin sound good?

One thing that would help related to this would be to include the trace ID for each trace in the scenario log output. I couldn't find a way to do this within the plugin, but I could do it with a hook:

```
    @Before
    public void logTraceId(Scenario scenario) {
        Span current = Span.current();
        if (current.getSpanContext().isValid()) {
            String link = "http://localhost:16686/trace/" + current.getSpanContext().getTraceId();
            scenario.log("trace ID: " + current.getSpanContext().getTraceId() + " -- " + link);
        }
    }
```

If we were able to find a good way to include something like this, the link pattern that is hardcoded to "http://localhost:16686/trace/..."  would be configurable.

I also have a tweak to PrettyFormatter that will add clickable hyperlinks for terminals that support it (like iTerm on macOS). See the commit on this branch: https://github.com/deejgregor/cucumber-jvm/commit/63ae9f20ce1408f1fc3f3e3d14b8b92d49528d00.

The combination of the plugin, the hook that logs the trace ID and link, and the tweak to the PrettyFormatter make it quite nice and easy to see the tracing details as a test is running.

### 📋 Checklist:

### Quick documentation

This will be moved into formal documentation as the approach solidifies.

#### Trying it out

The easy way to try this out, assuming you have Docker and some application you can try this on (something that is client-server based can be particularly cool to try, as long as it uses common libraries that are auto-instrumented by the OpenTelemetry Java Agent):
1. Run the Jaeger all-in-one Docker image: https://www.jaegertracing.io/docs/latest/getting-started/#all-in-one
2. Download the OpenTelemetry Java Agent: https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/latest
3. Set some environment variables (see below).
4. Build and install cucumber-jvm from this branch.
5. Run your tests with cucumber-jvm from this branch and the OpenTelemetry Java Agent by adding `-javaagent:<path-to-java-agent>/opentelemetry-javaagent.jar` to the JVM command line. Also make sure to active the cucumber-opentelemetry plugin with `--plugin io.cucumber.opentelemetry.OpenTelemetryTracingEventListener`.
6. Visit Jaeger at http://localhost: 16686/ to see the trace information.

Environment variables for item 3 above:
```
export OTEL_TRACES_EXPORTER=otlp # otlp is the default, but including here to be explicit
export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://localhost:4317 # this will be sent to Jaeger
export OTEL_METRICS_EXPORTER=none # Jaeger only supports metrics
export OTEL_LOGS_EXPORTER=none  # Jaeger only supports metrics
export OTEL_SERVICE_NAME=<some descriptive name>
```
Note: you can also use `OTEL_TRACES_EXPORTER=logging` to see the trace information on stdout (or `otlp,logging` to send to both).

Once the trace data starts getting sent, you should see the service name you set in `OTEL_SERVICE_NAME` show up in Jaeger, and then you can select the service and click "Find Traces" to see traces available. You should see one trace for each test case executed. Note: tracing data is sent when a span (in this case each test case or test step), so you won't see it until each of them *ends*. But even if a test case isn't yet finished, you should see each test step in the trace as it finishes. Test steps/cases that fail will be annotated with a red exclamation mark in the Jaeger UI and any exceptions received will be shown in the "Logs" section of a cucumber-genereated span. You can find a good amount of information pulled from the test cases/steps in the "Tags" section of each cucumber-generated span.

#### Random notes I don't want to forget

You don't need the OpenTelemetry Java Agent to use this. You can also use [manual instrumentation](https://opentelemetry.io/docs/instrumentation/java/manual/), as well. The OpenTelemetry SDK would need to be setup before this plugin gets called, and can be done with a `@BeforeAll` hook:
```
    @BeforeAll
    public static void autoConfigureOpenTelemetrySdk() {
        AutoConfiguredOpenTelemetrySdk.builder()
                        .setResultAsGlobal(true)
                        .build();
    }
```

Also, if you don't have the OpenTelemetry SDK initialized, all of this is just a no-op (this is a key part of the design of the OpenTelemetry API--that it can be safely used if the SDK isn't present, so it can be integrated widely).

----

### DIFF
--- a/cucumber-opentelemetry/src/main/java/io/cucumber/opentelemetry/OpenTelemetryTracingEventListener.java
+++ b/cucumber-opentelemetry/src/main/java/io/cucumber/opentelemetry/OpenTelemetryTracingEventListener.java
@@ -1,0 +1,289 @@
+package io.cucumber.opentelemetry;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.UUID;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import io.cucumber.plugin.ConcurrentEventListener;
+import io.cucumber.plugin.event.EventPublisher;
+import io.cucumber.plugin.event.HookTestStep;
+import io.cucumber.plugin.event.PickleStepTestStep;
+import io.cucumber.plugin.event.Result;
+import io.cucumber.plugin.event.Status;
+import io.cucumber.plugin.event.TestCaseEvent;
+import io.cucumber.plugin.event.TestCaseFinished;
+import io.cucumber.plugin.event.TestCaseStarted;
+import io.cucumber.plugin.event.TestStep;
+import io.cucumber.plugin.event.TestStepFinished;
+import io.cucumber.plugin.event.TestStepStarted;
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanBuilder;
+import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.context.Scope;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+
+public class OpenTelemetryTracingEventListener implements ConcurrentEventListener {
+    // FUTURE: Make this configurable
+    public static final boolean START_NEW_TRACE_FOR_EACH_TEST_CASE = true;
+
+    public static final boolean DEBUG_SPAN_SCOPE = false;
+
+    // Match code locations of the form: <package>.<class>.<method>(<args>)
+    // $1 = package, $2 = class, $3 = method, $4 = args (if any)
+    public static final Pattern CODE_LOCATION_CLASS_METHOD_PATTERN = Pattern
+            .compile("^([^(]*)\\.([^.(]*)\\.([^.(]*)\\((.*)\\)$");
+
+    // OpenTelemetry attribute keys that are not in the standard trace semantic
+    // conventions.
+    public static final AttributeKey<String> ATTRIBUTE_KEY_CUCUMBER_EVENT = AttributeKey.stringKey("cucumber.event");
+    public static final AttributeKey<String> ATTRIBUTE_KEY_CUCUMBER_STATUS = AttributeKey.stringKey("cucumber.status");
+    public static final AttributeKey<List<String>> ATTRIBUTE_KEY_CODE_FUNCTION_ARGS = AttributeKey
+            .stringArrayKey("code.functionargs");
+    public static final AttributeKey<String> ATTRIBUTE_KEY_CODE_LOCATION = AttributeKey.stringKey("code.location");
+    public static final AttributeKey<List<String>> ATTRIBUTE_KEY_SOURCE_TAGS = AttributeKey
+            .stringArrayKey("source.tags");
+    public static final AttributeKey<String> ATTRIBUTE_KEY_CASE_KEYWORD = AttributeKey.stringKey("testcase.keyword");
+    public static final AttributeKey<String> ATTRIBUTE_KEY_CASE_NAME = AttributeKey.stringKey("testcase.name");
+    public static final AttributeKey<String> ATTRIBUTE_KEY_CASE_ID = AttributeKey.stringKey("testcase.id");
+    public static final AttributeKey<String> ATTRIBUTE_KEY_STEP_TYPE = AttributeKey.stringKey("step.type");
+    public static final AttributeKey<String> ATTRIBUTE_KEY_STEP_PATTERN = AttributeKey.stringKey("step.pattern");
+    public static final AttributeKey<String> ATTRIBUTE_KEY_STEP_KEYWORD = AttributeKey.stringKey("step.keyword");
+    public static final AttributeKey<String> ATTRIBUTE_KEY_STEP_TEXT = AttributeKey.stringKey("step.text");
+    public static final AttributeKey<String> ATTRIBUTE_KEY_STEP_ARGUMENT = AttributeKey.stringKey("step.argument");
+    public static final AttributeKey<String> ATTRIBUTE_KEY_HOOK_TYPE = AttributeKey.stringKey("hook.type");
+
+    private Tracer tracer = null; // do not use directly; call getTracer
+
+    // Multiple steps might be executing in parallel, so we might have multiple
+    // Spans/Scopes open at once. This is where we track each individually.
+    // XXX we might be able to simplify this to Map<Span, Scope> and get the
+    // current span with Span.current()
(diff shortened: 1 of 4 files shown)


SButterfly:
Hello. I occasionally found your MR!
My team is also interested in this plugin. However, we are integrated with GCP and use `com.google.cloud:spring-cloud-gcp-starter-trace` library to inject tracing in all our services.
Is there an option to unify spring sleuth & open telemetry?

cucumber-opentelemetry/src/main/java/io/cucumber/opentelemetry/OpenTelemetryTracingEventListener.java
```diff
@@ -0,0 +1,289 @@
...
+        } else if (result.getStatus() == Status.PASSED) {
+            span.setStatus(StatusCode.OK);
+        }
+
+        final Pair<Span, Scope> spanScope = testCaseToSpanScope
+                .remove(Pair.of(event.getTestCase().getId(), codeLocation));
+        if (spanScope != null) {
+            if (DEBUG_SPAN_SCOPE) {
+                if (!span.equals(spanScope.getKey())) {
+                    System.err.println(getClass().getSimpleName() + ": " +
```
jkronegg, line 226: Isn't ”System.err" a bit old fashioned? Maybe standard logging would be more appropriate.

deejgregor: Oops, yes, very much! And I'm clearly old fashioned in this regard--at least in code I'm not familiar with. ;-) I'll take care of that.

jkronegg:
Should really the plugin be a submodule of cucumber-jvm ? Because if we multiply the number of plugins, it would be difficult to differentiate which module is mandatory and which one is optional. At least, the module name should contain a "plugin" qualifier...

Also having a plugin as submodule forces the community to maintain it at each build, even if the plugin usage is low. As Smartbear stopped to pay Cucumber developpers, I don't think this is a good idea to increase the codebase with a plugin.

mpkorstanje:
As I said on Slack I’d rather not add any OTEL dependency to Cucumber. As a test framework we can’t afford a dependency tree, it could interfere with the system under test.

Using a separate module does address this somewhat. But it still requires knowledge extensive knowledge. And this isn't scalable. Hence the request to create an issue to discuss the current short comings of the plugin system.

With that in mind, I don't think there is much of a point in discussing the finer details of this plugin.

I do regret that I've not been able to get around to this yet. But so it is.

deejgregor:
>  Hence the request to create an issue to discuss the current short comings of the plugin system.

I'll open up a separate issue for plugin system around logging.

> With that in mind, I don't think there is much of a point in discussing the finer details of this plugin.
> 
> I do regret that I've not been able to get around to this yet. But so it is.

+1. You definitely have other priorities, and that's quite understandable. Thanks for your time in the past and the time you've been able to give to my questions, too.

> That said, it seems there is an interest in this plugin. I would encourage you to publish it on your own. It's a bit of paper work to get started, but nothing too onerous.

I'll close this issue and work on it separately.

deejgregor:
@jkronegg said:
> Should really the plugin be a submodule of cucumber-jvm ? Because if we multiply the number of plugins, it would be difficult to differentiate which module is mandatory and which one is optional. At least, the module name should contain a "plugin" qualifier...
> 
> Also having a plugin as submodule forces the community to maintain it at each build, even if the plugin usage is low. As Smartbear stopped to pay Cucumber developpers, I don't think this is a good idea to increase the codebase with a plugin.

My reasoning behind suggesting this was because OpenTelemetry support is starting to get much more common (particularly tracing), and I'm starting to see more libraries, frameworks, and applications add direct support for it. The API layer (which is all that would be linked here) is small (2 JARs, < 200 kB), stable, and no-op by default (implementations are delivered in separate, optional SDKs).  For more information see 1 and 2 [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/library-guidelines.md#requirements). I now realize that some of these details were only in the Slack thread, but not included in this PR (until now).

Based on feedback, I've closed this PR, and I'll look at releasing it separately and also file an issue about adding some logging capability to the plugin interface.

Thanks for your time!

deejgregor:
> Hello. I occasionally found your MR! My team is also interested in this plugin. However, we are integrated with GCP and use `com.google.cloud:spring-cloud-gcp-starter-trace` library to inject tracing in all our services. Is there an option to unify spring sleuth & open telemetry?

@SButterfly I found https://cloud.spring.io/spring-cloud-static/spring-cloud-gcp/1.1.0.RC2/multi/multi__spring_cloud_sleuth.html and based on that, it looks like this uses Spring Cloud Sleuth which is an older Spring library that uses OpenZipkin Brave to generate traces. Spring Cloud Sleuth has moved to Micrometer Tracing, and there is OpenTelemetry support for that: https://micrometer.io/docs/tracing

But I don't think this plugin will work for you as-is with the tracing infrastructure you have. You could probably make a version of the plugin that worked with Spring Cloud Sleuth and/or Brave, however.

murdos:
@deejgregor : Have you released this code as a library? 
I just tried the code from this PR, and the result is really interesting! Thanks!

deejgregor:
@murdos I have not yet had the chance--new job and we don't use the same test infrastructure. Plus, I'm new to publishing projects to Maven, so there's a little bit of a learning process for me to get over there. Feel free to take this and run with it, or if you want to work on it together, that'll help get it over the finish line, I think.

PR #2822, opened 2023-11-29
Fix test failure by comparing json instead of strings

RugvedB:

Created this PR to fix 1 flaky test - [test](https://github.com/cucumber/cucumber-jvm/blob/main/cucumber-core/src/test/java/io/cucumber/core/plugin/MessageFormatterTest.java#L22).

### 📋 Checklist:

----
### Further Details:
1. **How were these test identified as flaky?**
This test was identified as flaky by using an open-source research tool named [NonDex](https://github.com/TestingResearchIllinois/NonDex) which is responsible for finding and diagnosing non-deterministic runtime exceptions in Java programs.

2. **What was the error?**
When the test convert bytes to json string `String ndjson = new String(bytes.toByteArray(), UTF_8);`, the order of keys (i.e. seconds, nano, testRunStarted, testRunFinished, etc) is not guaranteed. Thus, when we compare this String variable 'ndjson' directly to hardcoded string `"{\"testRunStarted\":{\"timestamp\":{\"seconds\":10,\"nanos\":0}}}\n{\"testRunFinished\":{\"success\":true,\"timestamp\":{\"seconds\":15,\"nanos\":0}}}\n"`, it may fail sometimes because the order could be different than the expected order.
```
[ERROR] Failures: 
[ERROR]   MessageFormatterTest.test:35 
Expected: a string containing "{\"testRunStarted\":{\"timestamp\":{\"seconds\":10,\"nanos\":0}}}\n{\"testRunFinished\":{\"success\":true,\"timestamp\":{\"seconds\":15,\"nanos\":0}}}\n"
     but: was "{"testRunStarted":{"timestamp":{"nanos":0,"seconds":10}}}
{"testRunFinished":{"timestamp":{"nanos":0,"seconds":15},"success":true}}
"
```
3. **What is the fix?**
```
assertThat(expectedMap.entrySet(), everyItem(is(in(ndjsonMap.entrySet()))));
```
Instead of comparing json strings, we can convert json to Map and then compare the maps. This PR proposes to compare Maps instead of json strings in test. In the above code snippet, the assertion checks that every entry in expectedMap is present in ndjsonMap. If this assertion passes for every entry in expectedMap, it means that the key-value pairs in expectedMap are all present in ndjsonMap.

--------------
You can run the following commands to run the test using NonDex tool:
```
mvn -pl cucumber-core edu.illinois:nondex-maven-plugin:2.1.1:nondex -Dtest=io.cucumber.core.plugin.MessageFormatterTest#test
```

Test Environment:
```
java version 11.0.19
Apache Maven 3.9.5
```
Kindly let me know if this fix is acceptable.
Thank you

### DIFF
--- a/cucumber-core/src/test/java/io/cucumber/core/plugin/MessageFormatterTest.java
+++ b/cucumber-core/src/test/java/io/cucumber/core/plugin/MessageFormatterTest.java
@@ -1,5 +1,6 @@
 package io.cucumber.core.plugin;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import io.cucumber.core.eventbus.EventBus;
 import io.cucumber.core.runtime.TimeServiceEventBus;
 import io.cucumber.messages.types.Envelope;
@@ -9,12 +10,18 @@ import io.cucumber.messages.types.Timestamp;
 import org.junit.jupiter.api.Test;
 
 import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.time.Clock;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.UUID;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
-import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.everyItem;
+import static org.hamcrest.Matchers.in;
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.fail;
 
 public class MessageFormatterTest {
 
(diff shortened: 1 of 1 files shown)


cucumber-core/src/test/java/io/cucumber/core/plugin/MessageFormatterTest.java
```diff
@@ -32,9 +39,21 @@ void test() {
         bus.send(Envelope.of(testRunFinished));
 
         String ndjson = new String(bytes.toByteArray(), UTF_8);
-        assertThat(ndjson, containsString("" +
-                "{\"testRunStarted\":{\"timestamp\":{\"seconds\":10,\"nanos\":0}}}\n" +
-                "{\"testRunFinished\":{\"success\":true,\"timestamp\":{\"seconds\":15,\"nanos\":0}}}\n"));
+        String expectedJson = "{\"testRunStarted\":{\"timestamp\":{\"seconds\":10,\"nanos\":0}},\"testRunFinished\":{\"success\":true,\"timestamp\":{\"seconds\":15,\"nanos\":0}}}";
```
mpkorstanje, line 42: This meaningfully changes what is being tested. Json and Ndjson are distinctly different. See https://ndjson.org/

cucumber-core/src/test/java/io/cucumber/core/plugin/MessageFormatterTest.java
```diff
@@ -32,9 +39,21 @@ void test() {
...
+        try {
+            Map<String, Object> expectedMap = objectMapper.readValue(expectedJson, Map.class);
+
+            String[] ndjsonObjects = ndjson.split("\\n");
+            Map<String, Object> ndjsonMap = new HashMap<>();
+            for (String ndjsonObject : ndjsonObjects) {
+                ndjsonMap.putAll(objectMapper.readValue(ndjsonObject, Map.class));
+            }
+
+            assertThat(expectedMap.entrySet(), everyItem(is(in(ndjsonMap.entrySet()))));
```
mpkorstanje, line 53: Have you considered using `JSONAssert.assertEquals`?

mpkorstanje (changes requested):
Hi, thanks for discovering and offering to fix this.

The suggested fix is more complicated than needed and also incorrect. 

As a general pointer I would suggest declaring the expected ndjson as an array of strings, and splitting the actual ndjson into an array of strings. You can then compare the elements of this array using `JSONAssert.assertEquals`.

Unfortunately I do not have more time available to coach you on a better solution.

mpkorstanje:
That looks reasonable but please provide your suggestions as a commit. Note that to improve readability you can use shortcut syntax to create and initialize an array for the expected ndjson. 

https://docs.oracle.com/javase/tutorial/java/nutsandbolts/arrays.html

RugvedB:
> That looks reasonable but please provide your suggestions as a commit. Note that to improve readability you can use shortcut syntax to create and initialize an array for the expected ndjson.
> 
> https://docs.oracle.com/javase/tutorial/java/nutsandbolts/arrays.html

@mpkorstanje Thanks you for your guidance. I have pushed the changes. Feel free to let me know if further steps need to be taken to make it perfect. Also, let me know if you want me to squash the commits into 1 commit. Thanks for your time.

mpkorstanje (approved):
Cheers! This looks good. I've pushed a few cosmetic changes on your branch. 

It's not necessary to to put the name of the type into a variable. The type is right in front of it. Also I used static imports to keep the code a bit more succinct.

Squashing won't be necessary. It's the default mode for merging in this Git repo.

PR #2835, opened 2023-12-16
[JUnit Platform] Use EngineDiscoveryRequestResolver

mpkorstanje:

- :bank: Refactoring/debt/DX (improvement to code design, tooling, documentation etc. without changing behaviour)

### 📋 Checklist:


mpkorstanje:
The current implementation of the JUnit Platform Engine treats `@SelectClasspathResource` aimed at directories as if all resources in that directory should be considered. This is not the intended usage, for this `@SelectPackages` should be used.

Cucumbers documentation consistently uses `@SelectClasspathResource`.


This will become a problem if/when we start to use the `EngineDiscoveryRequestResolver` with (#2835, https://github.com/junit-team/junit5/pull/3705, https://github.com/junit-team/junit5/pull/3718) because `@SelectClasspathResource` will cause an exception to be thrown when used with a non-existing resource.

cucumber-junit-platform-engine/src/main/java/io/cucumber/junit/platform/engine/FeaturesPropertyResolver.java
```diff
@@ -0,0 +1,107 @@
...
+            return;
+        }
+        warnedWhenCucumberFeaturesPropertyIsUsed = true;
+        log.warn(
+            () -> "Discovering tests using the " + FEATURES_PROPERTY_NAME + " property. Other discovery " +
+                    "selectors are ignored!\n" +
+                    "\n" +
+                    "This is a work around for the limited JUnit 5 support in Maven and Gradle. " +
+                    "Please request/upvote/sponsor/ect better support for JUnit 5 discovery selectors. " +
+                    "For details see: https://github.com/cucumber/cucumber-jvm/pull/2498\n" +
```
marcphilipp, line 71: This should also be interesting for you (if/when it gets merged):
https://github.com/gradle/gradle/pull/29535

cucumber-junit-platform-engine/src/main/java/io/cucumber/junit/platform/engine/IsFeature.java
```diff
@@ -0,0 +1,12 @@
+package io.cucumber.junit.platform.engine;
+
+import org.junit.platform.commons.support.Resource;
+
+import java.util.function.Predicate;
+
+class IsFeature implements Predicate<Resource> {
+    @Override
+    public boolean test(Resource resource) {
+        return resource.getUri().toString().endsWith(".feature");
```
marcphilipp, line 10: Wouldn't checking `Resource.getName()` suffice here?

cucumber-junit-platform-engine/src/main/java/io/cucumber/junit/platform/engine/FeatureElementOrderingVisitor.java
```diff
@@ -0,0 +1,23 @@
...
+    @Override
+    public void visit(TestDescriptor descriptor) {
+        // TODO: Make order configurable
+        List<? extends TestDescriptor> sortedChildren = descriptor.getChildren().stream()
+                .filter(FeatureElementDescriptor.class::isInstance)
+                .map(FeatureElementDescriptor.class::cast)
+                .sorted(comparing(FeatureElementDescriptor::getLine))
+                .collect(Collectors.toList());
+        sortedChildren.forEach(descriptor::removeChild);
+        sortedChildren.forEach(descriptor::addChild);
```
marcphilipp, line 21: I wonder if we should introduce an `TestDescriptor.orderChildren(UnaryOperator<TestDescriptor> orderer)` method (or similar). We could also make use of that in Jupiter. :thinking:

marcphilipp (commented):
Great to see this being used here now! :+1:

cucumber-junit-platform-engine/src/main/java/io/cucumber/junit/platform/engine/FeatureElementOrderingVisitor.java
mpkorstanje: That would be most useful! Though I'm not sure I understand the signature. Were you thinking of `Comparator<TestDescriptor>`?

cucumber-junit-platform-engine/src/main/java/io/cucumber/junit/platform/engine/IsFeature.java
mpkorstanje: Good one. Cheers.

cucumber-junit-platform-engine/src/main/java/io/cucumber/junit/platform/engine/FeatureElementOrderingVisitor.java
marcphilipp: Sorry, I meant `UnaryOperator<List<TestDescriptor>>`. While a `Comparator` makes sense as well, in some cases (think `Collections.shuffle`) it's easier to return a new list. The implementation should verify that no elements were added or removed.

cucumber-junit-platform-engine/src/main/java/io/cucumber/junit/platform/engine/Constants.java
```diff
@@ -261,6 +261,11 @@ public final class Constants {
      */
     public static final String EXECUTION_MODE_FEATURE_PROPERTY_NAME = "cucumber.execution.execution-mode.feature";
 
+    /**
+     * TODO: Document
```
mpkorstanje, line 265: Do this.

Issue #2886, opened 2024-05-22
Aspectj annotation is being ignored while using Spring and Cucumber v7.10.0 onwords


Hakky54:
### 👓 What did you see?

I am using Cucumber alongside with Spring ans Aspectj. This setup always worked. I was on version `7.2.3` of cucumber. I tried bumping it to 7.18.0 however some part of my tests failed. It was not picking up the aspectj annotations of the before and after step. I am not qure what is causing the issue. However If I downgrade to version 7.9.0 it still works. If I bump it to 7.10.0 it fails.

### ✅ What did you expect to see?

I would expect my existing test suite still be working

### 📦 Which tool/library version are you using?

JDK 17
Maven

### 🔬 How could we reproduce it?

1. Git clone https://github.com/Hakky54/mutual-tls-ssl.git
2. Adjust pom.xml in the root directory of the project. Change version of cucumber to 7.10.0
3. run `mvn clean verify` it will throw nullpointer exception

The before and after call in LogExecutionTimeAspect.java is being ignored somehow while it is still working on version 7.9.0

mpkorstanje:
Interesting. Haven't quite got the time to dig into a bug project like this, but could you put `@CucumberContextConfiguration` on a separate class?

Hakky54:
I have tried that, but then I end up getting the well known error messsage:

```
io.cucumber.core.backend.CucumberBackendException: Please annotate a glue class with some context configuration.

For example:

   @CucumberContextConfiguration
   @SpringBootTest(classes = TestConfig.class)
   public class CucumberSpringConfiguration { }
Or: 

   @CucumberContextConfiguration
   @ContextConfiguration( ... )
   public class CucumberSpringConfiguration { }

	at io.cucumber.spring.SpringFactory.start(SpringFactory.java:100)
	at io.cucumber.core.runner.Runner.buildBackendWorlds(Runner.java:134)
	at io.cucumber.core.runner.Runner.runPickle(Runner.java:70)
	at io.cucumber.junit.PickleRunners$NoStepDescriptions.lambda$run$0(PickleRunners.java:151)
	at io.cucumber.core.runtime.CucumberExecutionContext.lambda$runTestCase$5(CucumberExecutionContext.java:129)
	at io.cucumber.core.runtime.RethrowingThrowableCollector.executeAndThrow(RethrowingThrowableCollector.java:23)
[...]

PR #2887, opened 2024-05-24
[Spring] Document CucumberContextConfiguration semantics

mpkorstanje:

Cucumber uses Spring Test Context Manager framework. This framework was written for JUnit and assumes that there is a "test instance". Cucumber however uses multiple step definition classes and so it has multiple test instances.

Originally `@CucumberContextConfiguration` was added to signal to Spring which class should be used to configure the application context from. But as people also expected mock beans and other features provided by Springs test execution listeners to work (#2661) the annotated instance was only instantiated but never initialized by Spring.

This changed the semantics somewhat as now features that depend on the bean being initialized stopped working (#2886). Unfortunately, there is little that can be done here. Spring expects that the instance provided to the Test Context Manager to be an uninitialized bean. The solution for this is to put the context configuration and step definitions in different classes.

Cleaning up the examples to follow this pattern should avoid this problem somewhat in the future. Though I won't go as far as recommending people do this. Putting everything in one class looks quite nice. And generally still works.

Closes: #2886

### 📋 Checklist:

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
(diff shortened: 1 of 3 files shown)


Hakky54 (approved):

mpkorstanje:
Todo: Extend example to use `@TestConfiguration` as an inner class of the Cucumber context configuration.

PR #2902, opened 2024-07-05
Faster caching glue

mpkorstanje:
### 🤔 What's changed?

Work in progress.  Parking this in a pull request so I won't forget about it.

Currently the `CachingGlue` is rather limited in what it can cache. Step definitions are derived from parameter types, and because these can be `Disposable`, this means that we dispose of the whole cache after each scenario.

But glue can be reused between scenarios if 
 * The step definitions are non-disposable.
 * The step definitions are the same as the previous scenario
 * The the language of the scenario is the same.
 
  
  
- :boom: Breaking change (incompatible changes to the API)

### 📋 Checklist:

----

### DIFF
--- a/cucumber-core/src/main/java/io/cucumber/core/runner/CachingGlue.java
+++ b/cucumber-core/src/main/java/io/cucumber/core/runner/CachingGlue.java
@@ -44,6 +44,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.TreeMap;
 
@@ -81,6 +82,7 @@ final class CachingGlue implements Glue {
     private final Map<String, CoreStepDefinition> stepDefinitionsByPattern = new TreeMap<>();
 
     private final EventBus bus;
+    private StepTypeRegistry stepTypeRegistry;
 
     CachingGlue(EventBus bus) {
         this.bus = bus;
@@ -228,7 +230,12 @@ final class CachingGlue implements Glue {
         return docStringTypeDefinitions;
     }
 
-    void prepareGlue(StepTypeRegistry stepTypeRegistry) throws DuplicateStepDefinitionException {
+    StepTypeRegistry getStepTypeRegistry() {
+        return null;
+    }
+
+    void prepareGlue(Locale locale) throws DuplicateStepDefinitionException {
+        stepTypeRegistry = new StepTypeRegistry(locale);
         StepExpressionFactory stepExpressionFactory = new StepExpressionFactory(stepTypeRegistry, bus);
 
         // TODO: separate prepared and unprepared glue into different classes
@@ -425,7 +432,6 @@ final class CachingGlue implements Glue {
     }
 
     void removeScenarioScopedGlue() {
-        stepDefinitionsByPattern.clear();
         removeScenarioScopedGlue(beforeHooks);
         removeScenarioScopedGlue(beforeStepHooks);
         removeScenarioScopedGlue(afterHooks);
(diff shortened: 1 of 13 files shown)


jkronegg:
That would be great! (`CachingGlue.prepareGlue()` lasts 50% of the total test duration on my usual test project, ~600 test scenarios)

jkronegg:
Shouldn't the glue be reusable also if 
- the parameterTypeDefinitions do not change 
- the docStringTypeDefinitions do not change
- the dataTableTypeDefinitions do not change ?

I pushed the branch [`faster-caching-glue-julien`](https://github.com/cucumber/cucumber-jvm/tree/faster-caching-glue-julien) to tentatively implement the rules you proposed, but I ran into an issue: if the step definitions are cached, the StepDefined event is not emitted as duplicate anymore, making one test to fail, `io.cucumber.core.runtime.RuntimeTest#generates_events_for_glue_and_scenario_scoped_glue` (which expects multiple StepDefined events to be thrown). I'm not sure if the test-case can be changed... (is this an API change?)

I have a Cucumber plugin which gets a lot of StepDefined events: I need to remove the duplicates, so it would not be an issue when the duplicate events are not emitted.

When changing this test-case to await only 2 StepDefined events instead of 4, the build passes and on my usual test project (~600 test scenarios, 230 step definitions), I get incredible results:

| version | `CachingGlue.prepareGlue()` impact on test project (IntelliJ profiler) | total Cucumber test duration on test project |
|---------|---------------------------------------------------------|--------------------------------------------|
| 7.20.1 | 54% of execution time | 6.1 seconds |
| `faster-caching-glue-julien` | 0.5% of execution time | 3.3 seconds |

More or less twice faster... 😁

mpkorstanje:
Unfortunately, what I don't like about my current solution is that it involves way too much code. It is too easy to mess it up and piles complexity on top of complexity. And most of that complexity is unnecessary in my opinion.

If we only had to consider `cucumber-java` this would be a trivial problem. All step definitions are static so we would only have to use the locale as the cache key. But the workarounds used to make `cucumber-java8` work add a lot of complexity. It was originally hacked in and now about one third of the users use it. So we can't get rid of it easily, quickly, and possibly not at all.

With `cucumber-java8` step definitions are registered just after the test has started. So each step definition is unique for each scenario. And because step definitions retain a reference to a lambda, so they're state full and must be disposed between scenarios. 

And that makes for a really complicated cache key.

Then to make matters worse, each time a step definition is defined Cucumber emits a message. This generates a lot (scenarios * step definitions) messages. These messages contain an identifier that is used to emit another a message when a step is executed. 

So what I'm hoping to do is change the design of Cucumber so step definitions can be discovered at different times during the execution process. Each of these would be treated as a distinct set. This would then allow a cache key to be shaped like `cache-key = (locale, {glue, glue, glue, ...})`. 

Then to ensure that lambda step definitions consistently re-create the same key, they'd have to be created without any state. Which means the reference to the lambda they invoke must be stored in some execution context that is bound to the scenario scope instead (something like JUnit 5).

And then, once we can ensure step definitions are only created once, we can also ensure a message for each step definition is only emitted once. 

So I'm looking for an architecture where if done correctly, the solution to those problems just falls out of it.

jkronegg:
When your start caching data, this means that your are not following the Don't Repeat Yourself principle anymore. But of course sometimes you don't have the choice...

Regarding the `cucumber-java8` usage: it's maybe imported by mistake, but not used in practice (I did this at the beginning: "oh java8, that's modern, i'll take this one"). Can we detect this pattern?

jkronegg:
@mpkorstanje 

> Unfortunately, what I don't like about my current solution is that it involves way too much code. It is too easy to mess it up and piles complexity on top of complexity. And most of that complexity is unnecessary in my opinion.
> ... 
> So what I'm hoping to do is change the design of Cucumber so step definitions can be discovered at different times during the execution process. Each of these would be treated as a distinct set. This would then allow a cache key to be shaped like `cache-key = (locale, {glue, glue, glue, ...})`.
> ...
> So I'm looking for an architecture where if done correctly, the solution to those problems just falls out of it.

Is it okay for you if I work on the faster caching glue for the Cucumber v7 and we keep the architecture rework for Cucumber v8 ?
Because having a 50% faster Cucumber framework is really interesting and could be obtained with a limited development effort.

mpkorstanje:
Sure!

Though I don't think it wil be possible to robustly solve the problems that come from cucumber-java8. I'd be happy to be proven wrong.

PR #2923, opened 2024-09-08
Fix test flakiness due to undeterministic order of list of strings

adityabandal:
### 🤔 What's changed?

Fixed the implementation of flaky should_one_entry_for_each_failing_feature() test. Previous implementation results in non deterministic ordering of strings which may occasionally fail between multiple runs, as explained in the next section.

#### Proposed fix
- Convert the output `ByteArrayOutputStream` into strings and subsequently to `List` of strings (strings are separated by newline character).
- Sort this list of strings. Sorting will maintain the ordering of strings between different runs.
- Compare sorted list with expected list of strings, which always passes. 

The implementation of io.cucumber.core.plugin.RerunFormatterTest#should_one_entry_for_each_failing_feature was flaky. The ordering of strings `"classpath:path/first.feature:2"` and `"classpath:path/second.feature:2"` separated with newline changes between different runs due to the underlying implementation of cucumber Runtime builder. Thus, after compressing the string it may result in one of the following string - 
- `"classpath:path/first.feature:2\nclasspath:path/second.feature:2\n"`
- `"classpath:path/second.feature:2\nclasspath:path/first.feature:2\n"`

However, in the previous implementation we were always comparing it with `"classpath:path/first.feature:2\nclasspath:path/second.feature:2\n"`, which is not always true.

#### Steps to reproduce flaky tests
- The flakiness of this test was verified using [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool. NonDex explores different behaviors of under-determined APIs and reports test failures under different explored behaviors
- Run following command to run nonDex for targeted test
  - `./mvnw edu.illinois:nondex-maven-plugin:2.1.7:nondex -pl cucumber-core -Dtest=io.cucumber.core.plugin.RerunFormatterTest#should_one_entry_for_each_failing_feature`

### 📋 Checklist:

----

### DIFF
--- a/cucumber-core/src/test/java/io/cucumber/core/plugin/RerunFormatterTest.java
+++ b/cucumber-core/src/test/java/io/cucumber/core/plugin/RerunFormatterTest.java
@@ -12,6 +12,9 @@ import org.junit.jupiter.api.Test;
 import org.opentest4j.TestAbortedException;
 
 import java.io.ByteArrayOutputStream;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
 import static io.cucumber.core.plugin.Bytes.bytes;
 import static java.util.Arrays.asList;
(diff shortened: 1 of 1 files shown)


mpkorstanje:
Cheers! Though instead of fixing the tests, perhaps you can change the formatter to behave non-deterministically instead?

PR #2926, opened 2024-09-21
[Core] Support custom UUID generators in test runners

mpkorstanje:
### 🤔 What's changed?

Made it  possible to actually use the custom UUID generators in test runners

With #2703 a faster UUID generator was introduced. And while the
configuration options were added, they were not actually used by 
`cucumber-junit`, `cucumber-junit-platform-engine` and
`cucumber-testng`.

### 📋 Checklist:

### DIFF
--- a/.revapi/api-changes.json
+++ b/.revapi/api-changes.json
@@ -69,6 +69,26 @@
       }
     }
   ],
+  "7.20.0": [
+    {
+      "extension": "revapi.differences",
+      "id": "intentional-api-changes",
+      "ignore": true,
+      "configuration": {
+        "differences": [
+          {
+            "ignore": true,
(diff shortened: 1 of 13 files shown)


jkronegg (approved):
There are other calls to `UUID.randomUUID()` in the code, e.g `io.cucumber.junit.platform.engine.FeatureResolver` or `io.cucumber.core.runner.CoreHookDefinition` (the call in the `create()` method consumes about 1.2% of the total time on by usual test project).

Should this PR only be related to UUID generators used in test runners, or in the whole Cucumber framework ? (based on the title, I assume it's only in test runners, but having it for the whole framework could be also desirable)

mpkorstanje:
Ooh good one. I'll check it out.

Issue #2925, opened 2024-09-21
Out of IncrementingUuidGenerator capacity


tester12345132:
### 👓 What did you see?

- It looks like IncrementingUuidGenerator cannot close/reuse sessionId´s if testcases have a short execution time
- CucumberException: Out of IncrementingUuidGenerator capacity. Please reuse existing instances or use another UuidGenerator implementation instead.

```
Stacktrace: 
Exception in thread "main" java.util.ServiceConfigurationError: io.cucumber.core.eventbus.UuidGenerator: Provider io.cucumber.core.eventbus.IncrementingUuidGenerator could not be instantiated
	at java.base/java.util.ServiceLoader.fail(ServiceLoader.java:582)
	at java.base/java.util.ServiceLoader$ProviderImpl.newInstance(ServiceLoader.java:804)
	at java.base/java.util.ServiceLoader$ProviderImpl.get(ServiceLoader.java:722)
	at java.base/java.util.ServiceLoader$3.next(ServiceLoader.java:1395)
	at io.cucumber.core.runtime.UuidGeneratorServiceLoader.loadSingleUuidGeneratorOrDefault(UuidGeneratorServiceLoader.java:60)
	at io.cucumber.core.runtime.UuidGeneratorServiceLoader.loadUuidGenerator(UuidGeneratorServiceLoader.java:46)
	at io.cucumber.core.runtime.Runtime$Builder.build(Runtime.java:180)
	at io.cucumber.core.cli.Main.run(Main.java:85)
	at ta.TestRunner.main(TestRunner.java:59)
Caused by: io.cucumber.core.exception.CucumberException: Out of IncrementingUuidGenerator capacity. Please reuse existing instances or use another UuidGenerator implementation instead.
	at io.cucumber.core.eventbus.IncrementingUuidGenerator.<init>(IncrementingUuidGenerator.java:102)
	at jdk.internal.reflect.GeneratedConstructorAccessor87.newInstance(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
	at java.base/java.util.ServiceLoader$ProviderImpl.newInstance(ServiceLoader.java:780)
	... 7 more
```
### ✅ What did you expect to see?

[...]

PR #2931, opened 2024-10-05
Lazily start IncrementingUuidGenerator sessions

### LINKED ISSUE #2930: 7.20.0 regression: TestEngine with ID 'cucumber' failed to discover tests
### 👓 What did you see?

Build of PR opened by renovate with new Cucumber version is failing: https://github.com/jhipster/jhipster-lite/pull/11057:
>  Error:  Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.5.0:test (default-test) on project jhlite: 
Error:  
Error:  Please refer to /home/runner/work/jhipster-lite/jhipster-lite/target/surefire-reports for the individual test results.
Error:  Please refer to dump files (if any exist) [date].dump, [date]-jvmRun[N].dump and [date].dumpstream.
Error: [ERROR] There was an error in the forked process
Error:  TestEngine with ID 'cucumber' failed to discover tests
Error:  org.apache.maven.surefire.booter.SurefireBooterForkException: There was an error in the forked process
Error:  TestEngine with ID 'cucumber' failed to discover tests
Error:  	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork(ForkStarter.java:628)
Error:  	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.run(ForkStarter.java:285)
Error:  	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.run(ForkStarter.java:250)
Error:  	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.executeProvider(AbstractSurefireMojo.java:1337)
Error:  	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.executeAfterPreconditionsChecked(AbstractSurefireMojo.java:1137)
Error:  	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.execute(AbstractSurefireMojo.java:971)
Error:  	at com.gradle.maven.cache.extension.d.c.a(SourceFile:27)

Build running fine, as it was with 7.19.0

cucumber-junit-platform-engine
cucumber-spring
maven-failsafe-plugin
gradle-develocity

_No response_

### 📚 Any additional context?

Our cucumbers tests are executed with maven-failsafe-plugin, but here the error happens with the execution of maven-surefire-plugin

mpkorstanje:

Even when not used, the `IncrementingUuidGenerator` is instantiated with each Cucumber execution. This somewhat fundamental to the way the `ServiceLoader` mechanism works.

Each time an incrementing generator is created, a new session is started for that generator. This means that after 255 executions all sessions are exhausted.

Unfortunately, when using the JUnit Platform, Maven issues a discovery request for each individual class. And Cucumber typically participates in discovery along with JUnit Jupiter. So after requests for 255 classes, Cucumber fails discovery as seen in #2930.

Fixes #2930, #2925

### 🤔 What's changed?

Lazily start IncrementingUuidGenerator sessions

@jkronegg best I could think of. Got any other ideas?

### 📋 Checklist:

  - [X] I have added/updated tests to cover my changes.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- [Core] Lazily start IncrementingUuidGenerator sessions([#2931](https://github.com/cucumber/cucumber-jvm/pull/2931) M.P. Korstanje)
 
 ## [7.20.0] - 2024-10-04
 ### Added
--- a/cucumber-core/src/main/java/io/cucumber/core/eventbus/IncrementingUuidGenerator.java
+++ b/cucumber-core/src/main/java/io/cucumber/core/eventbus/IncrementingUuidGenerator.java
@@ -68,7 +68,7 @@ public class IncrementingUuidGenerator implements UuidGenerator {
     /**
      * Computed UUID MSB value.
      */
-    final long msb;
+    private long msb;
 
     /**
      * Counter for the UUID LSB.
@@ -77,13 +77,13 @@ public class IncrementingUuidGenerator implements UuidGenerator {
 
     /**
      * Defines a new classloaderId for the class. This only affects instances
-     * created after the call (the instances created before the call keep their
-     * classloaderId). This method should be called to specify a classloaderId
-     * if you are using more than one class loader, and you want to guarantee a
-     * collision-free UUID generation (instead of the default random
-     * classloaderId which produces about 1% collision rate on the
-     * classloaderId, and thus can have UUID collision if the epoch-time,
-     * session counter and counter have the same values).
+     * created after the first call to {@link #generateId()} (the instances
+     * created before the call keep their classloaderId). This method should be
+     * called to specify a {@code classloaderId} if you are using more than one
+     * class loader, and you want to guarantee a collision-free UUID generation
+     * (instead of the default random classloaderId which produces about 1%
+     * collision rate on the classloaderId, and thus can have UUID collision if
+     * the epoch-time, session counter and counter have the same values).
      *
      * @param classloaderId the new classloaderId (only the least significant 12
      *                      bits are used)
@@ -94,6 +94,10 @@ public class IncrementingUuidGenerator implements UuidGenerator {
     }
 
     public IncrementingUuidGenerator() {
+
+    }
+
+    private long initializeMsb() {
         long sessionId = sessionCounter.incrementAndGet();
         if (sessionId == MAX_SESSION_ID) {
             throw new CucumberException(
@@ -103,7 +107,7 @@ public class IncrementingUuidGenerator implements UuidGenerator {
         }
         long epochTime = System.currentTimeMillis();
         // msb = epochTime | sessionId | version | classloaderId
-        msb = ((epochTime & MAX_EPOCH_TIME) << 24) | (sessionId << 16) | (8 << 12) | classloaderId;
+        return ((epochTime & MAX_EPOCH_TIME) << 24) | (sessionId << 16) | (8 << 12) | classloaderId;
     }
 
     /**
@@ -114,6 +118,11 @@ public class IncrementingUuidGenerator implements UuidGenerator {
      */
     @Override
     public UUID generateId() {
+        if (msb == 0) {
+            // Lazy init to avoid starting sessions when not used.
+            msb = initializeMsb();
+        }
+
         long counterValue = counter.incrementAndGet();
         if (counterValue == MAX_COUNTER_VALUE) {
             throw new CucumberException(
--- a/cucumber-core/src/test/java/io/cucumber/core/eventbus/IncrementingUuidGeneratorTest.java
+++ b/cucumber-core/src/test/java/io/cucumber/core/eventbus/IncrementingUuidGeneratorTest.java
@@ -1,8 +1,8 @@
 package io.cucumber.core.eventbus;
 
 import io.cucumber.core.exception.CucumberException;
-import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.function.ThrowingSupplier;
 
 import java.io.IOException;
 import java.lang.reflect.Field;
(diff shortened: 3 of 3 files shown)


jkronegg:
**TLDR**: I suggest adding a default `UuidGenerator.initialize()` method which does nothing, but is implemented by `IncrementingUuidGenerator`, and add some test-cases. I can update the PR if you want.

---
The issue #2925 can reproduced easily in `UuidGeneratorServiceLoaderTest` with two different patterns:
- many calls to the `loadUuidGenerator()` method on one `UuidGeneratorServiceLoader` instance
- calling once `loadUuidGenerator()` on many `UuidGeneratorServiceLoader` instance

This corresponds to the two following unit tests:

    @Test
    void test_case_2_no_exception_when_calling_loadUuidGenerator_many_times() {
        Options options = () -> null;
        UuidGeneratorServiceLoader loader = new UuidGeneratorServiceLoader(
                UuidGeneratorServiceLoaderTest.class::getClassLoader,
                options);
        UuidGenerator actual = null;
        for (int i=0; i<1000; i++) {
            actual = loader.loadUuidGenerator();
        }
        assertThat(actual, instanceOf(RandomUuidGenerator.class));
    }

    @Test
    void test_case_2_no_exception_when_instantiating_UuidGeneratorServiceLoader_many_times() {
        Options options = () -> null;
        UuidGenerator actual = null;
        for (int i=0; i<1000; i++) {
            UuidGeneratorServiceLoader loader = new UuidGeneratorServiceLoader(
                    UuidGeneratorServiceLoaderTest.class::getClassLoader,
                    options);
            actual = loader.loadUuidGenerator();
        }
        assertThat(actual, instanceOf(RandomUuidGenerator.class));
    }

>  This somewhat fundamental to the way the `ServiceLoader` mechanism works.

I think it's more related to the logic used to select the UUID generator. When calling the `UuidGeneratorServiceLoader.loadUuidGenerator` the generator will be selected [based on the logic summarized in the test file](https://github.com/cucumber/cucumber-jvm/blob/main/cucumber-core/src/test/java/io/cucumber/core/runtime/UuidGeneratorServiceLoaderTest.java#L21-L36):

> | #   | uuid-generator property   | Available services                                                                  | Result                                                                           |
> |-----|---------------------------|-------------------------------------------------------------------------------------|----------------------------------------------------------------------------------|
> | 1   | undefined                 | none                                                                                | exception, no generators available                                               |
> | 2   | undefined                 | RandomUuidGenerator, IncrementingUuidGenerator                                      | RandomUuidGenerator used                                                         |
> | 3   | RandomUuidGenerator       | RandomUuidGenerator, IncrementingUuidGenerator                                      | RandomUuidGenerator used                                                         |
> | 4   | undefined                 | RandomUuidGenerator, IncrementingUuidGenerator, OtherGenerator                      | OtherGenerator used                                                              |
> | 5   | RandomUuidGenerator       | RandomUuidGenerator, IncrementingUuidGenerator, OtherGenerator                      | RandomUuidGenerator used                                                         |
> | 6   | undefined                 | RandomUuidGenerator, IncrementingUuidGenerator, OtherGenerator, YetAnotherGenerator | exception, cucumber couldn't  decide multiple (non default) generators available |
> | 7   | OtherGenerator            | RandomUuidGenerator, IncrementingUuidGenerator, OtherGenerator, YetAnotherGenerator | OtherGenerator used                                                              |
> | 8   | IncrementingUuidGenerator | RandomUuidGenerator, IncrementingUuidGenerator, OtherGenerator, YetAnotherGenerator | IncrementingUuidGenerator used                                                   |
> | 9   | IncrementingUuidGenerator | RandomUuidGenerator, IncrementingUuidGenerator                                      | IncrementingUuidGenerator used                                                   |
> | 10  | OtherGenerator            | none                                                                                | exception, generator OtherGenerator not available                                |
> | 11  | undefined                 | OtherGenerator                                                                      | OtherGenerator used                                                              |
> | 12  | undefined                 | IncrementingUuidGenerator, OtherGenerator                                           | OtherGenerator used                                                              |
> | 13  | undefined                 | IncrementingUuidGenerator                                                           | IncrementingUuidGenerator used                                                   |

In other words, a custom generator written outside Cucumber has the priority over the `RandomUuidGenerator` and the `IncrementingUuidGenerator`, even if the `cucumber.uuid-generator` property is not specified. To implement this behavior, it is mandatory to iterate over the `ServiceLoader<UuidGenerator>.iterator()`, thus to instantiate the each generator. I don't think we should change this logic (even if the UUID generator is not very popular within the Cucumber community, changing behavior is risky).

Iterating the `ServiceLoader` more than once could be avoided by using caching, but we would need to take a lot of parameters into account (discard the cache when the `cucumber.uuid-generator` property changes or when the classloader changes or loads new classes). But this would be very complex and error-prone due to the many different existing usages of Cucumber.

Lazy loading is the only simple solution that I can see (see the Hyrum's law ;-) ), but I think it should be done only once and not on every call to `generateId()` (because it would decrease the performance). Instead, I suggest introducing a default `initialize()` method which is called only for the selected generator. Something like:

    public interface UuidGenerator extends Supplier<UUID> {
        UUID generateId();

        default UUID get() {
            return generateId();
        }
    
       default void initialize() {
       }
    }

In the `UuidGeneratorServiceLoader`: 

    ....
    UuidGenerator selectedUuidGenerator = ...// decide which generator to use
    selectedUuidGenerator.initialize();
    ...

In the `IncrementingUuidGenerator`:

    public class IncrementingUuidGenerator implements UuidGenerator {
        ...
        public void initialize() {
            initializeMsb();
        }
    }

mpkorstanje:
I think this will do for now.

The ideal solution would be to have the `UuidGenerator` return a supplier instance that keeps the state internally. 

```
public interface UuidGenerator {

    Supplier<UUID> supplier();
}
```

And then have the `UuidGeneratorServiceLoader` return that supplier only for the selected generator.  Fortunately the `UuidGenerator` is experimental so I don't feel too bad about breaking the contract. But I'd rather avoid doing that on a patch version.

francoissey:
I have just upgrade to 7.20.1 and I run into this issue, could we please reopen it?

jkronegg:
@francoissey it seems that you are experiencing another bug variant on this mechanism. Could you please open a new issue to describe your setup and behaviour ? (BTW, this thread is a merged Pull Request: it's not supposed to be reopened)

Theo-Hafsaoui:
Hello, I might be experiencing an issue caused by this regression. Reproducing it is however cumbersome due to the heavy deployment process, and I was wondering if you could estimate which version this bug was introduced in?

mpkorstanje:
Please start a new issue. Otherwise we got nothing to work off.

And be sure to use the `cucumber-bom` to enforce dependency alignment as well as `mvn dependency:tree -Dverbose` to verify alignment.

PR #2939, opened 2024-11-04
Fix flaky test-cases in HtmlFormatterTest

hermya:
### 🤔 What's changed?
Changed the way of asserting fields in class `HtmlFormatterTest.java`

The Assertion for `bytes` to contain the string:
```js
 window.CUCUMBER_MESSAGES = [
                {"testRunStarted":{"timestamp":{"seconds":10,"nanos":0}}},
                {"testRunFinished":{"success":true,"timestamp":{"seconds":15,"nanos":0}}}
       ];"
```
is **flaky** in nature. Precisely, the value assigned to  `window.CUCUMBER_MESSAGES` is a *list of objects* in `js`, but apparently derived from a *List of HashMaps* in java. HashMaps being unordered, can sometimes return a result like:
```js
 window.CUCUMBER_MESSAGES = [
                {"testRunStarted":{"timestamp":{"nanos":0, "seconds":10}}}, # different order, same object/contents
                {"testRunFinished":{"success":true,"timestamp":{"seconds":15,"nanos":0}}}
       ];"
```
The test may fail on a **different jvm** pertaining to the hashcode implementation. <br>
~The fix contains logic that parses the bytes, converting them to `List<LinkedHashMap<String, Object>>` which can be safely used for **equals check** (safer than string/bytes comparison).~ (fix modified in the commit : https://github.com/cucumber/cucumber-jvm/pull/2939/commits/6ae3bf6ce42de987284c35b79334bd02172bd463)

**Fixed using nondex plugin to identify Flaky tests**
Use the following command to test the flakiness of a test:
>mvn -pl ./cucumber-core edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=io.cucumber.core.plugin.HtmlFormatterTest -DnondexRuns=5 -Dmode=ONE

For entire repo:
>mvn edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dmode=ONE

### 📋 Checklist:

  - [X] I have added/updated tests to cover my changes.

----

### DIFF
--- a/cucumber-core/src/test/java/io/cucumber/core/plugin/HtmlFormatterTest.java
+++ b/cucumber-core/src/test/java/io/cucumber/core/plugin/HtmlFormatterTest.java
@@ -1,5 +1,7 @@
 package io.cucumber.core.plugin;
 
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import io.cucumber.core.eventbus.EventBus;
 import io.cucumber.core.runtime.TimeServiceEventBus;
 import io.cucumber.messages.types.Envelope;
@@ -12,15 +14,20 @@ import io.cucumber.messages.types.StepDefinitionPatternType;
 import io.cucumber.messages.types.TestRunFinished;
 import io.cucumber.messages.types.TestRunStarted;
 import io.cucumber.messages.types.Timestamp;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 import java.io.ByteArrayOutputStream;
+import java.nio.charset.StandardCharsets;
 import java.time.Clock;
 import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.UUID;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
-import static io.cucumber.core.plugin.Bytes.bytes;
-import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.MatcherAssert.assertThat;
 
 class HtmlFormatterTest {
@@ -37,12 +44,14 @@ class HtmlFormatterTest {
 
         TestRunFinished testRunFinished = new TestRunFinished(null, true, new Timestamp(15L, 0L), null);
         bus.send(Envelope.of(testRunFinished));
-
-        assertThat(bytes, bytes(containsString("" +
+        List<LinkedHashMap<String, Object>> actual = rectifiedRepresentation(getWindowCucumberMessagesFromBytes(bytes));
+        List<LinkedHashMap<String, Object>> expected = rectifiedRepresentation("" +
                 "window.CUCUMBER_MESSAGES = [" +
                 "{\"testRunStarted\":{\"timestamp\":{\"seconds\":10,\"nanos\":0}}}," +
                 "{\"testRunFinished\":{\"success\":true,\"timestamp\":{\"seconds\":15,\"nanos\":0}}}" +
-                "];\n")));
+                "];\n");
+
+        assertThat(actual, equalTo(expected));
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


mpkorstanje (changes requested):
This adds a lot of complexity to what is a very simple test, only to satisfy a tool based finding. Unless this actually affects people, I don't think think this worth addressing. If you do have a solution that retains the current level of simplicity I would be willing to consider it.

I also can't help but notice that you, the author of https://github.com/cucumber/cucumber-jvm/pull/2923 and the nondex-maven-plugin are all related to the University of Illinois. What is going on?

hermya:
I understand and agree that this adds a lot of complexity to the test. I'll take another look at the code and try to fix it internally  to address ordering. 

### Regarding [nondex](https://github.com/TestingResearchIllinois/NonDex)
We are a group of student in the same [course](https://siebelschool.illinois.edu/academics/courses/cs527). We track popular open-source projects from this [repository](https://github.com/TestingResearchIllinois/idoft). We try to solve multiple variations of flaky test cases, one of the popular ones is **ID** type. These test cases (such as the one in this PR), can fail in different `jvms`, depending on how ordering changes for documented unordered collections like `HashMaps` or `HashSets`. **Nondex** helps in highlighting such tests by running them in multiple iterations, each with a **slight change introduced in ordering** of such collections.

jkronegg:
The second commit ([6ae3bf6](https://github.com/cucumber/cucumber-jvm/pull/2939/commits/6ae3bf6ce42de987284c35b79334bd02172bd463)) simplifies `HtmlFormatterTest` but requires to change the production code. 

Usually, I don't like the idea to change production code just to correct a flaky test. But here, I think it makes sense because the produced HTML report will be more stable (e.g. if the produced HTML is committed to a code repository, there will be no artificial commit differences when only the order changes).

I don't think there will be some performance issues by sorting the JSON properties. But there may be a small effect on existing customers as the current order ("seconds", then "nanos") changes to "nanos" then "seconds" due to alphabetical sort.

hermya:
Hello @jkronegg

I agree that this may have a small effect on the customer's end. However, it depends on how the object <br>` window.CUCUMBER_MESSAGES` is unpacked or reused in UI. There can be two cases:
- **Existing code is deployed in different jvms without testing**: If any issue would arise in the UI due to non-deterministic ordering of "time" and "nanos", they would've been reported before.
- **New code "possibly" breaks some UI test-cases**: This is scary, but will only cause issues in cases where the object is used without unpacking OR its string value is used as it is.

In both cases, I think that the fault would be simple to locate and mitigate.

mpkorstanje:
I didn't think ordering the properties of every message written by Cucumber would be a good idea.  Json objects are specified as unordered collections of properties. Though in practice, they retain insertion order.  By sorting them alphabetically we'd invite Hyrum's law. 

Using `JSONAssert.assertEquals` we can compare JSON without writing a complex normalization machinery  from 0fa8f847dce46df8e2d7c56cd6fcb04ac5942cc2 ourselves.

cucumber-core/src/test/java/io/cucumber/core/plugin/HtmlFormatterTest.java
```diff
@@ -86,11 +89,17 @@ void ignores_step_definitions() throws Throwable {
...
-                "];\n")));
+        assertEquals("[" +
+                "{\"testRunStarted\":{\"timestamp\":{\"nanos\":0,\"seconds\":10}}}," +
+                "{\"testRunFinished\":{\"success\":true,\"timestamp\":{\"nanos\":0,\"seconds\":15}}}" +
+                "]",
+            extractCucumberMessages(bytes), STRICT);
     }
 
+    private static String extractCucumberMessages(ByteArrayOutputStream bytes) {
+        Pattern pattern = Pattern.compile("^.*window\\.CUCUMBER_MESSAGES = (\\[.+]);.*$", Pattern.DOTALL);
```
jkronegg, line 100: The `Pattern` should compiled only once, so should be as a constant in the class.

mpkorstanje: We'll revisit that once there are more than two testcases. :smile:

hermya:
Hello @mpkorstanje <br>
First of all, this is a great solution 👍  as it is risk-free for the clients dependent on this JSON! However, this test-case will still fail because, on different JVMs, the order of JSON may not be the same. This is the same case for which I raised the PR: **The Assertion for `bytes` to contain the string**
```js
 window.CUCUMBER_MESSAGES = [
                {"testRunStarted":{"timestamp":{"seconds":10,"nanos":0}}},
                {"testRunFinished":{"success":true,"timestamp":{"seconds":15,"nanos":0}}}
       ];"
```
is **flaky** in nature. Precisely, the value assigned to  `window.CUCUMBER_MESSAGES` is a *list of objects* in `js`, but apparently derived from a *List of HashMaps* in java. HashMaps being unordered, can sometimes return a result like:
```js
 window.CUCUMBER_MESSAGES = [
                {"testRunStarted":{"timestamp":{"nanos":0, "seconds":10}}}, # different order, same object/contents
                {"testRunFinished":{"success":true,"timestamp":{"seconds":15,"nanos":0}}}
       ];"
```
**The actual solution** 🥇   to this would be to use: `org.skyscreamer.jsonassert.JSONCompareMode.NON_EXTENSIBLE;` in assertEquals. Why? Here's the [documentation](https://jsonassert.skyscreamer.org/apidocs/org/skyscreamer/jsonassert/JSONCompareMode.html)

mpkorstanje:
Can you reproduce that with non-dex?

hermya:
Ah sorry, my confusion `STRICT_ORDERING` applies to objects in an array, not the keys inside it. My bad, this is a fully correct fix 🥇 

Cheers!

PR #2944, opened 2024-11-28
Fix Null Pointer Exception and improve getNormalizedLanguage method

Mr-Lucasz:
### 🤔 What's changed?

This PR fixes a potential Null Pointer Exception in the `getNormalizedLanguage` method by adding null checks for the `GherkinDialect` object and its `getLanguage` method. It also improves the regular expression used to replace unwanted characters in the language string.

This change is motivated by the need to enhance code robustness and prevent runtime exceptions due to null values. It addresses potential null pointer exceptions and ensures that the language normalization process is more reliable.

Feedback on the null checks and the improved regular expression for replacing characters in the language string would be appreciated.

### 📋 Checklist:

- [x] I agree to respect and uphold the [Cucumber Community Code of Conduct](https://github.com/cucumber/.github/tree/main?$tab=coc-ov-file$)

----

### DIFF
--- a/cucumber-java/src/codegen/java/GenerateI18n.java
+++ b/cucumber-java/src/codegen/java/GenerateI18n.java
@@ -1,14 +1,10 @@
-import freemarker.template.Configuration;
-import freemarker.template.Template;
-import freemarker.template.TemplateException;
-import freemarker.template.TemplateExceptionHandler;
-import io.cucumber.gherkin.GherkinDialect;
-import io.cucumber.gherkin.GherkinDialectProvider;
-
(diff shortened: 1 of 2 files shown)


mpkorstanje:
This pull request doesn't make much sense. 

* If you infer the nullability of the elements being null checked, then they are non null.
* The language does not contain any `$` signs so they need not be removed.
* The stated motivation, to prevent null pointers, that never happen, apparently by replacing them with illegal argument exceptions, does not actually make the code more robust.

I also get the strong impression that this pull request was written by AI.

PR #2960, opened 2025-01-13
Snippet.template() Now Accepts a Step Parameter

danielmiladinov:
### 🤔 What's changed?

`Snippet.template()` Now Accepts a `Step` Parameter

Although none of the `Snippet` implementations within the project itself need it, receiving the current `step` as a parameter would allow certain external snippet implementations (specifically, in my case, my [burpless](https://github.com/danielmiladinov/burpless) library that wraps Cucumber-JVM and makes it easier to use from Clojure) to decide, based on information about the step argument, which `MessageFormat` to return: a default value, one specialized for steps with a `DataTableArgument`, and one specialized for steps with a `DocStringArgument`.

- [x] I agree to respect and uphold the [Cucumber Community Code of Conduct]

### DIFF
--- a/cucumber-core/src/main/java/io/cucumber/core/backend/Snippet.java
+++ b/cucumber-core/src/main/java/io/cucumber/core/backend/Snippet.java
@@ -1,5 +1,6 @@
 package io.cucumber.core.backend;
 
+import io.cucumber.core.gherkin.Step;
 import org.apiguardian.api.API;
 
 import java.lang.reflect.Type;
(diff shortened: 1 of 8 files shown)


mpkorstanje:
As is this is a breaking change.

It would have been better to start with an issue to describe the problem first. I had a look at the documentation for Burpless and right nowI don't quite understand what problem you would solve here.

So I don't see a good reason to accept this.

> Burpless' implementation of the Cucumber Backend interface is responsible for adding StepDefinitions to the Glue instance provided to it during the call to loadGlue(), and they must return ParameterInfo lists that match what the Cucumber runtime discovered while parsing the feature file(s) into Gherkin steps, or else Cucumber will report that step as undefined.
>
> The current design of the Cucumber JVM library tries to make it very easy to identify the code that should run for a particular Gherkin step - assuming that your JVM language is strongly typed, and has excellent annotation support. Just annotate your methods with the appropriate annotation(s), and the cucumber runtime does the rest!
>
> Coming from Clojure, that's two strikes against us.

Though I'm not very familiar with Clojure, reading the this I think I there maybe a misunderstanding.

Matching steps to step definitions is done exclusively by matching the step text to the cucumber or regular expression. The parameter infos are used as type hints. `Object.class` can be used if the type isn't known.

There is a sanity check on the arity of the step definition that is invoked but this check is omitted if no parameter info is provided at all.

https://github.com/cucumber/cucumber-jvm/blob/6ec14f7edce8d9c8a5d679b3d1eae22fbc7b8675/cucumber-core/src/main/java/io/cucumber/core/runner/PickleStepDefinitionMatch.java#L37

danielmiladinov:
>So I think it would be a good idea to start from scratch with an issue with the root cause of your problem.
Thanks, I'll do that!

danielmiladinov:
You were right; I didn't need to change the public API. I could get the information I need by making my own pickles out of the feature file, which will give me the argument information associated with each step text that I need to supply the correct parameter info.

PR #2971, opened 2025-02-01
[Core] Improve caching glue performance

jkronegg:
### 🤔 What's changed?

The `CachingGlue` performance has been improved by caching the following elements:

- parameter types
- data table types
- docString types
- step definitions

The cache is invalidated when a parameter type or a step definition is added/removed, or when the language changes.
The cache is not used when scenario-scoped glue is present (e.g. cucumber-java8).

This improves the `CachingGlue` performance.

On a personal project with about 250 step definitions and 1000 test scenarios, the performance is the following:

| Version | Test duration |
|---------|----------------|
|7.20.1 (original) | 8.2 seconds |
|7.21.0-SNAPSHOT (this PR) | 3.4 seconds (2.4x faster) |

I'm not 100% sure about whether `cucumber-java8` has still the same behavior, but at least the unit tests did not change.
I accidentally deleted the original PR, sorry.

### 📋 Checklist:


cucumber-core/src/main/java/io/cucumber/core/runner/CachingGlue.java
```diff
@@ -130,11 +136,13 @@ public void addAfterStepHook(HookDefinition hookDefinition) {
     @Override
     public void addParameterType(ParameterTypeDefinition parameterType) {
         parameterTypeDefinitions.add(parameterType);
+        dirtyCache = true;
     }
 
     @Override
     public void addDataTableType(DataTableTypeDefinition dataTableType) {
         dataTableTypeDefinitions.add(dataTableType);
+        dirtyCache = true;
```
mpkorstanje, line 145: To wit, data tables are not involved in regex creation. So they shouldn't dirty the cache.

cucumber-core/src/main/java/io/cucumber/core/runner/CachingGlue.java
```diff
@@ -160,6 +168,7 @@ public void addDefaultDataTableCellTransformer(
     @Override
     public void addDocStringType(DocStringTypeDefinition docStringType) {
         docStringTypeDefinitions.add(docStringType);
+        dirtyCache = true;
```
mpkorstanje, line 171: To wit, doc strings are not involved in regex creation. So they shouldn't dirty the cache.

cucumber-core/src/main/java/io/cucumber/core/runner/CachingGlue.java
```diff
@@ -228,17 +237,38 @@ List<DocStringTypeDefinition> getDocStringTypeDefinitions() {
...
-        docStringTypeDefinitions.forEach(dtd -> stepTypeRegistry.defineDocStringType(dtd.docStringType()));
+        if (mustRebuildCache) {
+            // parameters changed from the previous scenario => re-register them
+            parameterTypeDefinitions.forEach(ptd -> {
+                ParameterType<?> parameterType = ptd.parameterType();
+                stepTypeRegistry.defineParameterType(parameterType);
+                emitParameterTypeDefined(ptd);
+            });
+            dataTableTypeDefinitions.forEach(dtd -> stepTypeRegistry.defineDataTableType(dtd.dataTableType()));
+            docStringTypeDefinitions.forEach(dtd -> stepTypeRegistry.defineDocStringType(dtd.docStringType()));
```
mpkorstanje, line 270: These can be kept outside.

mpkorstanje (commented):
Okay. Technically this looks way more manageable then I expected.

To keep the scope of this PR manageable, I think it would be prudent to initially always consider the cache dirty in the presence of `ScenarioScoped` glue code. 

It means we won't get the performance gains for `cucumber-java8` but we can then look removing `ScenarioScoped` from `cucumber-java8` separately.  If there are other backends that do use `ScenarioScoped` glue, they can either stop using that or contribute the next improvement.

cucumber-core/src/main/java/io/cucumber/core/runner/CachingGlue.java
mpkorstanje: Ah no, I see the problem. The step registry also gets dirtied.

cucumber-core/src/main/java/io/cucumber/core/runner/CachingGlue.java
```diff
@@ -442,30 +481,43 @@ private List<PickleStepDefinitionMatch> stepDefinitionMatches(URI uri, Step step
...
+        dirty |= removeScenarioScopedGlue(docStringTypeDefinitions);
+        if (removeScenarioScopedGlue(parameterTypeDefinitions)) {
+            dirty = true;
+            dirtyCache = true;
+        }
+        dirty |= removeScenarioScopedGlue(defaultParameterTransformers);
+        dirty |= removeScenarioScopedGlue(defaultDataTableEntryTransformers);
+        dirty |= removeScenarioScopedGlue(defaultDataTableCellTransformers);
+        if (dirty) {
+            stepDefinitionsByPattern.clear();
```
mpkorstanje, line 503: Is this still necessary given that `hasScenarioScopedGlue` will also clear the step definition patterns?

cucumber-core/src/main/java/io/cucumber/core/runner/CachingGlue.java
```diff
@@ -229,17 +243,40 @@ List<DocStringTypeDefinition> getDocStringTypeDefinitions() {
...
 
         // TODO: separate prepared and unprepared glue into different classes
-        parameterTypeDefinitions.forEach(ptd -> {
-            ParameterType<?> parameterType = ptd.parameterType();
-            stepTypeRegistry.defineParameterType(parameterType);
-            emitParameterTypeDefined(ptd);
-        });
-        dataTableTypeDefinitions.forEach(dtd -> stepTypeRegistry.defineDataTableType(dtd.dataTableType()));
-        docStringTypeDefinitions.forEach(dtd -> stepTypeRegistry.defineDocStringType(dtd.docStringType()));
+        if (mustRebuildCache) {
```
mpkorstanje, line 270: I think all code in the `prepareGlue` method can be kept within this if statement. Do you agree?

cucumber-java8/src/main/java/io/cucumber/java8/ClosureAwareGlueRegistry.java
```diff
@@ -0,0 +1,113 @@
...
+import io.cucumber.core.backend.Glue;
+import io.cucumber.core.backend.HookDefinition;
+import io.cucumber.core.backend.ParameterTypeDefinition;
+import io.cucumber.core.backend.StepDefinition;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
+
+final class ClosureAwareGlueRegistry implements LambdaGlueRegistry {
```
mpkorstanje, line 17: This class needs a sanity check to ensure that the number of abstract glue definitions on the first run matches the number or registered/updated glue definitions on any subsequent runs.


```
if(registered != definitions.size()) {
   throw new CucumberException...
}
```

Could be invokes at the end of `buildWorld`?

cucumber-core/src/main/java/io/cucumber/core/runner/CachingGlue.java
jkronegg: Good catch.

Given that `Runner.runPickle()` prepares the glue for each pickle:
```
        try {
            ...
            glue.prepareGlue(localeForPickle(pickle));
            ...
        } finally {
            glue.removeScenarioScopedGlue();
            ...
        }
```
then we don't need to call `stepDefinitionsByPattern.clear()` in `removeScenarioScopedGlue()`. Clearing the `stepDefinitionsByPattern` will be done in `prepareGlue()` when we run the next pickle.

cucumber-core/src/main/java/io/cucumber/core/runner/CachingGlue.java
jkronegg: You mean, including the before/after hook/step hook event emitting or only the default transformers ? It would be ok to include the default transformers in the `if`. But the before/after hook and step hook event emitting behavior will change: currently, we emit the events for each call to `prepareGlue` (that is every time we run a pickle, if I'm right). If we enclose the emit throwing into the `if (mustRebuildCache)` guard, then we will only emit events only when the cache is refreshed. This makes the `compatibility` checks to fail.

mpkorstanje: Okay. I'll look at that.

mpkorstanje: Only emitting messages when the cache is updated is the expected behavior. Otherwise we're emitting duplicate messages for hooks and step definitions that have already been defined.

mpkorstanje: I think the failure was caused by non-determinism around the ordering of hooks.

mpkorstanje:
TODO:
 * Remove workaround for https://github.com/cucumber/cucumber/issues/1232
 * Deprecate scenario scoped glue.

mpkorstanje (approved):
Looks good! Thanks for doing the initial leg work.

This also had a positive knock on effect on the number of messages created by Cucumber. Because we're reusing step definitions they don't need to be redefined for each scenario.

So that's an improvement of `# scenarios * # step  definitions`  to just `# step definitions`.

PR #2973, opened 2025-02-04
[junit-platform] Document how to run a single feature with Gradle

TrueJacobG:
### 🤔 What's changed?

Resolve TODO in README.md

Resolve TODO in README.md

### 📋 Checklist:

- [X] I have updated the documentation.

### DIFF
--- a/cucumber-junit-platform-engine/README.md
+++ b/cucumber-junit-platform-engine/README.md
@@ -180,7 +180,21 @@ mvn test -Dsurefire.includeJUnit5Engines=cucumber -Dcucumber.plugin=pretty -Dcuc
 
 #### Gradle
 
-TODO: (I don't know how. Feel free to send a pull request. ;))
+Remember to define Cucumber properties before running the test to ensure that your build.gradle (or build.gradle.kts) correctly passes system properties to the test task.
+
+```groovy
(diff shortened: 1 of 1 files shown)


mpkorstanje:
Cheers! Great PR title. 😄

tveeders:
Just wanted to add this bit of information to the docs, too :)
Glad to see its already a PR. I can confirm it works

mpkorstanje:
Yeah sorry. My time is a bit fragmented right now. Plenty of short intervals to look at issues, not enough time to do anything.

mpkorstanje:
@tveeders since you both mentioned that this works, how did you solve the problem of executing Cucumber twice with Gradle?

Currently when running `gradle test -Dcucumber.features="path/to/example.feature:10"` Cucumber will be executed twice. Once directly through the Cucumber Engine, and once indirectly through the Suite Engine. 

Normally the direct execution of Cucumber wouldn't do anything because there were no selectors provided to the JUnit Platform, but the `cucumber.features` property overrides that.

Then the second execution happens when the Suite engine runs which also runs Cucumber, which again now only looks at the `cucumber.features` property. 

For Maven this is solved by excluding the Cucumber Engine.

Issue #2994, opened 2025-05-08
Assertion failing when using a parameterized type in a cucumber step with


jean-tissot:
### 👓 What did you see?

When using @DefaultDataTableEntryTransformer to transform cucumber DataTables to objects, an assertion fails in [DataTableTypeRegistryTableConverter](https://github.com/cucumber/cucumber-jvm/blob/main/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistryTableConverter.java#L109) (at line 109) for parameterized types :
```java
assert listElementType instanceof OtherType || listElementType instanceof OptionalType;
```

### ✅ What did you expect to see?

The java assertion should not fail when the target type is a parameterized type.
When disabling java assertions, the DataTable conversion works perfectly (so why is there a restrictive assertion ?)

### 📦 Which tool/library version are you using?

- Java 21
- Cucumber 7.22.1
- Junit 5.12.1

### 🔬 How could we reproduce it?

I created a POC to reprocue the issue : [cucumber-jvm-assert-issue](https://github.com/jean-tissot/cucumber-jvm-assert-issue). Explanations on how to reproduce the issue are in the README (execute `mvn test`).

### 📚 Any additional context?

<pre>
java.lang.AssertionError
    at io.cucumber.datatable.DataTableTypeRegistryTableConverter.convert(DataTableTypeRegistryTableConverter.java:109)
    at io.cucumber.datatable.DataTable.convert(DataTable.java:478)
    at io.cucumber.core.stepexpression.StepExpressionFactory.lambda$createExpression$0(StepExpressionFactory.java:66)
    at io.cucumber.core.stepexpression.DataTableArgument.getValue(DataTableArgument.java:20)
    at io.cucumber.core.runner.PickleStepDefinitionMatch.runStep(PickleStepDefinitionMatch.java:46)
    ...
</pre>

mpkorstanje:
Thanks for the clear report. Looks like the biggest complication will adding a few more tests!

>  When disabling java assertions, the DataTable conversion works perfectly (so why is there a restrictive assertion ?)

Mostly to document the expectations and catch bugs where they happen. 

The expectation is that the `convert` method covered all types by the time it reaches an assertion. And the assertion no longer holds because the `Parameterized` type was added after the other types. And it looks like I didn't write a test that covered this code path. It also looks  like the call to `toSingleTon` in `convert` should also be possible for `Parameterized` types.

jean-tissot:
Hello,
[...]

PR #2995, opened 2025-05-11
Convert parameterized objects with @DefaultDataTableEntryTransformer

jean-tissot:
Fixes: #2994

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+## Fixed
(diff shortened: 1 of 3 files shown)


jean-tissot:
Hello,
I don't know if this bug fix should be added to the changelog so I didn't add it. I can do it if necessary (is it a [Core] fix or a [Java] fix ?)

mpkorstanje:
That's okay! It's a [Core] fix. The [Java] fix would be used for `cucumber-java` module. A bit confusing but we also had `cucumber-scala`, `cucmber-groovy` and a few others.

mpkorstanje (commented):
On review, I only had one small remark. Which was easy enough to fix along with the changelog.

So thanks for the PR! This allowed me to release the fix today. 

Should be available once https://github.com/cucumber/cucumber-jvm/actions/runs/14960479934/job/42021698326 completes.

jean-tissot:
I saw your fix. Indeed, I didn't think of this to make bugs easier to find when the test fails.
Thanks for merging it!

Issue #3001, opened 2025-05-22
Switch builtin plugins to messages


mpkorstanje:
### 🤔 What's the problem you're trying to solve?

Cucumber-JVM currently has two event based protocols for communication with plugins. `io.cucumber:messages` and the objects included in `io.cucumber:cucumber-plugin`. With the latter no longer being developed it would be good to deprecate and removed.

### ✨ What's your proposed solution?

Update the following plugins to use `messages`

* [ ] JsonFormatter https://github.com/cucumber/cucumber-jvm/pull/2888
* [ ] PrettyFormatter https://github.com/cucumber/cucumber-jvm/pull/3012
* [ ] ProgressFormatter https://github.com/cucumber/cucumber-jvm/pull/3029
* [ ] RerunFormatter https://github.com/cucumber/cucumber-jvm/pull/3028
  * [ ]  Add pickle location to the Pickle message.
  * [ ] Use pickle location from pickle in all plugins
* [ ] TeamCityPlugin https://github.com/cucumber/cucumber-jvm/pull/3007
  * [ ]  Find a replacement for Suggested Snippets in the message protocol.
* [ ] TimelineFormatter 
* [ ] UnusedStepsSummaryPrinter
* [ ] UsageFormatter
* [ ] DefaultSummaryPrinter
  * [ ]  Find a replacement for Suggested Snippets in the message protocol.

PR #3028, opened 2025-07-17
Use messages in summary formatter

mpkorstanje:
### 🤔 What's changed?

Switch the summary and progress plugins to messages and extract to the [pretty-formatter module](https://github.com/cucumber/pretty-formatter).

Part of https://github.com/cucumber/cucumber-jvm/issues/3001

The summary is mostly a visual report. Does it look good? Can it be better?

### 📋 Checklist:

- [x] https://github.com/cucumber/query/pull/77


cucumber-core/src/main/java/io/cucumber/core/plugin/RerunFormatter.java
```diff
@@ -1,65 +1,48 @@
...
  * Formatter for reporting all failed test cases and print their locations
  * Failed means: results that make the exit code non-zero.
  */
 public final class RerunFormatter implements ConcurrentEventListener {
 
-    private final UTF8PrintWriter out;
-    private final Map<URI, Collection<Integer>> featureAndFailedLinesMapping = new LinkedHashMap<>();
+    private final PrintWriter writer;
+    private final Map<String, Set<Integer>> featureAndFailedLinesMapping = new LinkedHashMap<>();
+    private final Query query = new Query();
```
mpkorstanje, line 35: Using query here seems over kill but as pickles don't have their location build in, this is the easiest way to get it. Consider putting the pickle location in the message.

PR #3050, opened 2025-08-18
Use a message based TeamCity formatter

mpkorstanje:
### 🤔 What's changed?

Replace the `TeamCityPlugin` with the message based implementation from [teamcity-formatter](https://github.com/cucumber/teamcity-formatter/).

This achieves several goals:
 * Move the internal code base away from the events from the plugin module.
 * Extract common parts of Cucumber into modules that can be shared.

Partially implement: https://github.com/cucumber/cucumber-jvm/issues/3001

While the existing implementation and new implementation have sufficient test coverage, it is not impossible that something was structurally overlooked.

If you do find a problem after release please create a new issue.

### 📋 Checklist:

- [x] Make plugin concurrency aware
- [x] https://github.com/cucumber/query/pull/102


cucumber-core/src/main/java/io/cucumber/core/plugin/Plugins.java
```diff
@@ -98,6 +99,8 @@ private EventPublisher getOrderedEventPublisher(EventPublisher eventPublisher) {
     private static EventPublisher createCanonicalOrderEventPublisher(EventPublisher eventPublisher) {
         final CanonicalOrderEventPublisher canonicalOrderEventPublisher = new CanonicalOrderEventPublisher();
         eventPublisher.registerHandlerFor(Event.class, canonicalOrderEventPublisher::handle);
+        // Pass through for messages
+        eventPublisher.registerHandlerFor(Envelope.class, canonicalOrderEventPublisher::handle);
```
mpkorstanje, line 103: Not needed anymore

Issue #3010, opened 2025-06-05
High Overhead in FileFeatureOrigin Caused by Repeated Canonical File Resolution


lslonina:
### 🤔 What's the problem you're trying to solve?

In configurations with a significant number of tests, the test discovery time is noticeably long. In our project setup:

* 200+ feature files
* 6000+ containers
* 20000+ tests

**Runtime environment:** Windows (might behave differently on Linux or other Windows file systems)
**Java version:** 21

On development machines running Windows, it takes approximately **60–70 seconds** to discover features and start test execution. That is true also in case of filtering using tags.

Upon analysis, the slowdown seems to be caused by the `FileFeatureOrigin` class calling `FileSource`, which in turn spends most of the time in the `file.getCanonicalFile()` method:

* [[FeatureOrigin.java#L90](https://github.com/cucumber/cucumber-jvm/blob/731f3999958a5935df1309869c6507e40a686f9a/cucumber-junit-platform-engine/src/main/java/io/cucumber/junit/platform/engine/FeatureOrigin.java#L90)](https://github.com/cucumber/cucumber-jvm/blob/731f3999958a5935df1309869c6507e40a686f9a/cucumber-junit-platform-engine/src/main/java/io/cucumber/junit/platform/engine/FeatureOrigin.java#L90)
* [[FileSource.java#L73](https://github.com/junit-team/junit5/blob/7d02a6a2521617ab8502f8614794b7bb0a568603/junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FileSource.java#L73)](https://github.com/junit-team/junit5/blob/7d02a6a2521617ab8502f8614794b7bb0a568603/junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FileSource.java#L73)

---

### ✨ What's your proposed solution?

We tried two approaches to reduce the discovery time:

1. **Parallel creation of `FeatureDescriptions`**
   By first collecting all features and then creating `FeatureDescriptions` in parallel, we reduced processing time to **\~20 seconds**.
   Related code:
[...]

PR #3084, opened 2025-09-30
[JUnit Platform Engine] Use FileSource.withPosition

mpkorstanje:
### 🤔 What's changed?

Creating a `FileSource` involves a call to `File.getCanonicalFile`. This call can be slow because it involves the file system. Especially if the file system is networked.

By using `FileSource.withPosition` we can reuse the canonical file from the feature for its nodes.

Fixes: #3010

### 📋 Checklist:

### DIFF
--- a/cucumber-junit-platform-engine/pom.xml
+++ b/cucumber-junit-platform-engine/pom.xml
@@ -13,7 +13,7 @@
 
     <properties>
         <hamcrest.version>3.0</hamcrest.version>
-        <junit-jupiter.version>5.13.4</junit-jupiter.version>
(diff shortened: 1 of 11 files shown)


cucumber-junit-platform-engine/pom.xml
```diff
@@ -13,7 +13,7 @@
 
     <properties>
         <hamcrest.version>3.0</hamcrest.version>
-        <junit-jupiter.version>5.13.4</junit-jupiter.version>
+        <junit-jupiter.version>5.14.0-RC1</junit-jupiter.version>
```
mpkorstanje, line 16: Replace with non RC version.

PR #3102, opened 2025-10-26
Let JsonFormatter implement Serializer

laeubi:
Once in a while I need to write json output in some of my plugins as well but it is surprisingly hard to do so as the API to do so is all internal and non trivial to implement.

On the other hand the JsonFormatter already has everything we need here, so this lets JsonFormatter now implement Serializer to conviently support this use-case

@mpkorstanje what do you think? I just (again) encountered the problem that my custom hack for this broke the Eclipse-Cucumber plugin, having a semi official API for that would be great and it does not exposes to much of the internals but still allow to reliable write out the JSON (e.g. to a file or websocket) in custom plugins.

### DIFF
--- a/cucumber-core/src/main/java/io/cucumber/core/plugin/JsonFormatter.java
+++ b/cucumber-core/src/main/java/io/cucumber/core/plugin/JsonFormatter.java
@@ -1,6 +1,7 @@
 package io.cucumber.core.plugin;
 
 import io.cucumber.jsonformatter.MessagesToJsonWriter;
+import io.cucumber.messages.MessageToNdjsonWriter.Serializer;
 import io.cucumber.messages.types.Envelope;
 import io.cucumber.plugin.ConcurrentEventListener;
 import io.cucumber.plugin.event.EventPublisher;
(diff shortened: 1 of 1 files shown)


mpkorstanje:
Would this be sufficient?

```java
public final class ObjectToJsonWriter {
    
    @Override
    public void writeValue(Writer writer, Object value) throws IOException {
        Jackson.OBJECT_MAPPER.writeValue(writer, value);
    }
}
```

mpkorstanje:
@laeubi you can now use 

 * https://github.com/cucumber/messages-ndjson for serialization and deserialisation.

In conjunction with:

* https://github.com/cucumber/messages/blob/main/java/src/main/java/io/cucumber/messages/MessageToNdjsonWriter.java
* https://github.com/cucumber/messages/blob/main/java/src/main/java/io/cucumber/messages/NdjsonToMessageIterable.java
* https://github.com/cucumber/messages/blob/main/java/src/main/java/io/cucumber/messages/NdjsonToMessageReader.java
 
For the latter you'll have to pass `new Deserialiser()::readValue`.  Due to version constraints, that can't make it into `messages-ndjson` yet.

PR #3120, opened 2025-11-25
Register and filter glue classes

### LINKED ISSUE #3118: Add API to Manually Register Step Classes

When running Cucumber-JVM under GraalVM Native Image, the default glue discovery mechanism fails because it relies on classpath scanning. 
Native Image does not support directory traversal or dynamic classpath scanning, which results in all step definitions being reported as undefined. Even if the classes are preserved via `reflect-config.json`, Cucumber still cannot discover them unless a package path is provided to --glue. 
However, --glue only accepts package names, not explicit class names, so there is no way to bypass scanning and directly register known step classes. This makes it very difficult to run Cucumber tests in environments where reflection and scanning are restricted.

Introduce an API or configuration option that allows users to manually register step definition classes with the Cucumber runtime, without relying on package scanning. For example, a new CLI option or programmatic API such as:
```java
CucumberRuntime runtime = CucumberRuntime.builder()
    .addGlueClass(CommonSteps.class)
    .addGlueClass(SqlSteps.class)
    .addGlueClass(ExternalApiSteps.class)
    .build();
```
or a CLI flag like:
```shell
--glue-class jiuli.glue.steps.CommonSteps --glue-class jiuli.glue.steps.SqlSteps
```

This would let users explicitly list all step classes they want to include, ensuring compatibility with GraalVM Native Image and other restricted environments. It would also reduce reliance on fragile package scanning and make step registration more predictable.

Impl a owned `Backend` and replace cucumber's  `JavaBackend`.
It is not good to CI.

### 📚 Any additional context?

_No response_

### LINKED ISSUE #3137: Filter glue class candidates before loading

If I use the `Runtime.builder()` I can supply a classloader `withClassLoader` but the actuall class scanning can not be customized.

There should be a way to injects some kind of `GlueClassScanner` that allows to plug in different techniques.
Currently always `ClasspathScanner` is used with no way to customize that.

_No response_

### 📚 Any additional context?

_No response_

mpkorstanje:
### 🤔 What's changed?

Adds supports for registering individual glue classes via the `cucumber.glue.classes` property or the `--glue-classes` CLI option. 

Additionally, supports filtering glue classes by name when scanning the class-path via the `cucumber.glue.{included,excluded}-class-name-pattern` properties and `--glue-{included,excluded}-class-name-pattern` CLI options.

Only classes that are included by one of the include patterns and not excluded by any of the exclude patterns will be loaded during classpath scanning. This should speed up Cucumber significantly. The filters are not applied against classes that registered explicitly.

No include pattern is configured by default. Cucumber-JVM has been around for 14 years now and it would be quite complicated for existing implementations to either rename their glue classes to fit the existing pattern, or work out a pattern that matches existing classes.

To facilitate registering and filtering glue classes the `Backend` now has a new `loadGlue(Glue glue, GlueDiscoveryRequest request)` method. With the `GlueDiscoveryRequest` containing a collection of what exactly should be discovered. Currently only classes and URIs are supported but this can be trivially expanded to include packages, modules, ect.

Closes: https://github.com/cucumber/cucumber-jvm/issues/3118
Closes: https://github.com/cucumber/cucumber-jvm/issues/3137

### 📋 Checklist:


cucumber-core/src/main/java/io/cucumber/core/backend/Backend.java
```diff
@@ -15,7 +16,20 @@ public interface Backend {
...
+
+    /**
+     * Invoked once before all features. This is where steps and hooks should be
+     * loaded.
+     *
+     * @param glue           Glue that provides the steps to be executed.
+     * @param glueClassNames The classes of glue to be loaded.
+     */
+    default void loadGlueClasses(Glue glue, Set<String> glueClassNames) {
+        // TODO: Refactor out a request object.
```
mpkorstanje: TODO:
 * Refactor out a request object.

cucumber-junit-platform-engine/src/main/java/io/cucumber/junit/platform/engine/CucumberConfiguration.java
```diff
@@ -149,6 +150,16 @@ public List<URI> getGlue() {
                 .collect(Collectors.toList());
     }
 
+    @Override
+    public Set<String> getGlueClasses() {
```
mpkorstanje: Needs a test

cucumber-java/src/main/java/io/cucumber/java/JavaBackend.java
```diff
@@ -30,18 +32,37 @@ final class JavaBackend implements Backend {
 
     @Override
     public void loadGlue(Glue glue, List<URI> gluePaths) {
+        loadGlueClassesImpl(glue, scanForClasses(gluePaths));
+    }
+
+    @Override
+    public void loadGlueClasses(Glue glue, Set<String> glueClassNames) {
```
mpkorstanje: Needs a test

cucumber-core/src/main/java/io/cucumber/core/backend/discovery/GlueDiscoverySelectorResolver.java
```diff
@@ -0,0 +1,6 @@
+package io.cucumber.core.backend;
+
+public final class GlueDiscoverySelectorResolver {
```
mpkorstanje, line 19: WIP.

mpkorstanje:
@jkronegg any suggestions for a default class name pattern?

mpkorstanje:
@laeubi any suggestions for a default class name pattern?

laeubi:
Looking at the calculator example here

- https://github.com/cucumber/cucumber-jvm-examples/tree/main/calculator-java-cli/src/test/java/com/example/calculator

it seems `*StepDefinitions.class` seems a reasonable default.

mpkorstanje:
I wrote that example, so I'd be confirming my own biases if I used that. :smile:

jkronegg:
Intuitively, I would say `*StepDefs`. But by looking at the 10 Minute Cucumber Tutorial, it's more like `*StepDefinitions`.

And according to ChatGPT with the prompt "_In projects that use Java Cucumber BDD tests, what are the most common pattern for the class name which defines the steps ?_", it gives the following table:

|Pattern | Frequency |
| -- | --|
|`*Steps` | ##### |
|`*StepDefinitions` | ####_ |
|`*StepDef` | ##___ |
|`*StepDefs` | ##___ |
|`*Definitions` | #____|

I would prefer to use something the evokes "step" and "definition", because a step definition is not the same as a step. Thus `*StepDefinitions` or `*StepDefs` would be okay.

Maybe the default pattern could include multiple common patterns, e.g. `*StepDefinitions|*StepDef[s]`. That's a bit suboptimal, but then it gives other opportunities to tune the configuration for very specific cases.

Or as it's for v8, do you want by default to force users to write the steps definitions in files with a single file pattern ? In that case, it's fine using `*StepDefinitions`.

mpkorstanje:
@jkronegg how should I interpret your ChatGPT divination? It sounds like you asked it to confabulate.

mpkorstanje:
> Or as it's for v8, do you want by default to force users to write the steps definitions in files with a single file pattern ? In that case, it's fine using *StepDefinitions.

I don't know yet. Ideally nearly everything would fit some naming pattern. But it looks like this isn't the case even for Cucumber itself.

So we might just have to prompt people to configure it when scanning is slow.

jkronegg:
Regarding ChatGPT divination: I would not buy the list and frequency as-is, of course. We all know that the provided patterns and frequencies are invented, but it's a plausible answer, potentially correct, but potentially wrong. 
But it's the same when I wrote "_Thus `*StepDefinitions` or `*StepDefs` would be okay_.": it's a plausible answer, potentially correct, but potentially wrong. 😁😉

mpkorstanje:
Ah. I see. I was hoping you'd have a codebase you could grep through for glue class names. Or really anything that contains `io.cucumber` and provide a summary based of that.

jkronegg:
I built a Python script to build the frequency based on GitHub Search results (with GitHub Copilot help). These are the results:

```
 Other                  437  ( 43.7%)
 Steps                  329  ( 32.9%)
 StepDefinitions         61  (  6.1%)
 StepDef                 48  (  4.8%)
 StepDefs                43  (  4.3%)
 StepDefinition          37  (  3.7%)
 Step                    33  (  3.3%)
 Definitions              8  (  0.8%)
 StepsDefinitions         3  (  0.3%)
 Glue                     1  (  0.1%)
```

So to summarize:
- Variant of `Step` and plurals total about 36%
- Variants of `StepDefinition` and plurals total about 10%.
- Variants of `StepDef` and plurals total about 9%
- `Definitions` is about 1%

Note that these figures are sorted the same as the "hallucinated" ChatGPT answer. Maybe by chance? (or is this a confirmation biais?) 

Limitations:
- max 1000 results on the search API, so it's not very significative
- the repos quality is not guaranteed

See source code
[detectpatterns.py](https://github.com/user-attachments/files/30623766/detectpatterns.py)

When working on the repos from Cucumber user only, the results are a bit different:

```
StepDefinitions         23  ( 33.3%)
Steps                   21  ( 30.4%)
Other                   20  ( 29.0%)
StepDefs                 4  (  5.8%)
StepDef                  1  (  1.4%)
```

To pass the 1000 result limit, Copilot suggests to aggregate the result on a list of reference repos (which also guarantee the result quality). But then which repos need to be chosen?

So in the end, you end up with true number (not hallucinations), but still biased by the repos that are selected by you or by GitHub Search.... Thus I don't know whether we are in a better position or not to decide...😅

mpkorstanje:
> But then which repos need to be chosen?

Ideally? The ones you contribute to. Don't you have work repo's you can query? 😄

laeubi:
> > But then which repos need to be chosen?
> 
> Ideally? The ones you contribute to. Don't you have work repo's you can query? 😄

I think in the end one should choose whats reasonable / workable ... trying to fit everyone's need will likely fail and it is the projects privilege to decide and users can either adapt to such common standard or simply reconfigure.

I would more go the opposite: What should we choose that likely *not* matches *something else*, so `StepDefinitions` seems sound for me as its unlikely chosen for something else ([32 results on github](https://github.com/search?q=StepDefinitions&type=repositories)) while e.g. `Steps` seem to be generic enough to match something by accident ([98k results on github](https://github.com/search?q=Steps&type=repositories)).

mpkorstanje:
> I think in the end one should choose whats reasonable / workable ... trying to fit everyone's need will likely fail and it is the projects privilege to decide and users can either adapt to such common standard or simply reconfigure.

I'm certainly not going to be able to make everyone happy. I know that there is no perfect solution. But before picking a good enough default (if any) I'd like to have a better idea of what people call their step definition classes in the real world in projects they use.

Thinking about the problem, I'd like to do is come up with a set of conventional suffixes that cover nearly all use cases and then one extra suffix to catch everything else. For example `.*StepDefinitions`, `.*Hooks` and `.*Glue`. The first two would cover the steps and hooks, then the last one could be used for everything else - though I'm not a fan of the word glue, it is not a countable noun. 

We can then inform users that they can either configure the property or rename their classes to follow the conventional names.

> I would more go the opposite: What should we choose that likely not matches something else, so StepDefinitions seems sound for me as its unlikely chosen for something else ([32 results on github](https://github.com/search?q=StepDefinitions&type=repositories)) while e.g. Steps seem to be generic enough to match something by accident ([98k results on github](https://github.com/search?q=Steps&type=repositories)).

Good point.

jkronegg:
What about defining a single standard like `*StepDefinitions` in v8 default configuration, and provide an open rewrite recipe to migrate v7 to v8? This cooking recipe would find all classes that define steps in the user's project, and change the configuration accordingly.

cucumber-core/src/main/java/io/cucumber/core/options/Constants.java
```diff
@@ -131,6 +131,43 @@ public final class Constants {
...
+    /**
+     * Property name to set the pattern for excluded glue classes: {@value}
+     * <p>
+     * A glue class is included in classpath scanning if the include pattern
+     * matches it and the exclude pattern doesn't.
+     * <p>
+     * Example: {@code .*UnwantedStepDefinitions?|.*UnwantedHooks?}
+     */
+    @API(status = API.Status.EXPERIMENTAL, since = "8.0.0")
+    public static final String GLUE_EXCLUDED_CLASS_NAME_PATTERN_PROPERTY_NAME = "cucumber.glue.excluded-class-name-pattern";
```
mpkorstanje, line 170: 1. Because we're using a regex, we can't provide a comma separated list of patterns. 
2. Regexes are tricky, especially for the trivial patterns we intend to use.

It would make sense not to use regular expressions on the user interface but instead use a subset of the Ant Glob syntax. Here `*` would match `.+`. And the `.` should match `[.$]` to help deal with confusion around naming of nested classes.

laeubi: I think this is totally fine and users are often more confused by regexpression, simple pattern matching is enough ...

But the Example reads odd, maybe one can came up with something more relaistic, I think the most common case would be if StepDefinition would appear in the name of a domain class somewhere and one likely would then more use a full class name as it should be the exception that they share the _same package_ and _naming convention_.

mpkorstanje: Good point. And having done some research, I (re-)discovered that JUnit also matches with a regex when using `--include-classname`. 

https://docs.junit.org/6.1.3/running-tests/console-launcher.html

mpkorstanje:
> What about defining a single standard like *StepDefinitions in v8 default configuration, and provide an open rewrite recipe to migrate v7 to v8? This cooking recipe would find all classes that define steps in the user's project, and change the configuration accordingly.

That's a good possibility too.

mpkorstanje:
Having thought about the default include pattern problem.

Ideally when starting Cucumber JVM, all step definitions, hooks, ect would have been places in classes with a well known suffix. Unfortunately that ship sailed 14 years ago. 

Setting a default filter now would require that users either rename all their glue code to match the pattern or working out what the pattern should be. Renaming the glue code might be difficult. There seems to be a tendency for people to package step definitions rather APIs into libraries - I don't understand why, but here we are. Likewise working out what all the glue classes are will be difficult as it can be anything on the classpath.

Then keeping in mind that v8 will require Java 17+ I don't want to increase the migration burden any more than strictly necessary. 

Another angle to consider is the typical user of Cucumber JVM. While I reckon most people who've commented on this issue know their way around this isn't the case for every one using Cucumber. So it is sensible to ensure Cucumber works out of the box. Even if this is a little bit slower.

PR #3127, opened 2025-12-04
Reduce duplication between code and snippet generation

### LINKED ISSUE #3063: Reduce duplication between code and snippet generation
### 🤔 What's the problem you've observed?

Currently boths `GenerateI18n` classes for `cucumber-java[8]` and `SnippetGenerator` turn Gherkin keywords into code.  This code is now duplicated 3 times over. Each with a special exception for Emoji.

### ✨ Do you have a proposal for making it better?

Extract and reuse this code.

### 📚 Any additional context?

_No response_

ielatif:
Fixes #3063

As seen in https://github.com/cucumber/cucumber-jvm/pull/3062 currently boths GenerateI18n classes for cucumber-java[8] and SnippetGenerator turn Gherkin keywords into code. This code is now duplicated 3 times over. Each with a special exception for Emoji.

### DIFF
--- a/cucumber-core/src/main/java/io/cucumber/core/snippets/GherkinKeywordNormalizer.java
+++ b/cucumber-core/src/main/java/io/cucumber/core/snippets/GherkinKeywordNormalizer.java
@@ -30,10 +30,6 @@ public final class GherkinKeywordNormalizer {
         return language.replaceAll("[\\s-]", "_").toLowerCase();
     }
 
-    public static String capitalize(String str) {
-        return str.substring(0, 1).toUpperCase() + str.substring(1);
-    }
-
     private static String normalizeKeyword(String keyword) {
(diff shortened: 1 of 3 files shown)


cucumber-core/src/main/java/io/cucumber/core/codegen/GherkinKeywordNormalizer.java
```diff
@@ -0,0 +1,63 @@
+package io.cucumber.core.codegen;
```
ielatif: Introduced new package `io.cucumber.core.codegen`

ielatif:
Build fail with error : 

[ERROR] java.class.externalClassExposedInAPI: class io.cucumber.gherkin.GherkinDialect: A class from supplementary archives is used in a public capacity in the API. [io.cucumber:gherkin:jar:36.1.0]

@mpkorstanje  can you advice in which module should i put GherkinKeywordNormalizer class ?

Issue #2879, opened 2024-05-03
Support Provider instances with Pico Container


mpkorstanje:
### 🤔 What's the problem you're trying to solve?

Currently when using Cucumber with `cucumber-picocontainer` it is not possible to inject dependencies that that ultimately do not have a zero-argument constructor or require some setup. For example:

For example 

```java
public class MyStepDefintions {
  public MyStepDefinitions(WebDriver driver) {
    ...
  }
}
```

And it is also not possible to inject dependencies that partially depend on other inject dependencies and partially on some configuration. 

For example we may want to inject `PageObjects` that partially depend on other dependencies and partially on some configuration:

```java
SecuredPage securedPage = new SecuredPage(driver, project, "example", "top secret");
```

### ✨ What's your proposed solution?

Support Pico Containers `Providers` so that it becomes possible to write.

```java
public class WebDriverProvider implement Provider {

    public WebDriver provide() {
        return // create WebDriver instance here; 
    }
}
```

```java
public class SecuredPageProvider implements Provider {
    public SecuredPage provide(WebDriver driver, Project project) {
         return new SecuredPage(driver, project, "example", "top secret");
    }
}

### ⛏ Have you considered any alternatives or workarounds?

The current work around is to create a container object. But that quickly becomes ugly once multiple dependencies are involved.

### 📚 Any additional context?

 * https://stackoverflow.com/questions/78423694/how-to-better-handle-creation-of-required-pageobjects-when-implementing-loadable

**Comment by mpkorstanje** [comment]
On second thought. While the `Provider` interface is quite nice, there are some problems.

1. I don't really like the marker interface with magically named "provide" method. It is too easy to get it wrong.
2. It exposes Pico container details that were invisible in the past.
3. The provider must have a zero arg constructor. This is really old fashioned.
4. There can only be a single type created per provider. This creates lots of repetition.

So it might be nicer to declare a proper factory object:

```java
@Factory
public class WebDriverFactory {
   private final Project project;


   public WebDriverFactory(Project project) {
      this.project = project;
   }

   @Bean
   public WebDriver webDriver() {
       return // create WebDriver 
   }

   @Bean
    public SecuredPage provide(WebDriver driver) {
       return new SecuredPage(driver, project, "example", "top secret");
    }
}
```

But this isn't great either. Because now we have 2 annotations, to mark the factory class and methods involved. 

This could be simplified a bit by requiring the factory methods to be static.

```java
public class WebDriverFactory {


[...]

PR #3128, opened 2025-12-07
Support Provider instances with Pico Container

antagoony:
### 🤔 What's changed?

This is my approach to solve #2879.

Just saw the "help wanted" label, plus it sounds challenging ;-)

I read your thoughts and agree most of it. I do not totally follow the argument to "[...] declare a proper factory object". I think if any user decides to use PicoContainter he most likely wants to see it working as expected by this CI framework. Similarly Cucumber does not confuse Spring users. There is no new injection approach different to the Spring way of doing so.

Admittedly Cucumber-Spring does not automatically scan for all Spring components. Instead it scans exactly where the user wants it to do (by providing further Spring annotations on exactly that class that holds the `@CucumberContextConfiguration` annotation).

Following this idea, my PR comes up with something similar. It provides a `@PicoConfiguration` annotation that can be used to declare any `Provider` and/or `ProviderAdapter` classes that shall be used.

Not particular but in general -- what do you think of this approach?

### 📋 Checklist:

- [x] ~~I've changed the behaviour of the code~~ I've added new code

### DIFF
--- a/cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
+++ b/cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
@@ -65,15 +65,11 @@ public final class PicoFactory implements ObjectFactory {
         pico.start();
     }
 
-    static boolean hasCucumberPicoProvider(Class<?> clazz) {
-        return clazz.isAnnotationPresent(CucumberPicoProvider.class);
-    }
-
-    static boolean isProvider(Class<?> clazz) {
+    private static boolean isProvider(Class<?> clazz) {
         return Provider.class.isAssignableFrom(clazz);
     }
 
-    static boolean isProviderAdapter(Class<?> clazz) {
+    private static boolean isProviderAdapter(Class<?> clazz) {
         return ProviderAdapter.class.isAssignableFrom(clazz);
     }
 
@@ -97,7 +93,7 @@ public final class PicoFactory implements ObjectFactory {
     @Override
     public boolean addClass(Class<?> clazz) {
         if (isProvider(clazz)) {
-            providers.add(checkProperPicoProvider(clazz));
+            providers.add(requireConstructableProvider(clazz));
         } else {
             if (isInstantiable(clazz) && classes.add(clazz)) {
                 addConstructorDependencies(clazz);
@@ -116,7 +112,7 @@ public final class PicoFactory implements ObjectFactory {
     }
 
     @SuppressWarnings("unchecked")
-    private static Class<Provider> checkProperPicoProvider(Class<?> clazz) {
+    private static Class<Provider> requireConstructableProvider(Class<?> clazz) {
         if (!isProvider(clazz) || !isInstantiable(clazz) || !hasDefaultConstructor(clazz)) {
             throw new CucumberBackendException(String.format("" +
                     "Glue class %1$s was annotated with @CucumberPicoProvider; marking it as a candidate for declaring a"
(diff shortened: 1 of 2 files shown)


cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
```diff
@@ -31,8 +35,29 @@ public void start() {
...
+                }
+            }
+            for (Class<?> clazz : classes) {
+                // do not add the classes that represent a picocontainer
+                // ProviderAdapter/Provider, and also do not add those raw
+                // classes that are already provided (otherwise this causes
+                // exceptional situations, e.g. PicoCompositionException
+                // with message "Duplicate Keys not allowed. Duplicate for
+                // 'class XXX'")
+                if (!providers.contains(clazz) && !providedClasses.contains(clazz)) {
```
mpkorstanje: Logically okay. But please clean this up. There should be something in the streaming API that allows grouping by.

cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
```diff
@@ -31,34 +35,90 @@ public void start() {
...
+        if (pico.getLifecycleState().isStarted()) {
+            pico.stop();
+        }
         pico.dispose();
     }
 
     @Override
     public boolean addClass(Class<?> clazz) {
+        checkMeaningfulPicoAnnotation(clazz);
         if (isInstantiable(clazz) && classes.add(clazz)) {
```
mpkorstanje: It would make sense to split `@CucumberPicoProvider` annotated and other glue classes here. It would clean up the logic in `start` and prevent constructor dependencies being added to the container..

cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
```diff
@@ -31,34 +35,90 @@ public void start() {
...
     }
 
+    private static void checkMeaningfulPicoAnnotation(Class<?> clazz) {
+        if (clazz.isAnnotationPresent(CucumberPicoProvider.class)) {
+            CucumberPicoProvider annotation = clazz.getAnnotation(CucumberPicoProvider.class);
+            if (!isProvider(clazz) && (annotation.providers().length == 0)) {
+                throw new CucumberBackendException(String.format("" +
+                        "Glue class %1$s was annotated with @CucumberPicoProvider; marking it as a candidate for declaring "
+                        +
+                        "PicoContainer Provider classes. Please ensure that at least one the following requirements is satisfied:\n"
```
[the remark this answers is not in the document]
antagoony: The text has changed even more. There are two requirements now (implements `Provider`, has default constructor), and both must be satisfied.

cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
antagoony: There are two collections now, one for the `Provider` classes, one for the "normal" component classes.

cucumber-picocontainer/src/main/java/io/cucumber/picocontainer/PicoFactory.java
```diff
@@ -31,34 +35,90 @@ public void start() {
...
         if (isInstantiable(clazz) && classes.add(clazz)) {
             addConstructorDependencies(clazz);
         }
         return true;
     }
 
+    private static void checkMeaningfulPicoAnnotation(Class<?> clazz) {
+        if (clazz.isAnnotationPresent(CucumberPicoProvider.class)) {
+            CucumberPicoProvider annotation = clazz.getAnnotation(CucumberPicoProvider.class);
+            if (!isProvider(clazz) && (annotation.providers().length == 0)) {
```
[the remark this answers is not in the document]
antagoony: You are right, because that constructor is used when instantiating the `Provider` class within `PicoFactory#adapterForProviderClass(Class)`.

mpkorstanje:
Looks good at a glance. I'll probably merge this sometime next week pending availability!

mpkorstanje:
Cheers!

PR #3135, opened 2026-01-07
Ignore all class loading issues

laeubi:
In some rare circumstances it can happen that byte-codes are generated in "class" files (e.g. seems to be the case with groovy) that violates the contract of the java specification for some method names. In this case the ClasspathScanner fails with a ClassFormatError even though ClasspathScanner#safelyLoadClass likley seem to guard against classload errors.

To make this more robust and eve cover more cases (e.g. UnsatisfiedLinkError if the class uses native code with a static initializer or VerifyError if certificate verification fails) now the method catches all kind of LinkageError's to really safely load a class only if its possible.

### DIFF
--- a/cucumber-core/src/main/java/io/cucumber/core/resource/ClasspathScanner.java
+++ b/cucumber-core/src/main/java/io/cucumber/core/resource/ClasspathScanner.java
@@ -1,6 +1,5 @@
 package io.cucumber.core.resource;
 
-import io.cucumber.core.exception.UnrecoverableExceptions;
 import io.cucumber.core.logging.Logger;
 import io.cucumber.core.logging.LoggerFactory;
 


cucumber-core/src/main/java/io/cucumber/core/resource/ClasspathScanner.java
```diff
@@ -104,7 +104,7 @@ private Function<Path, Consumer<Path>> processClassFiles(
     private Optional<Class<?>> safelyLoadClass(String fqn) {
         try {
             return Optional.ofNullable(getClassLoader().loadClass(fqn));
-        } catch (ClassNotFoundException | NoClassDefFoundError e) {
+        } catch (ClassNotFoundException | LinkageError e) {
```
mpkorstanje: We might as well catch `Throwable` and use `rethrowIfUnrecoverable`.

laeubi: Whatever works best here!

PR #3139, opened 2026-01-14
Add Step info to @BeforeStep and @AfterStep hooks

### LINKED ISSUE #1805: Add Step information in @BeforeStep and @AfterStep hook

glaciousm:
## Summary

This PR implements the feature requested in #1805 - allowing `@BeforeStep` and `@AfterStep` hooks to receive step information (keyword, text, line number) alongside the existing `Scenario` parameter.

**New API:**
```java
@BeforeStep
public void beforeStep(Scenario scenario, Step step) {
    System.out.println("Running: " + step.getKeyword() + step.getText());
    System.out.println("At line: " + step.getLine());
}

@AfterStep  
public void afterStep(Scenario scenario, Step step) {
    if (scenario.isFailed()) {
        System.out.println("Failed step: " + step.getText());
    }
}
```

## Changes

- Add new `io.cucumber.java.Step` interface exposing step details
- Add internal `StepInfo` wrapper class implementing the `Step` interface
- Extend `HookDefinition` with `execute(state, step)` default method
- Update `JavaHookDefinition` to support 2-parameter step hooks
- Pass step info through `HookTestStep` execution chain

## Backward Compatibility

- Existing hooks with 0 or 1 parameter continue to work unchanged
- No breaking changes to the public API

## Test plan

- [x] All existing cucumber-core tests pass (665 tests)
- [x] All existing cucumber-java tests pass (115 tests)
- [x] Tested with a real Selenium/Cucumber project

Closes #1805


cucumber-core/src/main/java/io/cucumber/core/runner/HookTestStep.java
```diff
@@ -24,4 +43,105 @@ HookDefinitionMatch getDefinitionMatch() {
...
+     *                       hooks)
+     * @return               the next execution mode
+     */
+    ExecutionMode run(
+            TestCase testCase,
+            EventBus bus,
+            TestCaseState state,
+            ExecutionMode executionMode,
+            Step step
+    ) {
```
mpkorstanje: Duplicated. Please don't.

cucumber-java/src/main/java/io/cucumber/java/JavaHookDefinition.java
```diff
@@ -19,50 +19,82 @@ final class JavaHookDefinition extends AbstractGlueDefinition implements HookDef
...
 
-    private static Method requireValidMethod(Method method) {
+    private static Method requireValidMethod(HookType hookType, Method method) {
         Class<?>[] parameterTypes = method.getParameterTypes();
-        if (parameterTypes.length > 1) {
-            throw createInvalidSignatureException(method);
+        boolean isStepHook = hookType == HookType.BEFORE_STEP || hookType == HookType.AFTER_STEP;
+
+        // Step hooks allow 0, 1, or 2 parameters; other hooks allow 0 or 1
+        int maxParams = isStepHook ? 2 : 1;
```
mpkorstanje: This is clever. But it would be easier to understand if this method checked the hooktype and then applied the right validations/exceptions for each type.

cucumber-java/src/main/java/io/cucumber/java/Step.java
```diff
@@ -0,0 +1,50 @@
+package io.cucumber.java;
+
+import io.cucumber.plugin.event.StepArgument;
```
mpkorstanje: Note to self: Step definition users should never see the `io.cucumber.plugin` package. But then what should they see? 

* `io.cucumber.datatable.DataTable` / `io.cucumber.docstring.DocString`?
* A `io.cucumber.java.StepArgument` wrapper?

mpkorstanje (commented):
Overall this looks like a decent approach.

* The `HookTestStep::run` method is duplicated. I don't think that this is necessary. In `TestStep::executeStep` it should be possible to pass the step argument and propagate it down to `HookDefinition::execute` by adding an extra argument to `StepDefinitionMatch::runStep`. If it goes unused in other places, that is okay.

* `io.cucumber.plugin.event.StepArgument` shouldn't be visible to users but I don't know what to request that yet.

glaciousm:
@mpkorstanje  i did made some changes but i need your opinion here.. i really dont like the ,null in the step.run, i would prefer overloading but i was concerned you would see it as duplication. What is your preference? Overload the run or the ,null is fine?

glaciousm:
Added the requested documentation and tests:

  - Updated `cucumber-java/README.md` with documentation for the `Step` parameter in `@BeforeStep`/`@AfterStep` hooks
  - Added unit tests in `JavaHookDefinitionTest` covering the validation branches for step hooks
  - Added integration tests (`step-hooks.feature` + `StepHooksSteps.java`) to verify the feature works end-to-end

  Ready for review when you have time.

PR #3167, opened 2026-02-06
Memoize configuration parameters that are costly to create

henri-tremblay:
### 🤔 What's changed?

Remember the value of a configuration parameter to return the same result next time.

Configuration parameters like Pattern are really costly to create. And they are accessed repetitively for each scenario. In benchmarks generating tests to run without running them, it's more than 50% of the time.

If possible, I would like to see it included in the 7.x cycle but didn't know which branch to target.

I haven't tested memoization everywhere I used it because there's often more manipulation afterwards. If we go crazy, we could always keep the final returned result instead of the one after the compute. But I think that's a good start.

### 📋 Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Core] Update dependency io.cucumber:testng-xml-formatter.version to v0.8.0
 - [Core] Update dependency io.cucumber:usage-formatter.version to v0.2.0
 - [JUnit Platform Engine] Use JUnit Platform 6.0.2 (JUnit Jupiter 6.0.2) ([#3162](https://github.com/cucumber/cucumber-jvm/pull/3162))
+- [JUnit Platform Engine] Performance: Memoize configuration parameters ([#3167](https://github.com/cucumber/cucumber-jvm/pull/3167))
 - [All] Classes not designed for extension are now final. See [api-changes.json](./.revapi/api-changes.json) for details.
 - [All] Utility classes are no longer instantiatable. See [api-changes.json](./.revapi/api-changes.json) for details.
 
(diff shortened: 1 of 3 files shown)


mpkorstanje:
If there is a benefit to this then I'd prefer we copy JUnits design where there is [`CachingJupiterConfiguration`](https://github.com/junit-team/junit-framework/blob/main/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/config/CachingJupiterConfiguration.java) and [`DefaultJupiterConfiguration`](https://github.com/junit-team/junit-framework/blob/main/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/config/DefaultJupiterConfiguration.java) which both implement the same interface.

mpkorstanje:
> @mpkorstanje would it be feasible to have the same configuration logic between the junit platform and the Cucumber CLI execution modes ? (that would be probably much more work than just caching)

No. The `CucumberTestEngine` supports a different sets of properties and options  from the `Runtime` class.

henri-tremblay:
I'm trying to give you a benchmark. I had one for JUnit 4. Let's see for 5.

mars1589 (changes requested):
In plugins(), we now cache the parsed plugin list, then append the publish/no-publish plugin to that list. Since the base list is memoized, if plugins() is called more than once on the same CucumberConfiguration, we can append again to the same list and end up with duplicate plugin entries. So plugins() stops being idempotent.

Could we keep memoization for the expensive parse part, but still return a fresh mutable list per call before adding publish plugin? For example, copy the memoized base list into a new ArrayList and then append publish plugin to that new list.

Also, can we add a regression test that calls plugins() twice on the same config instance (with PLUGIN_PROPERTY_NAME and publish token enabled) and asserts the second call does not increase count or duplicate entries?

mpkorstanje:
I've done some thinking about the architecture (see my rather messy notes in https://github.com/cucumber/cucumber-jvm/issues/3240). I think we can revisit this whole thing once that crystalizes a bit more.

Either way, for the PR!

PR #3183, opened 2026-05-11
Support `cucumber.execution.threads` property when executing from CLI

### LINKED ISSUE #2833: cucumber.properties missing support for threads value
`io.cucumber.core.options.CucumberPropertiesParser` pulls in many of the required arguments from various sources e.g. a properties file, environment or system properties

However, it appears to not be looking for number of threads to use (`--threads` via CLI) - was this intentional?

### 👓 What did you see?

Having setup multiple scenarios, and having added `cucumber.execution.threads=2` to properties file I expected tests to run in parallel

`cucumber.execution.threads` should be parsed and utilised

cucumber-core 7.14.1

Create multiple scenarios
Create `cucumber.properties` with the normal required values (e.g. glue, features and plugin)
Ensure within the properties file you add `cucumber.exeution.threads=2`
Run cucumber

You will see all tests are executed synchronously and the timeline report also confirms this

charliesheh:
Fixes #2833.

Adds support for configuring the number of execution threads through cucumber.properties using cucumber.execution.threads.

Validated with:
mvn -pl cucumber-core -am test -Dtest=CucumberPropertiesParserTest -Dsurefire.failIfNoSpecifiedTests=false


cucumber-core/src/main/java/io/cucumber/core/options/Constants.java
```diff
@@ -47,6 +47,11 @@ public final class Constants {
      */
     public static final String EXECUTION_ORDER_PROPERTY_NAME = "cucumber.execution.order";
 
+    /**
+     * Property name used to set the number of threads.
+     */
+    public static final String EXECUTION_THREADS_PROPERTY_NAME = "cucumber.execution.threads";
```
mpkorstanje, line 61: Please note https://github.com/cucumber/cucumber-jvm/issues/2833#issuecomment-1856357326, the documentation should explain that this property is only effective when the CLI is used. And unfortunately the documentation should be updated in a few places.

cucumber-core/src/main/java/io/cucumber/core/options/CucumberPropertiesParser.java
```diff
@@ -138,6 +139,11 @@ public RuntimeOptionsBuilder parse(CucumberPropertiesProvider properties) {
             BooleanString::parseBoolean,
             builder::setWip);
 
+        parse(properties,
+            EXECUTION_THREADS_PROPERTY_NAME,
```
mpkorstanje: For readability, properties are parsed in lexicographical order. Please move this call after the one for `EXECUTION_ORDER_PROPERTY_NAME`.

cucumber-core/src/main/java/io/cucumber/core/options/CucumberPropertiesParser.java
```diff
@@ -168,7 +176,8 @@ private <T> void parseAll(
             Collection<T> parsed = parser.apply(property);
             parsed.forEach(setter);
         } catch (Exception e) {
-            throw new CucumberException("Failed to parse '" + propertyName + "' with value '" + property + "'", e);
+            throw new CucumberException(
+                "Failed to parse '" + propertyName + "' with value '" + property + "'", e);
```
mpkorstanje: The above lines appear to have changed without reason.

mpkorstanje (changes requested):
Cheers! Implementation wise this look okay. I've left a few comments below.  It does need more  documentation.

In addition to the Javadoc, please add documentation to:
 * https://github.com/cucumber/cucumber-jvm/blob/main/cucumber-core/README.md
 * https://github.com/cucumber/cucumber-jvm/blob/main/cucumber-core/src/main/resources/io/cucumber/core/options/USAGE.txt

And finally add an entry to the changelog:
* https://github.com/cucumber/cucumber-jvm/blob/main/CHANGELOG.md

PR #3196, opened 2026-05-25
Improve DuplicateTypeNameException with location details

### LINKED ISSUE #3144: Give more details with DuplicateTypeNameException

_No response_

### 📚 Any additional context?

_No response_

wyf027:
## Summary

- Add `DuplicateParameterTypeException` mirroring the pattern used by `DuplicateStepDefinitionException`
- Detect duplicate user-defined parameter types before registration and report both locations
- Wrap built-in name conflicts from `cucumber-expressions` with the offending custom definition location

Fixes #3144

## Problem

`DuplicateTypeNameException` only reported that a name already exists (e.g. `There is already a parameter type with name iso8601Date`) without indicating where the duplicate definitions are, making misconfigured glue code hard to locate.

## Example messages

- User-defined duplicate: `Duplicate parameter type 'iso8601Date' in first location and second location`
- Built-in conflict: `Parameter type 'int' at custom int location conflicts with an existing parameter type of the same name`

## Test plan

- [x] `mvn -pl cucumber-core test -Dtest=DuplicateParameterTypeExceptionTest,CachingGlueTest`

### DIFF
--- a/cucumber-core/src/main/java/io/cucumber/core/runner/CachingGlue.java
+++ b/cucumber-core/src/main/java/io/cucumber/core/runner/CachingGlue.java
@@ -20,6 +20,7 @@ import io.cucumber.core.stepexpression.StepExpression;
 import io.cucumber.core.stepexpression.StepExpressionFactory;
 import io.cucumber.core.stepexpression.StepTypeRegistry;
 import io.cucumber.cucumberexpressions.CucumberExpression;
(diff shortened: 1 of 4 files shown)


mpkorstanje:
Thanks for looking at this. Unfortunately it looks like this PR was created without much thought, so I'll dismiss it without much thought.

PR #3208, opened 2026-06-29
Add test for nested Cucumber engine discovery under the suite engine

vasiliy-mikhailov:
Adds one test to `CucumberTestEngineTest` for the case where the Cucumber engine runs nested under the JUnit Platform Suite engine with `discovery-as-root-engine=false`.

The existing tests exercise discovery when Cucumber is the root engine, where `isRootEngine()` is effectively always true. The nested-under-suite case (a UniqueId with more than two segments, so `isRootEngine()` returns false) was not covered, so a regression making `isRootEngine()` always return true would still pass. The new test uses `EngineTestKit` to discover under `junit-platform-suite` with the property disabled and asserts the Cucumber engine still finds its features.

Additive only: +1 test, no production changes.


mpkorstanje:
Screenshot From 2026-06-29 08-42-14

PR #3210, opened 2026-06-30
[Core] Don't swallow event handler exceptions while processing a feature

### LINKED ISSUE #2748: Exceptions are swallowed (or not) depending on the `Envelope` content
### 👓 What did you see?

Let's say I have a plugin which process `Envelope` and may generate exceptions. Depending on the `Envelope` content, the exceptions are rethrowed or swallowed by Cucumber:

```java
public class MyPluginPlugin implements EventListener, Plugin {
    private final EventHandler<Envelope> testParameterTypeHandler = (event) -> {
        if (event.getParameterType().isPresent()) {
            throw new RuntimeException("getParameterType");// will be swallowed
        } else if (event.getStepDefinition().isPresent()) {
            throw new RuntimeException("getStepDefinition");// will be swallowed
        } else if (event.getGherkinDocument().isPresent()) {
            throw new RuntimeException("getGherkinDocument");// will be swallowed
        } else if (event.getTestCaseStarted().isPresent()) {
            throw new RuntimeException("getTestCaseStarted");// will be swallowed
        } else if (event.getTestRunFinished().isPresent()) {
            //throw new RuntimeException("getTestRunFinished"); // not swallowed
        }
    };

    public void setEventPublisher(EventPublisher eventPublisher) {
        eventPublisher.registerHandlerFor(Envelope.class, this.testParameterTypeHandler);
    }

}
```

Note: this is not the real plugin, just a minimalistic one designed to show the problem.

I expect the exceptions generated by the plugin are all rethrown and not swallowed.

Cucumber Java 7.12.0

1. create a plugin with the code above
2. add the plugin to the run configuration of any project which uses Cucumber
3. run the test with the plugin
4. no exception is raised

### 📚 Any additional context?

As a workaround, I currently keep the generated exception aside and throw it when the `Envelope` contains a `TestRunFinished` event: 

```java
public class MyPluginPlugin implements EventListener, Plugin {
    RuntimeException runtimeException = null;

    private final EventHandler<Envelope> testParameterTypeHandler = (event) -> {
        if (event.getParameterType().isPresent()) {
            runtimeException = new RuntimeException("getParameterType");// will be swallowed
        } else if (event.getStepDefinition().isPresent()) {
            runtimeException = new RuntimeException("getStepDefinition");// will be swallowed
        } else if (event.getGherkinDocument().isPresent()) {
            runtimeException = new RuntimeException("getGherkinDocument");// will be swallowed
        } else if (event.getTestCaseStarted().isPresent()) {
            runtimeException = new RuntimeException("getTestCaseStarted");// will be swallowed
        } else if (event.getTestRunFinished().isPresent()) {
            //throw new RuntimeException("getTestRunFinished"); // not swallowed
            if (runtimeException != null) {
                throw runtimeException;
            }
        }
    };

    public void setEventPublisher(EventPublisher eventPublisher) {
        eventPublisher.registerHandlerFor(Envelope.class, this.testParameterTypeHandler);
    }

}
```

nikolauspschuetz:
## Summary

Exceptions thrown by an `EventListener` / event handler were **swallowed for some `Envelope` payloads but rethrown for others**, exactly as reported in #2748. A plugin that throws while handling, e.g., the `GherkinDocument` / `Source` / `Pickle` envelopes had its exception silently dropped, while the same plugin throwing on the `TestRunFinished` envelope failed the run. The behavior is now consistent: a failing handler always propagates.

## Root cause

`io.cucumber.core.runtime.CucumberExecutionContext.runFeatures(...)` runs feature setup inside `execute(...)`:

```java
private void execute(ThrowingRunnable runnable) {
    try {
        runnable.run();
    } catch (Throwable t) {
        // Collected in CucumberExecutionContext
        rethrowIfUnrecoverable(t);   // <-- only unrecoverable rethrown; everything else dropped
    }
}
```

The comment claims the throwable is "Collected", but `beforeFeature(...)` — which emits the `GherkinDocument`/`Source`/`Pickle` envelopes via `bus.sendAll(feature.getParseEvents())` — did **not** route through the `RethrowingThrowableCollector`. So a handler exception thrown there reached `execute(...)` and was silently dropped.

By contrast, `TestRunFinished` is emitted from `finishTestRun()` in the `finally` block, **outside** the `execute(...)` boundary, so its handler exceptions always propagated. That is the asymmetry (`CucumberExecutionContext.java`, `beforeFeature` around L124 and `execute` around L157).

The test-case, before/after-all-hook and runner paths already collect via `collector.executeAndThrow(...)`; only `beforeFeature` was missing.

## Fix

Route `beforeFeature(...)` through the existing `RethrowingThrowableCollector` so a failing handler is collected and rethrown at the end of the run, matching every other emission path.

```java
collector.executeAndThrow(() -> {
    bus.send(new TestSourceRead(...));
    bus.send(new TestSourceParsed(...));
    bus.sendAll(feature.getParseEvents());
});
```

Minimal change; the integration callers (JUnit 4, TestNG, JUnit Platform engine) call `beforeFeature` directly and already rely on immediate rethrow, so their behavior is unchanged.

## Test evidence

Added a regression test in `CucumberExecutionContextTest` that registers an `Envelope` handler throwing on the `GherkinDocument` envelope and asserts the exception propagates.

- **Without the fix:** `Expected java.lang.IllegalStateException to be thrown, but nothing was thrown.` (exception swallowed — reproduces the bug)
- **With the fix:** passes.

Verified locally (JDK 17 source level; build run with a JDK 21 toolchain because the project's checkstyle tooling requires it):

```
CucumberExecutionContextTest: Tests run: 4, Failures: 0, Errors: 0, Skipped: 0
cucumber-core (full):         Tests run: 659, Failures: 0, Errors: 0, Skipped: 69
cucumber-junit-platform-engine (full): Tests run: 204, Failures: 0, Errors: 0
```

`spotless:check` passes.

Closes #2748

---
Assisted-by: Claude Code (Anthropic, Opus 4.x)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 
 ### Fixed
+- [Core] Don't swallow exceptions thrown by event handlers while processing a feature ([#2748](https://github.com/cucumber/cucumber-jvm/issues/2748), Julien Kronegg)
 - [JUnit Platform Engine] Accept partial matches with `cucumber.filter.name` and align the behavior with JUnit 4 and CLI ([#3174](https://github.com/cucumber/cucumber-jvm/pull/3174))
 - [JUnit Platform Engine] Don't require global read lock ([#3103](https://github.com/cucumber/cucumber-jvm/pull/3103))
 
--- a/cucumber-core/src/main/java/io/cucumber/core/runtime/CucumberExecutionContext.java
+++ b/cucumber-core/src/main/java/io/cucumber/core/runtime/CucumberExecutionContext.java
@@ -123,9 +123,14 @@ public final class CucumberExecutionContext {
 
     public void beforeFeature(Feature feature) {
         log.debug(() -> "Sending test source read event for " + feature.getUri());
-        bus.send(new TestSourceRead(bus.getInstant(), feature.getUri(), feature.getSource()));
-        bus.send(new TestSourceParsed(bus.getInstant(), feature.getUri(), singletonList(feature)));
-        bus.sendAll(feature.getParseEvents());
+        // Collect and rethrow any failure (e.g. from an event handler in a
+        // plugin). Otherwise the failure is swallowed when running the features
+        // because CucumberExecutionContext.execute drops it (see #2748).
+        collector.executeAndThrow(() -> {
+            bus.send(new TestSourceRead(bus.getInstant(), feature.getUri(), feature.getSource()));
+            bus.send(new TestSourceParsed(bus.getInstant(), feature.getUri(), singletonList(feature)));
+            bus.sendAll(feature.getParseEvents());
+        });
     }
 
     public void runTestCase(Consumer<Runner> execution) {
--- a/cucumber-core/src/test/java/io/cucumber/core/runtime/CucumberExecutionContextTest.java
+++ b/cucumber-core/src/test/java/io/cucumber/core/runtime/CucumberExecutionContextTest.java
@@ -1,6 +1,8 @@
 package io.cucumber.core.runtime;
 
 import io.cucumber.core.eventbus.EventBus;
+import io.cucumber.core.feature.TestFeatureParser;
+import io.cucumber.core.gherkin.Feature;
 import io.cucumber.core.options.RuntimeOptions;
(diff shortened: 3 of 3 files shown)


mpkorstanje:
I doubt you've spend any time to understand what you're doing.

Screenshot From 2026-06-30 13-35-34

PR #3216, opened 2026-07-24
[JUnit Platform Engine] Do not fail discovery on an empty rerun file

### LINKED ISSUE #3215: FeatureWithLinesFileResolver fails discovery on an empty @SelectFile .txt with JUnit Platform 6.1.x ("selectors must not be empty")
### 👓 What did you see?

Discovering a JUnit Platform `@Suite` that selects an **empty** `.txt` file via `@SelectFile` throws a critical discovery issue on JUnit Platform **6.1.x**:

```
TestEngine with ID 'junit-platform-suite' encountered a critical issue during test discovery:

(1) [ERROR] [cucumber] FileSelector [path = 'target/cucumber-rerun/Example.txt', position = null] resolution failed.
    Source: FileSource [file = .../target/cucumber-rerun/Example.txt, filePosition = null]
    Cause: org.junit.platform.commons.PreconditionViolationException: selectors must not be empty
        at org.junit.platform.commons.util.Preconditions.notEmpty(Preconditions.java:155)
        at org.junit.platform.engine.support.discovery.SelectorResolver$Resolution.selectors(SelectorResolver.java:506)
        at io.cucumber.junit.platform.engine.FeatureWithLinesFileResolver.resolve(FeatureWithLinesFileResolver.java:30)
        ...
        at io.cucumber.junit.platform.engine.CucumberTestEngine.discover(CucumberTestEngine.java:65)
```

`FeatureWithLinesFileResolver.resolve(...)` passes the result of `parseFile(path)` straight into `Resolution.selectors(...)`. When the `.txt` file is empty, that set is empty, and JUnit Platform 6.1 now rejects an empty selector set with a `PreconditionViolationException`.

This is the standard rerun-on-failure workflow: a rerun suite selects the file the previous run wrote (`rerun:target/cucumber-rerun/Example.txt`), and when the initial run had no failures cucumber writes an **empty** rerun file. Because the exception is raised during discovery inside `CucumberTestEngine`, it becomes a *critical discovery issue* that fails the engine/suite and cannot be suppressed with `@Suite(failIfNoTests = false)`.

Discovery resolves to zero tests without error — an empty rerun file simply means "nothing to re-run" — as on JUnit Platform 6.0.x. The suite uses `@Suite(failIfNoTests = false)` to tolerate this, but it has no effect since the failure happens during discovery, before that check applies.

- `io.cucumber:cucumber-junit-platform-engine` 7.34.3 (the affected code is byte-for-byte identical on `main` and in the latest release 7.34.5)
- JUnit Platform / Jupiter 6.1.2 — also reproduces on 6.1.0 and 6.1.1; 6.0.x does not fail
- Java 25, Maven Failsafe 3.5.6

1. Use `cucumber-junit-platform-engine` 7.34.x on JUnit Platform 6.1.x.
2. Create an empty (0-byte) file `rerun.txt` (this is what the `rerun:` plugin writes when the initial run has no failures).
3. Add a suite that selects it:
   ```java
   import org.junit.platform.suite.api.SelectFile;
   import org.junit.platform.suite.api.Suite;

   @Suite(failIfNoTests = false)
   @SelectFile("rerun.txt")
   class TextRunner {
   }
   ```
4. Run the suite (IDE, Failsafe, or any launcher) → discovery fails with `PreconditionViolationException: selectors must not be empty` (the `failIfNoTests = false` does not prevent it).

Switching the JUnit Platform version back to 6.0.x makes the same setup discover 0 tests and pass.

### 📚 Any additional context?

Root cause: JUnit Platform 6.1 tightened `SelectorResolver.Resolution.selectors(...)` to reject an empty collection (`Preconditions.notEmpty`). Cucumber hands it an empty set for an empty file:

```java
// io.cucumber.junit.platform.engine.FeatureWithLinesFileResolver
Set<FeatureWithLinesSelector> selectors = parseFile(path)   // empty file -> empty set
        .stream()
        .map(FeatureWithLinesSelector::from)
        .collect(Collectors.toSet());
return Resolution.selectors(selectors);                     // throws on 6.1.x when empty
```

Suggested fix — return `Resolution.unresolved()` when there is nothing to select:

```diff
 Set<FeatureWithLinesSelector> selectors = parseFile(path)
         .stream()
         .map(FeatureWithLinesSelector::from)
         .collect(Collectors.toSet());
-return Resolution.selectors(selectors);
+return selectors.isEmpty()
+        ? Resolution.unresolved()
+        : Resolution.selectors(selectors);
```

Verified locally: patching this single line makes an empty-rerun-file suite discover 0 tests and pass on both JUnit Platform 6.0.3 and 6.1.2, with no change to the non-empty case. Happy to open a PR.

deivydasp:
### 🤔 What's changed?

`FeatureWithLinesFileResolver` now returns `Resolution.unresolved()` when the selected `.txt` rerun file yields no selectors, instead of passing an empty set to `Resolution.selectors(...)`.

Previously, selecting an **empty** rerun file (via `@SelectFile`) threw a critical discovery issue on JUnit Platform 6.1.x:

```
Cause: org.junit.platform.commons.PreconditionViolationException: selectors must not be empty
    at org.junit.platform.engine.support.discovery.SelectorResolver$Resolution.selectors(...)
    at io.cucumber.junit.platform.engine.FeatureWithLinesFileResolver.resolve(...)
```

JUnit Platform 6.1 tightened `Resolution.selectors(...)` to reject an empty collection; returning `unresolved()` restores discovering zero tests for the empty case and leaves the non-empty case unchanged. This is the standard rerun-on-failure flow: when the initial run has no failures, cucumber's `rerun:` plugin writes an empty file, and the `*ITRerun` suite that selects it then fails discovery (which `@Suite(failIfNoTests = false)` cannot suppress, since it happens during discovery).

Added `supportsEmptyRerunFile()` (modeled on the existing `supportsRerunFile()` test) plus an empty `src/test/resources/rerun/empty-rerun.txt` resource.

Fixes #3215.

On this branch the bundled test passes with or without the fix, because v7.x.x pins JUnit Platform 1.14.2 (via junit-bom 5.14.2), which predates the `selectors must not be empty` precondition. I kept the test as a behavioural spec of the empty-rerun case (per the request in #3215 to base it on the existing tests). Happy to adjust if you'd prefer it placed/verified differently.

### 📋 Checklist:

  - [x] I have added an entry to the "Unreleased" section of the CHANGELOG, linking to this pull request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- [JUnit Platform Engine] Do not fail discovery on an empty rerun file with JUnit Platform 6.1.x ([#3216](https://github.com/cucumber/cucumber-jvm/pull/3216))
 
 ## [7.34.5] - 2026-07-23
 ### Fixed
--- a/cucumber-junit-platform-engine/src/main/java/io/cucumber/junit/platform/engine/FeatureWithLinesFileResolver.java
+++ b/cucumber-junit-platform-engine/src/main/java/io/cucumber/junit/platform/engine/FeatureWithLinesFileResolver.java
@@ -27,6 +27,8 @@ final class FeatureWithLinesFileResolver implements SelectorResolver {
                 .stream()
                 .map(FeatureWithLinesSelector::from)
                 .collect(Collectors.toSet());
-        return Resolution.selectors(selectors);
+        return selectors.isEmpty()
+                ? Resolution.unresolved()
+                : Resolution.selectors(selectors);
     }
 }
--- a/cucumber-junit-platform-engine/src/test/java/io/cucumber/junit/platform/engine/CucumberTestEngineTest.java
+++ b/cucumber-junit-platform-engine/src/test/java/io/cucumber/junit/platform/engine/CucumberTestEngineTest.java
@@ -1126,13 +1126,12 @@ class CucumberTestEngineTest {
 
     @Test
     void supportsEmptyRerunFile() {
-        EngineTestKit.engine(ENGINE_ID)
+        EngineDiscoveryResults result = EngineTestKit.engine(ENGINE_ID)
                 .selectors(
                     selectFile("src/test/resources/rerun/empty-rerun.txt"))
-                .execute()
-                .allEvents()
-                .assertThatEvents()
-                .haveExactly(0, event(test()));
+                .discover();
+
+        assertThat(result.getDiscoveryIssues()).isEmpty();
     }
 
 }
--- a/cucumber-junit-platform-engine/src/test/java/io/cucumber/junit/platform/engine/CucumberTestEngineTest.java
+++ b/cucumber-junit-platform-engine/src/test/java/io/cucumber/junit/platform/engine/CucumberTestEngineTest.java
@@ -1124,4 +1124,15 @@ class CucumberTestEngineTest {
                 .haveExactly(2, event(scenario("scenario:3", "A single scenario")));
     }
 
+    @Test
+    void supportsEmptyRerunFile() {
+        EngineTestKit.engine(ENGINE_ID)
+                .selectors(
+                    selectFile("src/test/resources/rerun/empty-rerun.txt"))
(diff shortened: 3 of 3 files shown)


mpkorstanje (changes requested):
Looks good. But I guess that's because you're using generative AI.
The test you've written doesn't fail when used with JUnit 6.1. I.e. you didn't write a failing test before you wrote the implementation.

PR #3225, opened 2026-08-08
Improve DuplicateTypeNameException with glue locations

### LINKED ISSUE #3144: Give more details with DuplicateTypeNameException

arimu1:
### 🤔 What's changing?

When a `DuplicateTypeNameException` is raised (two parameter types share a name, or a custom type reuses a built-in name), the message now includes the **glue location(s)** of the conflicting definitions.

Cucumber Expressions is location-agnostic, so this is handled in `CachingGlue` by tracking user-defined `ParameterTypeDefinition`s while preparing glue, then throwing `DuplicateParameterTypeException` (same pattern as `DuplicateStepDefinitionException`).

**Example messages:**

- User-defined duplicate: `Duplicate parameter type 'iso8601Date' in first location and second location`
- Built-in conflict: `Duplicate parameter type 'int' in custom int location (conflicts with a built-in parameter type)` (original `DuplicateTypeNameException` kept as cause)

### ✨ Why?

Fixes #3144 — the bare `There is already a parameter type with name …` message did not say *where* the definitions came from or *what* to remove.

### 🧹 Checklist
- [x] Tests for both user/user and user/built-in collisions (`CachingGlueTest`, `DuplicateParameterTypeExceptionTest`)
- [x] No public API change (package-private exception, same failure path during glue prepare)

### Test plan
- [x] `mvn -pl cucumber-core test -Dtest=DuplicateParameterTypeExceptionTest,CachingGlueTest` (Temurin 21) — **28/28**

### DIFF
--- a/cucumber-core/src/main/java/io/cucumber/core/runner/CachingGlue.java
+++ b/cucumber-core/src/main/java/io/cucumber/core/runner/CachingGlue.java
@@ -20,6 +20,7 @@ import io.cucumber.core.stepexpression.StepExpression;
 import io.cucumber.core.stepexpression.StepExpressionFactory;
 import io.cucumber.core.stepexpression.StepTypeRegistry;
 import io.cucumber.cucumberexpressions.CucumberExpression;
+import io.cucumber.cucumberexpressions.DuplicateTypeNameException;
 import io.cucumber.cucumberexpressions.Expression;
 import io.cucumber.cucumberexpressions.ParameterByTypeTransformer;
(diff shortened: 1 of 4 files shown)


mpkorstanje:
Please give the contribution guidelines wrt AI a read.

https://github.com/cucumber/.github/blob/main/CONTRIBUTING.md#ai-usage

mpkorstanje (commented):
The solution looks weird. Given that you've used AI, I don't think it's worth the effort to explain why or how to improve it.

Issue #3054, opened 2025-08-28
Emit messages for @BeforeAll and @AfterAll hooks


mpkorstanje:
### 🤔 What's the problem you're trying to solve?

Resolve https://github.com/cucumber/common/issues/2273

PR #3226, opened 2026-08-08
fix(core): exit CLI after BeforeAll/AfterAll failures

### LINKED ISSUE #3104: Test execution does not stop after throwing exception from @AfterAll
### 👓 What did you see?

When an exception is thrown from the method annotated with @AfterAll hook, then test execution does not stop. It waits indefinitely. I am seeing below behaviour.

Image

I have written a custom logic to check the code coverage and I want to run this logic when all feature files have been run. I am using @AfterAll hook for this. When the coverage is not desired then I want to throw an AssertionError with a message. After this the test execution should stop and the error should be shown. However the same is not happening. I cant figure out what it is waiting for. I tried to debug in the classes Runtime.Java and CucumberExecutionContext.java but couldnt find anything.

Java 21
cucumber bom 7.29.0

_No response_

### 📚 Any additional context?

_No response_

arimu1:
### 🤔 What's changed?

When `@BeforeAll` / `@AfterAll` hooks throw, `Runtime.run()` rethrows so the failure is visible (these hooks are not yet fully reportable; see #3054). That rethrow previously skipped `Main.main`'s `System.exit`, so any remaining non-daemon threads (for example a non-daemon `Timer`) kept the JVM alive and the process appeared to hang.

`Main.execute` now catches recoverable throwables from `Runtime.run`, prints the stack trace to stderr, and returns exit status `1`, so `System.exit` always runs for the CLI entry point.

Fixes #3104

Maintainer repro from the issue:

```java
new Timer(false).scheduleAtFixedRate(...); // non-daemon
@AfterAll
public static void afterAll() {
    throw new RuntimeException();
}
```

This is intentionally a CLI shutdown fix, not the full #3054 work (emit messages for static hooks). Embedders that call `Runtime.run()` directly still see the rethrown exception. Happy to adjust if you'd rather wait for #3054 only.

### 📋 Checklist:

  - [x] I have added an entry to the "Unreleased" section of the [CHANGELOG](../CHANGELOG.md), linking to this pull request.

### AI assistance disclosure

This PR was developed with assistance from AI coding tools (Grok / Cursor harness), under human direction:
- Issue analysis, root-cause identification, and test design were human-led.
- Implementation and PR description were drafted with AI assistance and then reviewed/edited by a human before submission.
- I understand the change: `Main.execute` catches recoverable throwables from `Runtime.run` so CLI always reaches `System.exit` after BeforeAll/AfterAll failures (non-daemon threads no longer hang the process). Full static-hook messaging remains #3054 follow-up.

Per https://github.com/cucumber/.github/blob/main/CONTRIBUTING.md#ai-usage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Core] Support `cucumber.execution.threads` property when executing from CLI ([#3183](https://github.com/cucumber/cucumber-jvm/pull/3183))
 
 ### Fixed
+- [Core] CLI always calls `System.exit` after `@BeforeAll` / `@AfterAll` failures so non-daemon threads cannot hang the JVM ([#3104](https://github.com/cucumber/cucumber-jvm/issues/3104), [#3226](https://github.com/cucumber/cucumber-jvm/pull/3226))
 - [JUnit Platform Engine] Accept partial matches with `cucumber.filter.name` and align the behavior with JUnit 4 and CLI ([#3174](https://github.com/cucumber/cucumber-jvm/pull/3174))
 - [JUnit Platform Engine] Don't require global read lock ([#3103](https://github.com/cucumber/cucumber-jvm/pull/3103))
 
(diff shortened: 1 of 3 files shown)


mpkorstanje (commented):
Superseded by https://github.com/cucumber/cucumber-jvm/pull/3236. But thanks for the inspiration.

PR #3227, opened 2026-08-14
[Core] Report source locations for duplicate parameter types

### LINKED ISSUE #3144: Give more details with DuplicateTypeNameException

nikolauspschuetz:
### 🤔 What's changed?

Catch `DuplicateTypeNameException` in `CachingGlue` and rethrow with both definitions' source locations, so a duplicate parameter type points at the two registrations instead of just naming the type.

Fixes #3144 — the current error doesn't say where the duplicates are.

### 📋 Checklist:

- [x] I agree to respect and uphold the Cucumber Community Code of Conduct
  - [x] I have added an entry to the "Unreleased" section of the CHANGELOG, linking to this pull request.

----

ai-assisted (claude): I wrote and reviewed the diff and can explain it without AI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Core] Support `cucumber.execution.threads` property when executing from CLI ([#3183](https://github.com/cucumber/cucumber-jvm/pull/3183))
 
 ### Fixed
+- [Core] Report the source location of both definitions when a parameter type is registered twice with the same name ([#3144](https://github.com/cucumber/cucumber-jvm/issues/3144))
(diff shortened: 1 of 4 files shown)


mpkorstanje (changes requested):
Cheers! 

I saw that you use Claude Code. The results weren't really good. Catching and rethrowing the `DuplicateParameterTypeException` made the implementation way more complicated than needed. For example doing a group-by makes it trivial to detect duplicate names and removes the need for the `definedParameterTypes`, the `@Nullable` in `DuplicateParameterTypeDefinitionException` ect, ect.

I think you'd have gotten better results if you'd done this yourself. Anyway tests were good so I've patched this up into something mergeable.

PR #3230, opened 2026-08-31
Document `Scenario.attach` for screenshots

### LINKED ISSUE #2309: How to attach screenshot in the report using EventListener?
Hi, I want to capture screenshots using EventListener.
I know we can use EmbedEvent for this, but there how can i use Scenario.attach()?

AzazelSensei:
I added an After-hook example for Scenario.attach in cucumber-java/README.md.

You can only call attach from glue. A plugin EventListener never gets a Scenario; it sees EmbedEvent after the attachment.

Closes #2309

### DIFF
--- a/cucumber-java/README.md
+++ b/cucumber-java/README.md
@@ -188,6 +188,28 @@ public class StepDefinitions {
 }
 ```
 
+`Scenario.attach` puts a screenshot or other media into the report. Use it from
+glue code (hooks or step definitions). Plugin `EventListener`s do not receive a
+`Scenario` instance — they see an `EmbedEvent` after the attachment is made.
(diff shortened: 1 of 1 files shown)


mpkorstanje:
Thanks for offering, you've used an LLM in such a way that it wasn't useful. I would have to spend more time coaching you to improve the PR then actual fixing this myself would take. Normally I wouldn't mind, but given that you've used an LLM I reckon it likely you'll feed my feedback right back into to it - this doesn't seem like a good use of my time.

Superseded by: https://github.com/cucumber/cucumber-jvm/pull/3232. If you're interested in learning, you can have a look at that.

PR #3236, opened 2026-09-12
Emit messages for @BeforeAll and @AfterAll hooks

### LINKED ISSUE #2748: Exceptions are swallowed (or not) depending on the `Envelope` content

### LINKED ISSUE #3054: Emit messages for @BeforeAll and @AfterAll hooks

### LINKED ISSUE #3104: Test execution does not stop after throwing exception from @AfterAll

mpkorstanje:
### 🤔 What's changed?

* Emit `TestRunHookStarted` and `TestRunHookFinished`.
* Emit `Hook` messages for before and after all hooks.
* Add support for name and hook types to `StaticHookDefinition`.
* Use `System.exit` when `Main.run` ends with an exception to kill any lingering non-deamon threads.

The implementation isn't ideal. The `CucumberExecutionContext` is used by the CLI, TestNG, JUnit 4 and the JUnit Platform. When the hooks fail, the last three still expect an exception. Now that hooks emit messages but the CLI doesn't need these anymore. This results in a lot of exception juggling which is quite confusing. But it is the best we can do for now.

Closes: https://github.com/cucumber/cucumber-jvm/issues/3054
Fixes: https://github.com/cucumber/cucumber-jvm/issues/3104
Fixes: https://github.com/cucumber/cucumber-jvm/issues/2748

### 📋 Checklist:

- [x] Update plugins to produce synthetic failures for failed hooks 
  - [x] https://github.com/cucumber/junit-xml-formatter/pull/171
  - [x] https://github.com/cucumber/testng-xml-formatter/pull/58


cucumber-core/src/main/java/io/cucumber/core/runner/Runner.java
```diff
@@ -85,44 +99,69 @@ private Locale localeForPickle(Pickle pickle) {
...
-            throwAsUncheckedException(exception);
         } catch (CucumberInvocationTargetException e) {
-            Throwable throwable = removeFrameworkFrames(e);
+            throwable = removeFrameworkFrames(e);
+        } catch (Throwable e) {
+            UnrecoverableExceptions.rethrowIfUnrecoverable(e);
+            throwable = e;
+        }
+        emitTestRunHookFinished(start, throwable, testRunHookStartedId);
+        if (throwable != null && runnerOptions.isThrowOnFailuresInStaticHooks()) {
```
mpkorstanje, line 150: Try a variant of the `TestCaseResultObserver` for test run hooks instead. This results in exceptions from hooks being collected in both the bucket for test system failures and hook failures. As a consequence, formatters render them twice.

mpkorstanje: Doens't work because the `TestRunHookFinished` events are messages and don't carry real exceptions.

